PR #20, opened 2019-05-21
fix markers and paths not showing when arcdps plugin is not available

greaka:
added protobuf support for arcdps pipeline


dlamkins:
Looks great!

TybaIt:
I dont like where this is going. ArcDPS breaks on every  patch. Why would
you want this in the main branch and not like a module dll.

Dade Lamkins <notifications@github.com> schrieb am Di., 21. Mai 2019, 20:03:

greaka:
Because this is not breaking at all when it is not present.
You can check if Arc is present at all like that: https://github.com/dlamkins/Blish-HUD/blob/70a28be9c2544e29ed71e01eabec5f352fe9ef8b/Blish%20HUD/Overlay.cs#L156
Nothing is breaking if you use it right.

dlamkins:
@greaka Next time there is a patch that breaks ArcDPS, we'll probably just want to observe the behavior of the plugin to see if it continues reporting, if certain ArcDPS endpoints become unavailable or provide incorrect information, etc. mostly for documentation: something to write down.  With how it is now, I don't anticipate any problems since it is currently either sending or not sending information based on a state within ArcDPS and the service only reacts to three different specific messages, if I'm not mistaken.

PR #31, opened 2019-06-02
ApiService Module Permissions.

TybaIt:
Adds the ability to choose the permissions per module.

Modules can Get their subtoken with user defined permissions with

`GetModuleToken(IModule module, int days)`


TybaIt:
I will have to create a new fork after this one.

PR #33, opened 2019-06-05
:100: Cleaned code, fixed loading settings, changed location of setti…

TybaIt:
…ng entries

- The location of the api permission is now defined under the namespace setting of a module IF the developer of the module has stated an array of permission requirements in the module's info.
- Fixed loading settings.
- The ApiKeyRepository setting entry remains defined under the CoreSettings.
- The code is cleaned up some more.

### DIFF
--- a/Blish HUD/GameServices/ApiService.cs
+++ b/Blish HUD/GameServices/ApiService.cs
@@ -10,8 +10,9 @@ namespace Blish_HUD
 {
     public class ApiService : GameService
     {
-        public static string SETTINGS_ENTRY = "ApiKeyRepository";
-        public static string SETTINGS_ENTRY_PERMS = "ModulePermissions";
+        public Locale LANGUAGE = Locale.English;
+        public static string SETTINGS_ENTRY_APIKEYS = "ApiKeyRepository";
+        public static string SETTINGS_ENTRY_PERMISSIONS = "Permissions";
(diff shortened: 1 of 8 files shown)


TybaIt:
Module dev important functions/fields:
```cs
/// Checks if a main client is active to create subtokens from.
public bool Connected

/// Returns the Guid of the specified Guild Wars 2 API key. Required permission: Account.
        public Guid GetGuid(string apiKey)

/// Returns an array containing the USER set permissions of the specified module.
        public static TokenPermission[] GetModulePermissions(Modules.IModule module)

/// Checks if the active API client still conforms a character name in the repository. If not, disposes the client.
        public bool Invalidate()

/// Checks if the provided string is a valid Guild Wars 2 API key.
        public static bool IsKeyValid(string apiKey)

/// Returns a fool safe dictionary containing name and the first halfs of the actual keys currently in the settings.
        public static Dictionary<string, string> GetKeyIdRepository()

/// Checks if the Api Client has the given permissions.
        public bool HasPermissions(TokenPermission[] permissions, string apiKey = null)

/// Finds a key where the first half matchs the given id and removes it from the settings.
        public void RemoveKey(string id)

/// Registers a new Guild Wars 2 API key, overwriting an existing key of the same account.
        public void RegisterKey(string apiKey)

/// Gets a subtoken for the specified module.
        public string GetModuleToken(Modules.IModule module, int days)
```

PR #36, opened 2019-07-05
Arcdps

greaka:
for reference later:
- [x] model names
- [ ] logging (TBD)
- [x] fix file naming
- [x] switch to eventhandler
- [ ] remove the counter (okay in dev, remove in master)
- [ ] port negotiation (optional for now)
- [x] stopwatch instead of gametime
- [x] summary on the model names
- [x] fix csproj
- [x] fix the crash upon exit


Blish HUD/GameServices/ArcDpsService.cs
```diff
@@ -1,96 +1,122 @@
...
+                    _stopwatch.Restart();
+                    _hudIsActive = value;
+                }
+            }
+        }
+
+        /// <summary>
+        ///     Holds unprocessed combat data
+        /// </summary>
+        public event RawCombatEvent OnRawCombatEvent;
```
dlamkins: Can we have this be `EventHandler<RawCombatEventArgs>` where `RawCombatEventArgs` is a class that inherits from `EventArgs` with a property exposed for CombatEvent data?

Additionally (and I haven't been perfect at this - I've been fixing it when I see it), the event itself should not have "On" to prefix it.

Since this service won't be inherited and basically acts as a singleton, this is much less important, but normally, the method called to fire RawCombatEvent should be called `OnRawCombatEvent` and look like:

```
private void OnRawCombatEvent(RawCombatEventArgs e) {
RawCombatEvent?.Invoke(this, e);
}
```

Blish HUD/GameServices/ArcDpsService.cs
```diff
@@ -1,96 +1,122 @@
...
         {
-            Server = new NamedPipeServerStream("BHUDrender");
-            Mut = new Mutex(false);
+            _stopwatch = new Stopwatch();
+            _server = new SocketListener(10, 200_000);
+            _server.ReceivedMessage += MessageHandler;
+#if DEBUG
+            OnRawCombatEvent += data =>
+            {
+                Interlocked.Increment(ref Counter);
```
dlamkins: `Interlocked.Read()` isn't necessary on 64-bit systems, right?  If this isn't used other than in the counters rendered I am not sure it is necessary since the UI is already refreshing so quickly that getting the absolute latest value of the counter may not be necessary.

Blish HUD/GameServices/ArcDpsService.cs
```diff
@@ -1,96 +1,122 @@
...
-            Watch = Stopwatch.StartNew();
-            Task.Run(() =>
-            {
-                while (true)
-                    try
-                    {
-                        Server.WaitForConnection();
-                        var obj = Serializer.Deserialize<Arc>(Server);
-                        Server.Disconnect();
+            _server.Start(new IPEndPoint(IPAddress.Loopback, 8214));
```
dlamkins: Was port 8214 selected randomly?  This is pedantic and while I doubt it would be an issue, I think best practice would be to select an ephemeral port from the range 49152-65535 that's open from the ArcDps plugin and then provide someway for the service to determine that port (either via registry, saved file, etc.).

Seems like it'd be a lot of trouble for the chance to save maybe one edge case, so probably not worth doing.  I think at the very least, though, a port from a higher range (say 5 digits long) would make me feel better.  :)

Blish HUD/GameServices/ArcDpsService.cs
```diff
@@ -1,96 +1,122 @@
...
+            _server.Stop();
+            RenderPresent = false;
+        }
+
+        protected override void Update(GameTime gameTime)
+        {
+            TimeSpan elapsed;
+            lock (WatchLock)
+            {
+                elapsed = _stopwatch.Elapsed;
```
dlamkins: Could you elaborate on why you're using a stopwatch instead of the elapsed time provided by `gameTime`?

dlamkins (changes requested):
Take a look at some of my comments when you have a chance.  Also, please consider making the model names a little more descriptive.

If you would like the service to produce anything in the logs (such as when the ArcDps plugin is detected as being available), you can add the logger and use it.  Just add the following to the top of each class that you would like to allow to log:
`protected static readonly NLog.Logger Logger = NLog.LogManager.GetCurrentClassLogger();`

Blish HUD/GameServices/ArcDpsService.cs
greaka: the whole counter is just for debugging while this is broadly untested. it will defintely get removed in the future. i currently dont know how i would want to approach that in the current state. im open for suggestions

Blish HUD/GameServices/ArcDpsService.cs
greaka: 8 21 4
h u d
i just checked with the wikipedia article of commonly used ports and selected something thats free
i dont like port determination, because it adds points of failure. if you have a really good idea, then convince me though. the whole point of this application is ipc anyway ^^

Blish HUD/GameServices/ArcDpsService.cs
greaka: because i cannot access the elapsed time on the other thread. and if i wanted to use it, then i would have to set some bool "resetclock" or something and do it inside the update then. if you want that, i can change it to that way

greaka:
I would like to let the model names the same as they are currently. This is only the raw representation of what arcdps sends and to comply with the arcdps doc (which people have to lookup if they want to use this), I dont want to change any names of structs or fields.

Blish HUD/GameServices/ArcDpsService.cs
greaka: can you have a look now?

Blish HUD/GameServices/ArcDpsService.cs
dlamkins: Ahh, well that makes sense.  Thanks for the clarification!

dlamkins:
> I would like to let the model names be the same as they are currently. This is only the raw representation of what arcdps sends and to comply with the arcdps doc (which people have to lookup if they want to use this), I dont want to change any names of structs or fields.
> I could go for Ev -> Event (which is a reserved keyword) or cbt (which isnt better than Ev) and ag -> Agent (which only makes sense if you include the namespace)

Okay, makes sense if you're wanting to match the ArcDps.  Could we then just put some XML comments on them to clarify them a bit more?  Just to include the full names in the <summary>.

This is an aside, but, C# keywords are case sensitive, so you can name it `Event` if you really wanted to.  It would complain if you tried to name it `event`, though.  You can get around that by prefixing the name with @, so `private bool @event` is valid - not that I recommend actually doing that last one, though.  😆

Blish HUD/GameServices/ArcDpsService.cs
dlamkins: Okay, well for the clever choice, I'll have to give you this one!

Blish HUD/GameServices/ArcDpsService.cs
dlamkins: After reading more about it, I'm thinking my complaints are not that big of a deal.  If we find there is an issue later, we can change it before merging into [master](../../tree/master).

Blish HUD/GameServices/ArcDpsService.cs
dlamkins: Looks perfect!

dlamkins:
@greaka Latest CI fails to build with:

> (CoreCompile target) -> 
>   CSC : error CS2001: Source file 'C:\projects\blish-hud\Blish HUD\Custom\Pipeline.cs' could not be found. [C:\projects\blish-hud\Blish HUD\Blish HUD.csproj]
>   CSC : error CS2001: Source file 'C:\projects\blish-hud\Blish HUD\GameServices\ArcDps\SockerListener.cs' could not be found. [C:\projects\blish-hud\Blish HUD\Blish HUD.csproj]

Did you rename and delete the files outside of VS?  It doesn't look like the csproj file knew you changed the file names / deleted those files.

Blish HUD/GameServices/ArcDps/SocketListener.cs
```diff
@@ -0,0 +1,211 @@
...
+                acceptEventArg = new SocketAsyncEventArgs();
+                acceptEventArg.Completed += (sender, e) => ProcessAccept(e);
+            }
+            else
+            {
+                acceptEventArg.AcceptSocket = null;
+            }
+
+            _acceptedClientsSemaphore.WaitOne();
+            if (!_listenSocket.AcceptAsync(acceptEventArg)) ProcessAccept(acceptEventArg);
```
dlamkins: This is throwing an unhandled exception when the application attempts to close itself by calling `Exit()` on the BlishHud.cs instance.

image

dlamkins: You can replicate this by right-clicking on the Blish HUD icon in your tray and selecting "Close Blish HUD."

dlamkins (changes requested):

Blish HUD/GameServices/ArcDps/SocketListener.cs
greaka: This is a known bug, since I first implemented it.
The GC is faster with cleaning up than my `_server.Stop()` and I have no solution currently.
I just have to find out where the try except has to go, because it doesnt crash at the same line all the time. It just crashes wherever it currently is.
Is on the TODO

greaka:
> @greaka Latest CI fails to build with:
> 
> > (CoreCompile target) ->
> > CSC : error CS2001: Source file 'C:\projects\blish-hud\Blish HUD\Custom\Pipeline.cs' could not be found. [C:\projects\blish-hud\Blish HUD\Blish HUD.csproj]
> > CSC : error CS2001: Source file 'C:\projects\blish-hud\Blish HUD\GameServices\ArcDps\SockerListener.cs' could not be found. [C:\projects\blish-hud\Blish HUD\Blish HUD.csproj]
> 
> Did you rename and delete the files outside of VS? It doesn't look like the csproj file knew you changed the file names / deleted those files.
> 
> Edit: I checked, and this does seem to be what happened. Should be a quick fix.
> 
> image

Interesting, I built it afterwards and didnt get that crash

dlamkins (approved):
I think this is good!  Thanks for making the requested changes.  I think I'll review the potential for logging later.

To clarify about the counter: it's not that it needs to be removed in master (since it wouldn't be included in release builds anyways), I was just speaking out loud about the thread concerns and them being something we could address later if we noticed problems.  Basically "we'll review it again when we go to merge dev into master".

PR #39, opened 2019-07-08
Revert "Moduler modules rework"

dlamkins:
Reverts blish-hud/Blish-HUD#37

Wrong base.


dlamkins:
Items failing are due to the revert.  This is expected.

PR #40, opened 2019-07-09
ArcDps Common

greaka:
This PR aims to bring QoL features for consumers of the ArcDps integration.

The first commit brings the option to subscribe to specific skill ids - those are also buff ids

Depending on the feedback thats coming in the discord, I will also add some common Properties for access to everyone - like if and which mount is currently equipped
I just dont know yet what the best option would be to just conditionally listen to those events

ref: https://discordapp.com/channels/531175899588984842/534490472224391169/598194017787969537


greaka:
note to self: make the dictionary thread safe

greaka:
ref: https://discordapp.com/channels/531175899588984842/534490472224391169/598194017787969537

greaka:
until we get more info to what is needed, i want to merge this for now
this is currently stable

agaertner (approved):
Other than the mttb changes which will probably get changed in the future anyway I don't see anything wrong with this.

PR #41, opened 2019-07-10
Cleaned ContentModule some and "Errors" were changed to "Warn" due to the nature of the exceptions they were logging for.

dlamkins:


Blish HUD/GameServices/ContentService.cs
```diff
@@ -132,7 +132,7 @@ public enum FontStyle {
 
         private static Texture2D TextureFromFileSystem(string filepath) {
             if (!File.Exists(REF_FILE)) {
-                Console.WriteLine($"{REF_FILE} is missing!  Lots of assets will be missing!");
+                Logger.Warn("{refFileName} is missing!  Lots of assets will be missing!", REF_FILE);
```
greaka: The usage of `refFileName` is confusing. Either stick to `{}` with the format string format, or use the string interpolation `$"{REF_FILE}"` please.

greaka (changes requested):
looks good so far
Since you didnt open a draft pr, I assume that you want to merge this and reviewed it.

Blish HUD/GameServices/ContentService.cs
dlamkins: Could you please elaborate on that first option?  NLog best practice is to allow for the additional params to be sent to the target logger.  https://github.com/NLog/NLog/wiki/How-to-use-structured-logging

greaka: Ah, didn't know of that! I didn't use NLog in a *long* time, so I missed the structured logging update. I have to read into that.

greaka (approved):

greaka (approved):

greaka:
hmm, github seems to not like my late approvals..

dlamkins:
> hmm, github seems to not like my late approvals..
> EDIT: found the problem, I have no write access anymore 😄

All fixed!

PR #46, opened 2019-07-13
Added missing heart texture and switched to using const string for about message.

dlamkins:
Minor fix.


dlamkins:
Includes itself as a submodule.  😑

PR #48, opened 2019-07-14
fixed small bugs in arcdps service

greaka:


Blish HUD/GameServices/ArcDps/CombatParser.cs
```diff
@@ -44,7 +45,7 @@ private static (Ev, int) ParseEv(byte[] data, int offset)
             ushort srcMasterInstId;
             ushort dstMasterInstId;
             byte iff;
-            byte buff;
+            bool buff;
```
dlamkins: Check line 91 of this same file.  The constructor still expects it to be a byte:
https://github.com/blish-hud/Blish-HUD/blob/ffb4bd6a3b82f4d6617a83bb4cf38754ec2d7229/Blish%20HUD/GameServices/ArcDps/Models/Ev.cs#L8-L12

Failing in CI:

> "C:\projects\blish-hud\Blish HUD.sln" (default target) (1) ->
> "C:\projects\blish-hud\Blish HUD\Blish HUD.csproj" (default target) (2) ->
> (CoreCompile target) -> 
>   GameServices\ArcDps\CombatParser.cs(91,56): error CS1503: Argument 13: cannot convert from 'bool' to 'byte' [C:\projects\blish-hud\Blish HUD\Blish HUD.csproj]

greaka: yep, forgot to add it with the bulk of files that i modified 🙈

PR #43, opened 2019-07-10
Removed Musician assets, added new Book and Page controls, and added NewKrytan font.

TybaIt:
Added ./Controls/BasicWindow.cs, 
Added ./Controls/Book.cs
Added ./Controls/Page.cs
Removed Musician assets from ref/
Added newkrytan-12-regular bitmapfont files to MonoGame Pipeline. (./Content/fonts/newkrytan/)

PR #71, opened 2019-07-21
Removed Musician assets, added new Book and Page controls, and added NewKrytan font.

dlamkins:
Looks like #43 was confusing to look at because it was a PR to merge into master instead of dev.  That's why it was showing so much.


dlamkins:
Probably best to just have @TybaIt redo the PR instead of it coming in through my name.

PR #80, opened 2019-08-08
Changed RasterizerState to CullNone

### LINKED ISSUE #79: Default RasterizerState cull set to CounterClockwise (regression - should be CullNone)
https://github.com/blish-hud/Blish-HUD/blob/0404529197a3ba54db80e96afd5a50bcef2c4ecf/Blish%20HUD/Entities/World.cs#L40

Since most markers and paths are all a single face, culling was originally disabled (to allow you to see trails from under or behind them).  It was set to CounterClockwise in https://github.com/blish-hud/Blish-HUD/blame/0404529197a3ba54db80e96afd5a50bcef2c4ecf/Blish%20HUD/Entities/World.cs#L40

This needs to be set back to CullNone: https://github.com/blish-hud/Blish-HUD/blame/24ab3209489ddb71c3c8622bcc5959a7c6fdbad2/Blish%20HUD/Entities/World.cs#L35

**As an aside:**
When we begin rendering more full meshes, we'll likely be switching between culling and not culling mid-draw (so draw all single face entities without culling, then 3D objects with culling), which will likely be costly.  When the time comes, we should benchmark that vs. just duplicating the faces of the trails and markers on the other side of the entity (2x render faces vs changing the rasterizer state mid-draw potential 2 or more times).  This isn't important to us now, though.

Related to: https://github.com/blish-hud/Blish-HUD/issues/65

nickcode122:
Changed RasterizerState to CullNone to allow viewing trails from under or behind. Fixes #79

### DIFF
--- a/Blish HUD/Entities/World.cs
+++ b/Blish HUD/Entities/World.cs
@@ -37,7 +37,7 @@ namespace Blish_HUD.Entities {
             graphicsDevice.BlendState        = BlendState.AlphaBlend;
             graphicsDevice.DepthStencilState = DepthStencilState.DepthRead;
             graphicsDevice.SamplerStates[0]  = SamplerState.LinearWrap;
-            graphicsDevice.RasterizerState   = RasterizerState.CullCounterClockwise;
+            graphicsDevice.RasterizerState   = RasterizerState.CullNone;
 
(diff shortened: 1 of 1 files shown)


dlamkins (approved):
Looks perfect!  😄

PR #89, opened 2019-08-17
fix dev branch and some other common fields bugs

greaka:
forgot player on public dictionary on common fields, sry :/


greaka:
fixed some bugs that andy reported in the discord

agaertner:
Might be worth putting an Equals into Player model for good practic.

agaertner (approved):
Looks good.

PR #90, opened 2019-08-18
Minor cleanup to the GameIntegrationService and the InputService

dlamkins:
Removed some Console.WriteLine lines and replaced them with a Logger warn.  Also moved the MouseHook and KeyboardHook calls from the utils into the Input namespace.  Minor comment improvements, as well.


greaka (dismissed):
I like keeping comments that point you to SO if you copied it from there, otherwise all good 🙂

dlamkins:
> I like keeping comments that point you to SO if you copied it from there, otherwise all good 🙂

Alright, I put it back.  I think you're right.  🙂

dlamkins:
@greaka has approved this minor change in the channel since it was a very minor change:

> Just approve it yourself :grin:
https://discordapp.com/channels/531175899588984842/536970543736291346/612903474467962892

PR #99, opened 2019-08-22
couple of control fixes.

dlamkins:
- Minor fixes to `Container` control.
- Fixed error in `Menu` control that would throw exception if there were no `MenuItems` within it when calculating the height of the `Menu`.
- Fixed possible error in `TabbedWindow` control that could throw an exception if the window has no tabs before something tries to update or interact with the window.
- Fixed `StringUtil.WrapText()` from throwing an exception.
- Added `ClearChildren()` to `Container`s.
- Exposed `CurrentGameTime` through the `OverlayService`.
- Added new `LoadingSpinnerUtil.DrawLoadingSpinner()` and switched the `LoadingSpinner` to use the util.
- Changed `CornerIcon` to be a standard `Control` with `LoadingSpinner` rendered using `LoadingSpinnerUtil`.
- Updated `OverlayService` to define the `IconName` property on the `CornerIcon`.

The changes to the `CornerIcon` is a soft-breaking change as `BasicTooltipText` is no longer kept on `CornerIcon` (used to set the text shown when `CornerIcon` is hovered over).  `CornerIcon` now has `IconName` property which can be used to set this instead.  New constructors on the `CornerIcon` now allow for passing the `Icon`, `HoverIcon`, and also the `IconName`.

### DIFF
--- a/Blish HUD/Controls/Container.cs
+++ b/Blish HUD/Controls/Container.cs
@@ -22,7 +22,8 @@ namespace Blish_HUD.Controls {
 
         public event EventHandler<RegionChangedEventArgs> ContentResized;
 
-        protected readonly List<Control> _children;
+        protected List<Control> _children;
+
         [Newtonsoft.Json.JsonIgnore]
         public IReadOnlyCollection<Control> Children => _children.AsReadOnly();
 
@@ -124,6 +125,16 @@ namespace Blish_HUD.Controls {
             }
 
             return allDescendants;
+        }
+
+        public void ClearChildren() {
+            Control[] oldChildren = _children.ToArray();
+
+            _children = new List<Control>();
+
+            for (int i = 0; i < oldChildren.Length; i++) {
+                oldChildren[i].Parent = null;
+            }
         }
 
         public override Control TriggerMouseInput(MouseEventType mouseEventType, MouseState ms) {
--- a/Blish HUD/Controls/Container.cs
+++ b/Blish HUD/Controls/Container.cs
@@ -14,16 +14,15 @@ namespace Blish_HUD.Controls {
     /// <summary>
     /// A control that is capable of having child controls that are drawn when the container is drawn.
     /// Classes that inherit should be packaged controls that that manage their own controls internally.
-    /// For a Container with accessible children for non-packaged controls, use <see cref="AccessibleContainer"/>.
     /// </summary>
-    public abstract class Container:Control, IEnumerable<Control> {
+    public abstract class Container : Control, IEnumerable<Control> {
 
         public event EventHandler<ChildChangedEventArgs> ChildAdded;
         public event EventHandler<ChildChangedEventArgs> ChildRemoved;
 
         public event EventHandler<RegionChangedEventArgs> ContentResized;
 
-        protected List<Control> _children;
+        protected readonly List<Control> _children;
         [Newtonsoft.Json.JsonIgnore]
         public IReadOnlyCollection<Control> Children => _children.AsReadOnly();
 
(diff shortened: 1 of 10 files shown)


Blish HUD/Controls/Container.cs
```diff
@@ -126,6 +127,16 @@ public abstract class Container:Control, IEnumerable<Control> {
             return allDescendants;
         }
 
+        public void ClearChildren() {
```
greaka, line 130: why is this approach better than

```cs
_children.ForEach(c => c.Parent = null);
_children.Clear();
```
?

Blish HUD/Controls/Menu.cs
```diff
@@ -128,24 +124,24 @@ public class Menu : Container, IMenuItem {
         }
 
         public override void UpdateContainer(GameTime gameTime) {
-            int totalItemHeight = _children.Where(c => c.Visible).Max(c => c.Bottom);
+            int totalItemHeight = 0;
+
+            for (int i = 0; i < _children.Count; i++) {
+                totalItemHeight = Math.Max(_children[i].Bottom, totalItemHeight);
+            }
```
greaka, line 131: Why is this better than

```cs
var totalItemHeight = _children.Max(c => c.Bottom);
```
?

dlamkins: Saves the alloc from `GetEnumerator()` and from the LINQ call. I'm looking to reduce LINQ + foreach usage where I find it in the update or render loop.  One-offs are fine (like action from clicking a button or something).

Not a huge issue, but its 2 seconds switching between implementations with ReSharper.  I'll benchmark allocs and performance later and see if I can back these actions with numbers.  I'm mostly basing this on (primarily for Unity) performance write ups I've read.

Blish HUD/Controls/Container.cs
dlamkins: Changing the parent would modify the collection while it was enumerated (and throw an exception).  Setting `.Parent` the control calls `AddChild()` or `RemoveChild()` on the container.

Blish HUD/Controls/Menu.cs
dlamkins: @greaka Here, I got some benchmarks:

``` ini

BenchmarkDotNet=v0.11.5, OS=Windows 10.0.14393.447 (1607/AnniversaryUpdate/Redstone1), VM=VirtualBox
Intel Core i7-8550U CPU 1.80GHz (Kaby Lake R), 1 CPU, 1 logical core and 1 physical core
Frequency=1945297 Hz, Resolution=514.0603 ns, Timer=TSC
  [Host]     : .NET Framework 4.7.2 (CLR 4.0.30319.42000), 64bit RyuJIT-v4.7.3062.0
  DefaultJob : .NET Framework 4.7.2 (CLR 4.0.30319.42000), 64bit RyuJIT-v4.7.3062.0


```
|      Method |        Mean |      Error |     StdDev |      Median | Rank |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|------------ |------------:|-----------:|-----------:|------------:|-----:|-------:|------:|------:|----------:|
|    WithLinq | 3,649.66 ns | 109.791 ns | 309.667 ns | 3,553.09 ns |    2 | 0.0153 |     - |     - |      80 B |
| WithoutLinq |    71.29 ns |   1.668 ns |   4.423 ns |    69.81 ns |    1 |      - |     - |     - |         - |

greaka (approved):

dlamkins:
After the fact, we have found that there was another breaking change.  WindowTab went from struct to a class.  No module code will need to be changed, but all modules must be recompiled that utilized the WindowTab struct/class.

PR #105, opened 2019-08-25
Static chat code service, changed IKeyboard

agaertner:
- provided static chat code functions which modules can use to customize chat code copy-to-clipboard buttons.
- changed Keyboard and removed IKeyboard
- Keyboard is now static.

(also see PR for Musician -> https://github.com/blish-hud/Community-Module-Pack/pull/20)

### DIFF
--- a/Blish HUD/Controls/Intern/Keyboard.cs
+++ b/Blish HUD/Controls/Intern/Keyboard.cs
@@ -3,13 +3,8 @@ using System.Collections.Generic;
 using Blish_HUD.Controls.Extern;
 namespace Blish_HUD.Controls.Intern
 {
-    public class Keyboard : IKeyboard
+    public static class Keyboard
     {
-        /// <summary>
-        /// If true, uses normal key input instead of sending key event messages to GW2's window handle.
-        /// </summary>
-        public bool HardwareInput = false;
-
         private const uint WM_KEYDOWN = 0x0100;
         private const uint WM_KEYUP = 0x0101;
         private const uint WM_CHAR = 0x0102;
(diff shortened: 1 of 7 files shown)


Blish HUD/Controls/Intern/Keyboard.cs
```diff
@@ -74,17 +41,21 @@ public void Press(GuildWarsControls key)
             }
             else
             {
-                uint vkCode = (uint)VirtualKeyShorts[key];
+                uint vkCode = (uint)key;
                 ExtraKeyInfo lParam = new ExtraKeyInfo(){
                     scanCode = (char)PInvoke.MapVirtualKey(vkCode, MAPVK_VK_TO_VSC)
                 };
-                PInvoke.PostMessage(Gw2WndHandle, WM_KEYDOWN, vkCode, lParam.GetInt());
+                PInvoke.PostMessage(GameService.GameIntegration.Gw2Process.MainWindowHandle, WM_KEYDOWN, vkCode, lParam.GetInt());
```
dlamkins: Use `GameService.GameIntegration.Gw2WindowHandle` instead of pulling it from the `Gw2Process` since it is already exposed.

Blish HUD/Controls/Intern/Keyboard.cs
```diff
@@ -106,15 +77,15 @@ public void Release(GuildWarsControls key)
...
+                uint vkCode = (uint)key;
                 ExtraKeyInfo lParam = new ExtraKeyInfo
                 {
                     scanCode = (char)PInvoke.MapVirtualKey(vkCode, MAPVK_VK_TO_VSC),
                     repeatCount = 1,
                     prevKeyState = 1,
                     transitionState = 1
                 };
-                PInvoke.PostMessage(Gw2WndHandle, WM_KEYUP, vkCode, lParam.GetInt());
+                PInvoke.PostMessage(GameService.GameIntegration.Gw2Process.MainWindowHandle, WM_KEYUP, vkCode, lParam.GetInt());
```
dlamkins: Use `GameService.GameIntegration.Gw2WindowHandle` instead of pulling it from the `Gw2Process` since it is already exposed.

Blish HUD/GameServices/Gw2ChatCodeService.cs
```diff
@@ -0,0 +1,101 @@
+﻿using Microsoft.Xna.Framework;
+using System;
+using System.Collections.Generic;
+using System.Text.RegularExpressions;
+
+namespace Blish_HUD
+{
+    public class Gw2ChatCodeService : GameService
```
dlamkins: I think I'd like to put this in a `Blish_HUD.Common` namespace almost like a util.  Perhaps `Blish_HUD.Common.Gw2.ChatCode`?  GameServices are better suited for things that need access to an update loop.  This wouldn't be good as a service - especially since everything publicly exposed is static, anyways.

dlamkins: I think we should do that in a different PR, though.  The `Common` namespace will also be good if we bring things over for reading dat files in a structured way.

dlamkins (changes requested):
Update those members to the already exposed ones and seperate the gameservice out so that it can be put into the Common namespace at some point in a new PR and we should be good to merge.

Blish HUD/GameServices/Gw2ChatCodeService.cs
agaertner: Ok I move Gw2ChatCodeService to `Common.Gw2.ChatCode`
I will do this with the commit for the change regarding Gw2WndHandle in Keyboard.cs

dlamkins: Thanks!

Blish HUD/Controls/Intern/Keyboard.cs
dlamkins: Reviewed and it looks like this was fixed.  Also thanks for updating the scope.

dlamkins (approved):
Thanks for the fixes!

Issue #62, opened 2019-07-19
Discussion: Rework fallible Contexts for better error handling


greaka:
this is more of a request for discussion:
theres not really a way to predict exact timings of annual events
so what would you say if the marker module had an optional(!) service that it would call thats gets timings for events. we would have to manually insert those events (so every 2 months putting in some times), the festival service (because maybe some other module could benefit from it?) would either expose if it can reach the service at all (kinda like arcdps service currently) or default to every event active when the service is unreachable
that way we could expose markers to hide depending on festivals

![example image](https://cdn.discordapp.com/attachments/534490472224391169/601637629221666856/unknown.png)

ref: https://discordapp.com/channels/531175899588984842/534490472224391169/601637629221666857

dlamkins:
It may be worth putting together a "ContextService". Something that tries to piece together the available information to give an educated guess on the clients current state, etc. That could even be where detecting mounts from the ArcDps service could potentially live.

greaka:
I dont know, i think its good that we distinguish between stuff that needs 3rd party (read: my arcdps plugin) and the rest

dlamkins:
Oh, that's a good point.

I'm wondering now if we would benefit from restructuring things a little bit to have `Context`s that target specific parts of the game and each one uses information compiled from potentially multiple sources to try and describe the current state of things.

Then we could move some of the services in as contexts, instead.  `PlayerService` would be a good candidate for this since some data would be from `GwMumbleService` and some from `Gw2ApiService`, potentially.  It would become the `PlayerContext` or something like that.  Contexts could consume information provided by services to try and describe their current state and maybe indicate their "certainty" somehow.

It would be similar to how currently the `CameraService` basically acts as a proxy on top of the `Gw2MumbleService` despite the camera information technically being provided by the `Gw2MumbleService`.

Handling "certainty" seems like it would be the fairly difficult part.  For some things, it could simply be event based like "PlayerMounted" or "PlayerUnmounted," which would assert with full certainty what happened since they would require a known state change to occur in order to happen.

Alternatively, module writers could make sure that their module degrades gracefully if certain information is not defined or provided.

greaka:
I like the idea of having an abstract base context, that has a field for `isPopulated` or something and maybe some other stuff
That way modules could check if they can rely on those information

dlamkins:
[...]

Issue #50, opened 2019-07-16
move icon when on chinese client


greaka:
move the bhud icon one slot to the right when we detect a chinese client, as they have an additional icon on the left upper corner icon bar

REF: https://discordapp.com/channels/531175899588984842/534492012263636992/600677943093493760

dlamkins:
We can base this off of the localization, which could be convenient.  

It might be good to expose some advance features that allow for manually setting the number of icons ours are offset by to better future proof things, too.

greaka:
thought about that as well but what if people prefer english over chinese?
i prefer english over german for example

dlamkins:
That's what I figured would be a good use case for "It might be good to expose some advance features that allow for manually setting the number of icons ours are offset by to better future proof things, too."

It may be easier to detect the client specifically, though.  It's been a while since it was last discussed, so I forget: did we find a way to detect the Chinese client specifically?  If we can detect the chinese client, then that would probably be a better method.

greaka:
mumble link almost certainly should contain chinese characters ^^

dlamkins:
I suppose the Identity -> name likely would.  Not sure there is any guarantee on that, though.  I'll ask.

dlamkins:
Had some discussion about this, here: https://discordapp.com/channels/384735285197537290/384740380014280714/600730077688365087

Seems like a good option would be to use the Mumble connection to get the build at launch and compare it to the build endpoint's response.  The web API will return the non-chinese build number and it should not match the build number returned by the Mumble connection.

dlamkins:
REF: https://discordapp.com/channels/384735285197537290/384735523521953792/600743022790967306

We'll compare the build number provided by the MumbleLink API with the build number at: http://assetcdn.111.cgw2.com/latest/111 which represents the current Chinese build number.

dlamkins:
Resolved in #106

PR #106, opened 2019-08-25
Contexts + CdnInfoContext & Gw2ClientContext

dlamkins:
Added base structure for new `ContextsService`.  Also added two contexts `CdnInfoContext` and `Gw2ClientContext` which demonstrate detecting the Client version of the application using the Mumble Link service to detect the current build ID and then the asset CDNs to check which client we are using based on that current build ID.

This provides some of the potential implementation currently being discussed in #62. 

Additionally, if accepted, this contains contexts that would allow us to more easily solve #50.

### DIFF
--- a/Blish HUD/GameServices/Contexts/CdnInfoContext.cs
+++ b/Blish HUD/GameServices/Contexts/CdnInfoContext.cs
@@ -27,7 +27,7 @@ namespace Blish_HUD.Contexts {
 
                 bool parsedSuccessfully = true;
 
-                if (cdnVars.Length == 5) {
+                if (cdnVars.Length == 5 || (parsedSuccessfully = false)) {
                     parsedSuccessfully &= int.TryParse(cdnVars[0], out _buildId);
                     parsedSuccessfully &= int.TryParse(cdnVars[1], out _exeFileId);
                     parsedSuccessfully &= int.TryParse(cdnVars[2], out _exeFileSize);
--- a/Blish HUD/GameServices/Contexts/CdnInfoContext.cs
+++ b/Blish HUD/GameServices/Contexts/CdnInfoContext.cs
@@ -4,10 +4,16 @@ using Flurl.Http;
 
 namespace Blish_HUD.Contexts {
 
+    /// <summary>
+    /// Provides build information provided by the asset CDNs.
+    /// </summary>
     public class CdnInfoContext : Context {
 
         private static readonly Logger Logger = Logger.GetLogger(typeof(CdnInfoContext));
-        
+
+        /// <summary>
+        /// Structured information provided by one of the asset CDNs.
+        /// </summary>
         public struct CdnInfo {
 
             private readonly int _buildId;
@@ -42,8 +48,7 @@ namespace Blish_HUD.Contexts {
         private CdnInfo _standardCdnInfo;
         private CdnInfo _chineseCdnInfo;
 
-        private bool _loading   = false;
-        private int  _loadCount = 0;
+        private int _loadCount = 0;
 
         #region Context Management
 
@@ -53,10 +58,6 @@ namespace Blish_HUD.Contexts {
 
         /// <inheritdoc />
         protected override void Load() {
-            if (_loading) return;
-
-            _loading = true;
-
             GetCdnInfoFromCdnUrl(GW2_ASSETCDN_URL).ContinueWith((cdnInfo) => SetCdnInfo(ref _standardCdnInfo,   cdnInfo.Result));
             GetCdnInfoFromCdnUrl(GW2_CN_ASSETCDN_URL).ContinueWith((cdnInfo) => SetCdnInfo(ref _chineseCdnInfo, cdnInfo.Result));
         }
@@ -67,7 +68,9 @@ namespace Blish_HUD.Contexts {
         }
 
         private void GameIntegrationOnGw2Started(object sender, EventArgs e) {
-            this.DoUnload();
+            // Unload without DoUnload to avoid expiring the context
+            this.Unload();
+
             this.DoLoad();
         }
 
@@ -104,8 +107,6 @@ namespace Blish_HUD.Contexts {
             cdnInfo = ParseCdnInfo(result);
 
             if (++_loadCount >= TOTAL_CDN_ENDPOINTS) {
-                _loading = false;
-
                 ConfirmReady();
             }
         }
@@ -129,19 +130,19 @@ namespace Blish_HUD.Contexts {
         #endregion
 
         private ContextAvailability TryGetCdnInfo(ref CdnInfo cdnInfo, out ContextResult<CdnInfo> contextResult) {
-            if (_loading) return NotReady(out contextResult);
+            if (this.State != ContextState.Ready) return NotReady(out contextResult);
 
             if (cdnInfo.BuildId > 0) {
-                contextResult = new ContextResult<CdnInfo>(cdnInfo, true);
+                contextResult = new ContextResult<CdnInfo>(cdnInfo);
                 return ContextAvailability.Available;
             }
 
             if (cdnInfo.BuildId < 0) {
-                contextResult = new ContextResult<CdnInfo>(cdnInfo, false);
+                contextResult = new ContextResult<CdnInfo>(cdnInfo, "Failed to determine build ID from CDN.");
                 return ContextAvailability.Failed;
             }
 
-            contextResult = new ContextResult<CdnInfo>(cdnInfo, false);
+            contextResult = new ContextResult<CdnInfo>(cdnInfo, "Build ID has not been requested from the CDN.");
             return ContextAvailability.Unavailable;
         }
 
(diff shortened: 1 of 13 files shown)


Blish HUD/GameServices/Contexts/CdnInfoContext.cs
```diff
@@ -0,0 +1,128 @@
...
+
+                if (cdnVars.Length == 5) {
+                    parsedSuccessfully &= int.TryParse(cdnVars[0], out _buildId);
+                    parsedSuccessfully &= int.TryParse(cdnVars[1], out _exeFileId);
+                    parsedSuccessfully &= int.TryParse(cdnVars[2], out _exeFileSize);
+                    parsedSuccessfully &= int.TryParse(cdnVars[3], out _manifestFileId);
+                    parsedSuccessfully &= int.TryParse(cdnVars[4], out _manifestFileSize);
+                }
+
+                if (!parsedSuccessfully) {
```
greaka: minor bug: `parsedSuccessfully` can return the wrong result

```cs
bool parsedSuccessfully = false;

if (cdnVars.Length == 5) {
    parsedSuccessfully = true;
    ...
}
if (!parsedSuccessfully) {

File: Blish HUD/GameServices/Contexts/CdnInfoContext.cs [hunk 2]
```diff
@@ -0,0 +1,128 @@
...
+        private CdnInfo _standardCdnInfo;
+        private CdnInfo _chineseCdnInfo;
+
+        private int _loadCount = 0;
+
+        #region Context Loading
+
+        protected override void Load() {
+            GetCdnInfoFromCdnUrl(GW2_ASSETCDN_URL).ContinueWith((cdnInfo) => SetCdnInfo(ref _standardCdnInfo, cdnInfo.Result));
+            GetCdnInfoFromCdnUrl(GW2_CN_ASSETCDN_URL).ContinueWith((cdnInfo) => SetCdnInfo(ref _chineseCdnInfo, cdnInfo.Result));
```
greaka, line 62: is `ContinueWith` safe to use when `GetCdnInfoFromCdnUrl` returns `null`?

Blish HUD/GameServices/Contexts/Gw2ClientContext.cs
```diff
@@ -0,0 +1,92 @@
...
+
+            if (contextAvailability == ContextAvailability.Available) {
+                int standardBuildId = standardCdnContextResult.Value.BuildId;
+
+                if (currentBuildId == standardBuildId) {
+                    return true;
+                }
+            }
+
+            return false;
```
greaka: This is probably preference, I would write that a little bit different for readability:

```cs
if (contextAvailability != ContextAvailability.Available)
    return false;

var standardBuildId = standardCdnContextResult.Value.BuildId;
return currentBuildId == standardBuildId;
```

Blish HUD/GameServices/Contexts/Gw2ClientContext.cs
```diff
@@ -0,0 +1,92 @@
...
+
+            if (contextAvailability == ContextAvailability.Available) {
+                int chineseBuildId = chineseCdnContextResult.Value.BuildId;
+
+                if (currentBuildId == chineseBuildId) {
+                    return true;
+                }
+            }
+
+            return false;
```
greaka: see comment above

greaka (changes requested):
I like the propagation of the availability, even though c# makes it a bit weird to work with.
All in all a good concept. 


`Blish HUD/GameServices/Contexts/Gw2ClientContext.cs` needs a refresh on a game start, otherwise it will return the wrong client type whenever you switch the game client.
also, it will start logging that no version could be matched when the game build updates and we fetch a new mumble link build info

Blish HUD/GameServices/Contexts/CdnInfoContext.cs
dlamkins: Oh, good catch!  Is this too dirty of a fix?  https://dotnetfiddle.net/irafyW

Blish HUD/GameServices/Contexts/Gw2ClientContext.cs
dlamkins: Yep!  Could even just be:

```cs
if (contextAvailability != ContextAvailability.Available)
    return false;

return currentBuildId == standardCdnContextResult.Value.BuildId;
```

That var isn't very helpful at that point, really.

Blish HUD/GameServices/Contexts/Gw2ClientContext.cs
dlamkins: ✔ Agreed.

dlamkins:
@greaka I responded to a couple of the items here so far.  Just FYI, this implementation was not completed.  It was more of a proof of concept that, if folks thought was decent enough, could later be finished to be the first implementation of it.  Refreshing some of the context info, handling failed HTTP requests, etc. would all be included if the concept itself looked good to you.  

I just didn't want to spend too much time on it unless folks confirmed they liked the implementation.  😄 

If you do like the implementation (ignoring the flaws of the current implemented contexts you've mentioned), or have any additional thoughts, could you let me know over in #62? 😃

Blish HUD/GameServices/Contexts/CdnInfoContext.cs
greaka: I would like to advise against doing that. the compiler should be able to optimize both solutions, but a normal developer would definitely oversee that you didnt write `==`

Blish HUD/GameServices/Contexts/Gw2ClientContext.cs
greaka: I didnt want to change too much, but yes, I would omit the var as well

Blish HUD/GameServices/Contexts/CdnInfoContext.cs
dlamkins: Sounds good!  I'll put that change in tonight.

Blish HUD/GameServices/Contexts/Gw2ClientContext.cs
dlamkins: Great - I've got this fixed in my local branch.  I'll push it tonight so we can start checking some of these items off.

Blish HUD/GameServices/Contexts/CdnInfoContext.cs
dlamkins: Yes, there shouldn't be any problems with this.  I tested last night with Fiddler to ensure various request failures would be handled correctly.

`SetCdnInfo` handles `null` without issue with its first check.  `ContinueWith` will pass the result regardless.

The local branch that I'll be pushing tonight has some additional logging to ensure the response status code and status reason are logged if the request fails for some reason.

dlamkins:
- Resolved bug with CdnInfo where parsedSuccessfully could be true if the number of values was not 5. https://github.com/blish-hud/Blish-HUD/pull/106#discussion_r318001999
- Reduced complexity of "Is______ClientType" function.  https://github.com/blish-hud/Blish-HUD/pull/106#discussion_r318009487 & https://github.com/blish-hud/Blish-HUD/pull/106#discussion_r318009731
- Fixed bugs in CdnInfoContext as well as Gw2ClientContext.
- Contexts can now be unloaded.
- Gw2MumbleService now has an event "BuildIdChanged"
- OverlayService now, in addition to TacO, checks to see if the running client is the standard client or the Chinese client (moves the CornerIcons over if it is the Chinese client).

dlamkins:
@greaka While I think the primary focus of this PR is the Contexts implementation, I think it'd be good to have Lei test the artifact that is created from this build and ensure the Blish HUD `CornerIcon` moves over as expected on a real Chinese version of the client.  I did a good amount of testing with Fiddler to fake the build ID that is returned and was able to confirm it should work.  Also ensured that it would fallback gracefully if something fails with the request (and logs are much more detailed with why it failed, now).

If you like this latest implementation, I can finish writing up the XML docs for it and mark it ready for a true review.

Once we get this merged, we can start adding things like the `FestivalContext`, which I am fairly excited about.  🎉

greaka (dismissed):

dlamkins:
@greaka Please re-review the implementation when you have a chance.  I believe I have implemented everything now and have put in XML documentation where appropriate.  I've also added a `State` member to `Context`s which indicates if the `Context` is loading, ready, etc.

To prevent anybody from unregistering a `Context`, I added a `ContextHandle` which can be used to expire a `Context`.  The `ContextHandle` associated with a `Context` type is returned when you call `RegisterContext`.  The full workflow would look something like:

```cs
// Register the context when the module is loaded
var myContextHandle = GameService.Contexts.RegisterContext(new ExampleContext());

// Expire context when the module is unloaded
myContextHandle.Expire();
```

Also, I received confirmation that this implementation has fixed Lei's problem!  🙂 
https://cdn.discordapp.com/attachments/568470477543178261/617894300327477249/jt_2019-09-02_09-30-59.png

greaka (approved):

greaka:
I have multiple nitpicks in case the contexts are meant to be thread safe.
Having them threadsafe is probably a nice to have for now though

dlamkins:
Agreed!

PR #110, opened 2019-08-26
Optional ForGameWindowOnly added to Keyboard.

agaertner:
Small change.

I removed that Hardware bool when I cleaned this but I just remembered why it was there.

Some keys aren't correctly interpreted when they are sent as window messages, like LCONTROL, LSHIFT etc.  
This is an optional bool on the Press() Release() functions. Default is true. ie is sent to gw2 window as message.


greaka (changes requested):
`ForGameWindowOnly` is confusing, can you change it to be named something like `global` or `systemWide`?

agaertner:
Done.

greaka (approved):

PR #111, opened 2019-08-26
upward direction for Dropdown control.

agaertner:
Small change.

Just a bool so a dropdown's direction can be upward.

For dropdowns at the bottom of ~~clipping bounds~~, bottom of spritescreen.

### DIFF
--- a/Blish HUD/Controls/Dropdown.cs
+++ b/Blish HUD/Controls/Dropdown.cs
@@ -22,8 +22,9 @@ namespace Blish_HUD.Controls {
             private DropdownPanel(Dropdown assocDropdown) {
                 _assocDropdown = assocDropdown;
 
-                this.Location = _assocDropdown.AbsoluteBounds.Location + new Point(0, assocDropdown.Height - 1);
-                this.Size = new Point(_assocDropdown.Width, _assocDropdown.Height * _assocDropdown.Items.Count);
+                this.Height = _assocDropdown.Height * _assocDropdown.Items.Count;
+                this.Location = _assocDropdown.AbsoluteBounds.Location + new Point(0, assocDropdown.UpwardDirection ? 1 - this.Height : assocDropdown.Height - 1);
+                this.Width = _assocDropdown.Width;
                 this.Parent = Graphics.SpriteScreen;
                 this.ZIndex = int.MaxValue;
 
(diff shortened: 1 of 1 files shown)


dlamkins:
Mmm I don't know if I like this method and I feel this is kind of weird.  I don't think I've ever seen a dropdown menu go up.  Usually you just move the window or scroll downwards.  Is this because you have a dropdown at the bottom of a container and the parent container is clipping it?

If so, we just need to set "ClipsBounds" to false on the menu itself that pops out so that it can show outside of parent containers.

dlamkins:
Oh wait, `DropdownPanel` is already put directly into the `SpriteScreen`, so the parent container can't clip it anyways: https://github.com/blish-hud/Blish-HUD/blob/4c419ae1d0481011189ea20d13abde954a0f950b/Blish%20HUD/Controls/Dropdown.cs#L27

Could you send me a screenshot of what you're experiencing where it is getting cut off?  Just seems a bit odd as I don't think I see Dropdown controls that go up very frequently.  Usually you just move the window up some.  I can't check at the moment, but does the GW2 Dropdown do this by chance?  That's not a requirement, of course, but if it does, it'll make it easier to see the behavior in action.

agaertner:
I had a dropdown put just at the bottom of the entire screen.

The UpwardDirection works perfectly fine tho. Why just not give it to us ? :D
https://i.imgur.com/JChnNyn.png
(Yellow bar is the Mastery Experience bar.)

It clipped out of the screen area with the default drop direction.

>  I don't think I've ever seen a dropdown menu go up.

Really? I did, alot. Probably when its at the bottom of the screen. Think Planet Coaster and sandbox games have some where you get a big selection of objects and the panel for it is at the screen's bottom side.

dlamkins:
Interesting.  Well that's a lot of items in one Dropdown.  Probably not the best way to put them, but I suppose support for this is valid.  If anything, though, it should be calculated automatically when it is first clicked on - not with a manual property.  That'd be tedious to manage the property manually.  The Tooltip handles this automatically: https://github.com/blish-hud/Blish-HUD/blob/4c419ae1d0481011189ea20d13abde954a0f950b/Blish%20HUD/Controls/Tooltip.cs#L69-L79 and we can do something similar for the Dropdown.

If it opens upwards, will we want the list items to reverse as well (top item is at bottom, etc.)?

dlamkins:
It you would like, I can implement it using that method and PR it for you tonight.  Just let me know.

agaertner:
Yeah why not. :)

dlamkins:
Closing in favor of #112.

PR #112, opened 2019-08-27
Upwards dropdown

dlamkins:
Supersedes #111 allow for the dropdown direction to be chosen dynamically and automatically.

If a Dropdown is opened too close to the bottom of the screen (where the panel would go beyond the bottom), the item menu opens upwards to ensure the contents of the menu can still be seen.

Additionally, the control was cleaned up some and documentation was added to the control.

Here it is in action: https://gfycat.com/shoddygargantuanbream

@TybaIt How does that look?  🙂

### DIFF
--- a/Blish HUD/Controls/Dropdown.cs
+++ b/Blish HUD/Controls/Dropdown.cs
@@ -7,13 +7,16 @@ using Microsoft.Xna.Framework.Graphics;
 using MonoGame.Extended.TextureAtlases;
 
 namespace Blish_HUD.Controls {
-    public class Dropdown:Control {
+    /// <summary>
+    /// Represents a Guild Wars 2 Dropdown control.
+    /// </summary>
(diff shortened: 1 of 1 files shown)


Blish HUD/Controls/Dropdown.cs
```diff
@@ -7,13 +7,16 @@
...
+    public class Dropdown : Control {
 
-        private class DropdownPanel:Control {
+        private class DropdownPanel : Control {
 
             private Dropdown _assocDropdown;
 
-            protected int _highlightedItem = -1;
+            private int _highlightedItem = -1;
             private int HighlightedItem {
```
agaertner, line 20: Private field with getter and setter? Is this because SetProperty() invalidates the layout?

agaertner (approved):
Looks good. :)

PR #115, opened 2019-08-28
Removal of window message, extrakeyinfo, mouse access.

agaertner:
For Musician and upcoming Inquest and Special Forces Module. 
More access to buttons of the mouse. No window message.


dlamkins (approved):
Looks good.

PR #121, opened 2019-09-03
Modified control invalidation for better control updates

dlamkins:
Added support for SuspendLayout and ResumeLayout on controls. Modified invalidation behavior for controls so that they can recalculate without needing a full update cycle to occur.

Also fixed a bug that prevented controls from changing in size if they were first set to 0 for either their Width or their Height.


greaka (approved):

nickcode122 (approved):

dlamkins:
~~I'll be putting in a minor change tonight to update some documentation that @greaka pointed out to be about the `Invalidate()` call on `Control`s.~~

I've decided that, since `Control`s need some better documentation in general on all properties, I'll be PRing new documentation where this will get covered as part of that.

PR #122, opened 2019-09-04
Added BackgroundTexture property to Panels.

dlamkins:
I've added a new `BackgroundTexture` property to the `Panel` control (so `FlowPanel` and `ViewContainer` will also receive this change).

If defined, the texture will be drawn to the bounds of the `Panel`.  This reduces the number of cases where a custom control is needed and instead a `Panel` control with a `BackgroundTexture` can be enough to cover many situations.


nickcode122 (approved):
o7

PR #129, opened 2019-09-11
Added "notification-distance" attribute

QuitarHero:
A notification will display for the user depending on how far away they are from a marker, defined by `notification-distance="#"`.


dlamkins (approved):
Looks great!  👍

Issue #123, opened 2019-09-04
Festival Service


greaka:
Copy Paste from #62:

> this is more of a request for discussion:
theres not really a way to predict exact timings of annual events
so what would you say if the marker module had an optional(!) service that it would call thats gets timings for events. we would have to manually insert those events (so every 2 months putting in some times), the festival service (because maybe some other module could benefit from it?) would either expose if it can reach the service at all (kinda like arcdps service currently) or default to every event active when the service is unreachable
that way we could expose markers to hide depending on festivals
> 
> ![example image](https://cdn.discordapp.com/attachments/534490472224391169/601637629221666856/unknown.png)
> 
> ref: https://discordapp.com/channels/531175899588984842/534490472224391169/601637629221666857

Since the discussion derailed into a new context system and this system is in place now, we can now talk about the actual festival service

dlamkins:
Updating here to indicate where things currently, the path I am currently taking, and a dump of some of my notes.

Archomeda mentioned using https://api.guildwars2.com/v2/achievements/groups/18DB115A-8637-4290-A636-821362A3C4A8 which appears to be our best path forward.

Ultimately, this returns a list of category IDs representing daily categories.  It appears that during festivals, this list will contain the IDs of the festival dailies as well.

Note:
- This group ID (GUID in the request above) represents the Dailies group.
- I have confirmed that it is currently showing the "Halloween Daily" (79) category within the current dailies.
- [A list of the currently possible daily options (throughout the year) is easily visible here, for reference.](https://en.gw2treasures.com/achievement#18DB115A-8637-4290-A636-821362A3C4A8)  I've included a list of these, with their festival names, below.
- There is a **Lunar New Year Dailies** (102) daily in the category that always shows and is not the real one.  The correct category is **Daily Lunar New Year** (201).  The invalid one contains no achievements.

Here are the festivals which provide category IDs in the dailies group:
- Halloween (79)
- Wintersday (98)
- Super Adventure Festival (162)
- Lunar New Year (201)
- Festival of the Four Winds (213)
- Dragon Bash (233)

[...]

PR #131, opened 2019-10-22
FestivalContext

dlamkins:
WIP: Added the FestivalContext which provides a list of the active festivals.

Biggest missing item is that Gw2Sharp does not currently support accessing the achievement/groups endpoint due to a bug (fixed here: https://github.com/Archomeda/Gw2Sharp/issues/22) and an inconsistency in the API (see: https://github.com/arenanet/api-cdi/issues/657).  We'll have to update our Gw2Sharp dependency (which will be its own mini project).

This PR aims to satisfy #123.


Blish HUD/GameServices/Contexts/FestivalContext.cs
```diff
@@ -0,0 +1,156 @@
...
+            { 98,  Festival.Wintersday},
+            { 162, Festival.SuperAdventureFestival },
+            { 201, Festival.LunarNewYear },
+            { 213, Festival.FestivalOfTheFourWinds },
+            { 233, Festival.DragonBash }
+        };
+
+        private CancellationTokenSource _contextLoadCancellationTokenSource;
+
+        private List<Festival> _activeFestivals = new List<Festival>();
```
greaka, line 102: while it is technically possible that multiple festivals are active at once, theres no way anet will ever do this. for one, the decorations in cities would probably clip and they also stated that festivals are fillers between content releases. so i think they will never activate multiple festivals at the same time. 

what are your thoughts on providing just a single `_activeFestival`?

dlamkins: I do agree that is is particularly unlikely, but with the increased number of "events" happening lately, I worry that "festivals" may begin to include some smaller events that potentially overlap.  I also get that it's less convenient not being able to check for just an individual festival.  Do the two new functions provided in my latest commit make you feel better about this?

greaka: better :)

greaka (dismissed):

dlamkins:
Currently waiting for #172

dlamkins:
#172 has been merged so this is now ready once it is reviewed again.  :)

nickcode122 (approved):

PR #141, opened 2019-11-08
WIP: MVP implementations & wiring

dlamkins:
Currently a large WIP that needs a good amount of work.  Putting up as a draft so that we can start playing with builds.

As things start going slowly, I'll begin breaking the PR into chunks to have them added in.

Additionally, each "shared" view will need to have an explanation as to what purpose it serves, how to use it, and images showing what to expect it to look like.

I'll open up an issue in the morning where issues with the PR can be discussed.

Shared utility and layout views are being placed in the `Blish_HUD.Common.UI` namespace.


greaka:
that was an accident

greaka (approved):

PR #147, opened 2019-12-03
InputService Cleanup + KeyBinding Support

dlamkins:
I've broken this PR into a few stages to help make it a bit easier to review.  Please let me know if you have any questions or concerns.

29e3768
- Removed some classes related to the previous hook implementation.
- Removed the `HotkeysService`.
- There is now an `InputManager` which provides a basic method of starting an input hook thread of a designated type.
- `MouseManager` and `KeyboardManager` define the hook processing callback and manage/fire input events based on queued input.
- `InputService` now contains mostly just a reference to each of the managers (`GameService.Input.Keyboard` & `GameService.Input.Mouse`).

7ba030b
- Removed `NOMOUSEHOOK` from the default list of defined compiler constants.
- Updated all of the primary `Control`s that needed to have signatures and more updated to match the new `InputService` implementation.

bb5425e
- Added support for `KeyBinding`s.
- KeyBindings, if enabled, will listen for their combination to be pressed and will fire an event when that combination is completed.
- KeyBindings are implemented similar to how GW2 does them where they are a combination of a modifier key (Control, Shift, or Alt) and a primary key (and other key).
- KeyBindings can have a modifier key as the primary key if the modifier key is set to `None`.
- KeyBindings don't currently support blocking the last key press in the combination, as that was not necessary for what is needed at this time.  Support for blocking or "exclusive" keybindings will come down the road.

3094d8b
- Converted some of the code for the `HotkeyAssignmentWindow` and `HotkeyAssigner` controls to reference `KeyBinding` instead of `Hotkey` and commented out some code that would need significant reworks.
- Since these will likely both get replaced with `View`s, much of this will be replaced anyways.  

309ea7d
- Created a new static class containing the beginnings of a reference to the built-in keybindings for other modules and services to listen to.
- Currently just has "Interact/Action key."
- Can be found under `Common.Gw2.KeyBindings` (e.g. `Common.Gw2.KeyBindings.Interact`).
- Let me know if you feel there is a better place for this.

This PR allows for me to move forward with the TacO behavior implementation.

### BREAKING CHANGES
The Killproof module referenced an undocumented field on the previous implementation of the `InputService` which caused the hook to block in the callback.  This field was affectively noop'd to resolve the issue previously, but this new `InputService` no longer exposes this field and will cause the Killproof module to crash when loaded.  If approved, I'll need to get a PR in the Killproof module to ensure this reference is removed.


Blish HUD/GameServices/Input/KeyboardManager.cs
```diff
@@ -0,0 +1,100 @@
...
+            _inputBuffer = new ConcurrentQueue<KeyboardEventArgs>();
+        }
+
+        internal override void Update() {
+            while (_inputBuffer.TryDequeue(out var keyboardEvent)) {
+                _keysDown.Remove(keyboardEvent.Key);
+
+                if (keyboardEvent.EventType == KeyboardEventType.KeyDown) {
+                    _keysDown.Add(keyboardEvent.Key);
+                }
```
greaka, line 74: I'm a bit confused about this one

Blish HUD/GameServices/Input/KeysUtil.cs
```diff
@@ -0,0 +1,142 @@
...
+                {Keys.NumPad1, "1 (NUM)"},
+                {Keys.NumPad2, "2 (NUM)"},
+                {Keys.NumPad3, "3 (NUM)"},
+                {Keys.NumPad4, "4 (NUM)"},
+                {Keys.NumPad5, "5 (NUM)"},
+                {Keys.NumPad6, "6 (NUM)"},
+                {Keys.NumPad7, "7 (NUM)"},
+                {Keys.NumPad8, "8 (NUM)"},
+                {Keys.NumPad9, "9 (NUM)"},
+                {Keys.OemClear, "5"},
```
greaka: Did you intend this? Why 5?

Blish HUD/Controls/HotkeyAssigner.cs
```diff
@@ -112,7 +112,7 @@ public class HotkeyAssigner : Control {
                             );
 
             // Easy way to get a string representation of the hotkeys
-            string hotkeyRep = string.Join(" + ", _hotkeyDefinition.Keys);
+            string hotkeyRep = string.Join(" + ", _keyBindingDefinition.PrimaryKey);
```
greaka: shouldnt this be `string hotkeyRep =_keyBindingDefinition.GetBindingDisplayText()`?

greaka (changes requested):
In addition, I would like to bump the minor version with this pr, so we easily see if ppl run the breaking change when receiving logs.

Blish HUD/GameServices/Input/KeyboardManager.cs
greaka: the preview here is not showing everything. you remove something and then add it again shortly after

greaka:
I am not sure how much stuff gets reworked anyway when you move over to Views, so I just marked everything that I found and wasn't commented out.

Blish HUD/GameServices/Input/KeyboardManager.cs
dlamkins: [Link to Discord discussion](https://discordapp.com/channels/531175899588984842/536970543736291346/653335593635282945)

Summary: my current implementation is a bit hard to follow.  Benchmarks show there is also a negligible impact on performance (0.572μs processing 242 simulated key state changes) between this and the alternative.  In the interesting of making things easier to maintain, I'll be swapping this over to use a full if/else instead of relying on the conditional inside of the lists `Remove` function.

Blish HUD/GameServices/Input/KeysUtil.cs
dlamkins: [Link to Discord discussion](https://discordapp.com/channels/531175899588984842/536970543736291346/653347964177219632)

Summary: It should probably just be "Clear" or left without a definition.

Blish HUD/Controls/HotkeyAssigner.cs
dlamkins: Yes, you're right - good catch!

I think I'm going to go ahead and update the PR to not include these controls at all since they aren't working correctly anyways and are waiting on the changes introduced in the mvp-ui branch.  I can reintroduce them at that time.

dlamkins:
Still WIP while I do some further benchmarking on the keybind lookups.

greaka (dismissed):

dlamkins:
Found that right-click events don't appear to be working as expected.  I expect this is likely a mixup when differentiating right-clicks for the camera and those that are made within UI elements.  `_cameraDragging` (which causes `Update()` to exit early) is likely getting set to `true` for all right-clicks instead of just for those outside of the UI.

Blish HUD/GameServices/Input/MouseManager.cs
```diff
@@ -0,0 +1,189 @@
...
+
+            if (_cameraDragging && newEvent.EventType == MouseEventType.RightMouseButtonReleased) {
+                _cameraDragging = false;
+            } else if (_hudFocused && !_hookOverride) {
+                _mouseEvent = newEvent;
+
+                #if !NOMOUSEHOOK
+                return newEvent.EventType != MouseEventType.LeftMouseButtonReleased;
+                #endif
+            } else if (newEvent.EventType == MouseEventType.RightMouseButtonPressed) {
```
dlamkins, line 168: Confirm this is set under the correct circumstances (see line below this):

`_cameraDragging = true;`

dlamkins: Actually, the `_cameraDragging` variable is being properly set.  The issue comes from the break statement on line 134 of MouseManager.cs which prevents the `RightMouseButtonPressed`, `RightMouseButtonReleased`, and `MouseWheelScrolled` events from being triggered when they've been intercepted by the mouse hook (i.e. when the mouse is over one of our UI elements).

dlamkins: This was resolved in f8f1e5d

dlamkins:
I'm going to work on the Killproof module now to remove the incompatible field reference now.

greaka (approved):

dlamkins:
> I'm going to work on the Killproof module now to remove the incompatible field reference now.

This has now been fixed in this copy of the module: https://github.com/blish-hud/KillProof-Module

I need to update it a bit to change how it mentions needing updates since right now it checks the old repo for the last build version and reports there is a new build available if the version number is _different_.

dlamkins:
> > I'm going to work on the Killproof module now to remove the incompatible field reference now.
> 
> This has now been fixed in this copy of the module: https://github.com/blish-hud/KillProof-Module
> 
> I need to update it a bit to change how it mentions needing updates since right now it checks the old repo for the last build version and reports there is a new build available if the version number is _different_.

https://github.com/blish-hud/KillProof-Module/commit/5abb333150644c4a637a7f96edc6fb4419b9215c now reflects these needed changes in [Release v1.4.3](https://github.com/blish-hud/KillProof-Module/releases/tag/v1.4.3%2B9).

Issue #169, opened 2020-03-13
Make user settings path an option.


BlishHUD:
**Bennieboj at 3/13/2020 8:15:42 PM +00:00**
> Is it possible to let BHUD use a folder inside the application folder (instead of %USERPROFILE%\Documents\Guild Wars 2\addons\blishhud), thus making it truly self contained?

REF: https://discordapp.com/channels/531175899588984842/534490472224391169/688118110368170007

dlamkins:
Implemented in #171 

`& ".\Blish HUD.exe" --settings .\`

PR #171, opened 2020-03-16
Added support for launch parameters.

dlamkins:
Likely more supported parameters to come in the future.

Added support for launch parameters including:

- ProcessName(process/p)
- WindowName(window/w)
- UserSettingsPath (settings/s)
- RefPath (ref/r)

Wiki document provides details and example usage: https://github.com/blish-hud/Blish-HUD/wiki/Launch-Options

Implements a solution for #169.


greaka (approved):
never worked with entrypoint, but it looks good and simple

Issue #145, opened 2019-11-30
fieldOfView <= 0 or >= PI


BlishHUD:
**Ni at 11/30/2019 3:39:22 PM +00:00**
> Well yeah.. it crashed 15 hours ago :x
> ```01:18:45.8048|FATAL|Blish_HUD.DebugService|Blish HUD encountered a fatal crash!
> System.ArgumentException: fieldOfView <= 0 or >= PI
>    at Microsoft.Xna.Framework.Matrix.CreatePerspectiveFieldOfView(Single fieldOfView, Single aspectRatio, Single nearPlaneDistance, Single farPlaneDistance, Matrix& result)
>    at Blish_HUD.CameraService.Update(GameTime gameTime) in E:\C#\BlishHUD\Blish HUD\GameServices\CameraService.cs:line 34
>    at Blish_HUD.BlishHud.Update(GameTime gameTime) in E:\C#\BlishHUD\Blish HUD\BlishHud.cs:line 128
>    at Microsoft.Xna.Framework.Game.DoUpdate(GameTime gameTime)
>    at Microsoft.Xna.Framework.Game.Tick()
>    at MonoGame.Framework.WinFormsGameWindow.RunLoop()
>    at Microsoft.Xna.Framework.Game.Run(GameRunBehavior runBehavior)
>    at Blish_HUD.Program.Main() in E:\C#\BlishHUD\Blish HUD\Program.cs:line 45```

REF: https://discordapp.com/channels/531175899588984842/534492173362528287/650360233939566622

dlamkins:
Fixed in #172!

FieldOfView is now clamped when read from the MumbleLink API.

PR #172, opened 2020-03-25
Implement Latest Gw2Sharp (0.9.3)

dlamkins:
This PR works to reimplement Gw2Sharp to utilize it for the Mumble service and to update our WebApi service utilization.

Breaking:
- Camera and Player service have been removed and fields have been restructured under the Mumble service itself.
- Fields on the WebApi service have been updated.
- UiSize is no longer exposed on the GraphicsService and is instead listed under the Mumble services' UI field.
- Gw2ApiManager exposes the root WebApi client instead of V2 directly now.
- ChatCode util was removed under the `Common.Gw2` namespace since Gw2Sharp supports generating and parsing these directly.


greaka:
since this is some pretty big milestone, and also includes some breaking changes, id like to propose bumping the version to 0.5 with this PR

dlamkins:
> since this is some pretty big milestone, and also includes some breaking changes, id like to propose bumping the version to 0.5 with this PR

Agreed!  🙂 

Also noting here that we'll need to retarget again to the new 9.3 release.  Doesn't appear to have anything breaking coming from the 9.1 or 9.2 we're currently targeting.

https://github.com/Archomeda/Gw2Sharp/releases/tag/v0.9.3

Blish HUD/GameServices/ContentService.cs
```diff
@@ -108,6 +109,12 @@ public enum FontStyle {
 
         public Microsoft.Xna.Framework.Content.ContentManager ContentManager => Blish_HUD.BlishHud.ActiveContentManager;
 
+        private ICacheMethod _sharedRenderServiceCache;
+
+        public ICacheMethod GetRenderServiceCacheMethod() {
+            return _sharedRenderServiceCache ?? (_sharedRenderServiceCache = new MemoryCacheMethod());
```
greaka, line 115: Why don't we use the ArchiveCacheMethod here?

Blish HUD/GameServices/Gw2Mumble/PlayerCamera.cs
```diff
@@ -0,0 +1,69 @@
...
+
+        /// <summary>
+        /// The closest distance entities are rendered.
+        /// </summary>
+        public float NearPlaneRenderDistance { get; set; } = 0.01f;
+
+        /// <summary>
+        /// The farthest distance entities are rendered.
+        /// </summary>
+        public float FarPlaneRenderDistance  { get; set; } = 1000.0f;
```
greaka: Do you want modules to be able to set this? Maybe we want an internal set and have it exposed somewhere if needed.

Blish HUD/GameServices/Gw2Mumble/PlayerCamera.cs
```diff
@@ -0,0 +1,69 @@
...
+        #region Mumble Proxied Fields
+
+        /// <inheritdoc cref="IGw2MumbleClient.CameraPosition"/>
+        public Vector3 Position => _position;
+
+        /// <inheritdoc cref="IGw2MumbleClient.CameraFront"/>
+        public Vector3 Forward => _forward;
+
+        /// <inheritdoc cref="IGw2MumbleClient.FieldOfView"/>
+        public float FieldOfView => MathHelper.Clamp(_fieldOfView, 0.01f, (float)Math.PI - 0.01f);
```
greaka: Put the clamp into the Update function, so it doesn't need to be calculated whenever someone reads it. We need it at least once anyway, so we can only gain here.

Blish HUD/GameServices/Gw2WebApi/ManagedConnection.cs
```diff
@@ -0,0 +1,58 @@
...
+        private readonly Connection _internalConnection;
+
+        public IConnection Connection => _internalConnection;
+
+        private readonly IGw2WebApiClient _internalClient;
+
+        public IGw2WebApiClient Client => _internalClient;
+
+        public ManagedConnection(string accessToken, ICacheMethod webApiCache, ICacheMethod renderCache = null, TimeSpan? renderCacheDuration = null) {
+            string ua = $"BlishHUD/{Program.OverlayVersion} ";
```
greaka: I think there is an excess whitespace in there, or is it intentional? Why?

Blish HUD/GameServices/Modules/Managers/Gw2ApiManager.cs
```diff
@@ -1,39 +1,53 @@
...
 using Gw2Sharp.WebApi.V2.Models;
 
 namespace Blish_HUD.Modules.Managers {
     public class Gw2ApiManager {
 
-        private readonly HashSet<TokenPermission> _permissions;
+        private const int SUBTOKEN_LIFETIME = 7;
 
-        private Connection _gw2ApiConnection;
+        private static readonly List<Gw2ApiManager> _apiManagers = new List<Gw2ApiManager>();
```
greaka, line 14: I would like to propose that we use another naming convention for static fields, so it is clear in the code.

Blish HUD/GameServices/Modules/Managers/Gw2ApiManager.cs
```diff
@@ -1,39 +1,53 @@
...
 
-        private Gw2WebApiClient _gw2ApiClient;
+        internal static void RenewAllSubtokens() {
+            foreach (var apiManager in _apiManagers.ToArray()) {
```
greaka: You shouldn't need to call `.ToArray()` here.

greaka (changes requested):

greaka:
This PR is also missing the version change to 0.5, I think it should be changed before we merge it.

Blish HUD/GameServices/Gw2WebApi/ManagedConnection.cs
dlamkins: This was due to this:
https://discordapp.com/channels/384735285197537290/589031897963692034/692548369742168084

> So taking a closer look at the UA appending that is done, a space is not placed in between.  This causes the UA format to be incorrect (I believe due to having multiple / without being split by a  ).  Appending my provided UA with a space allows it to work.

It appears that this was fixed in 9.2, though, so I'll get that extra space removed. :)

https://github.com/Archomeda/Gw2Sharp/commit/022d7296cc3dc64c883116105987a626136edc74

Blish HUD/GameServices/Modules/Managers/Gw2ApiManager.cs
dlamkins: Currently we're using _camelCase for all non-public fields.  Naming convention here is indicating public vs. not public, which is fairly common from what I can tell.

We can certainly discuss potential alternatives, but I don't think this PR should be the place for that discussion or ground zero for those changes.

Blish HUD/GameServices/Modules/Managers/Gw2ApiManager.cs
dlamkins: Good catch.

dlamkins:
FYI - CI has the final say on the version which is why the assembly version itself hasn't been updated from 4.0 in quite a while.  It'll inject and override whatever is specified in the project file with the incrementing build value included.

That being said, I've gone ahead and upped the project's assembly version as requested.

Blish HUD/GameServices/ContentService.cs
dlamkins: I've had a mixed experience with `ArchiveCacheMethod` in the past and troubles getting it to flush to the disk correctly causing it to not keep a cache of things requested that session until the session ends and it flushes to disk.  This causes all requests to get re-requested every request.  There were recent changes to it after some other folks brought up having similar issues, so I'll need to review those.

Ultimately, I need more time to review the current implementation in Gw2Sharp to ensure it works correctly when we use it and will have a PR to update this to an `ArchiveCacheMethod` once I have done that.  Using the `MemoryCacheMethod` for now will at least cover caching for the active session until that work is done.

Blish HUD/GameServices/Gw2Mumble/PlayerCamera.cs
dlamkins: I personally have no preference either way.  It was a holdover from the original `CameraService` and was kind of left like that in case any modules wanted to play with them.  

I can't think of any strong use cases for doing that, though, so for now I've removed the public setters on them and if we ever want to make them accessible down the road we can add them back or provide some other method of exposing them without causing a breaking change.

Blish HUD/GameServices/Modules/Managers/Gw2ApiManager.cs
greaka: agreed

greaka (dismissed):

dlamkins:
Sorry for the re-review.  I removed the, now redundant, ChatCode utility as Gw2Sharp can now take care of that.

greaka (approved):

Blish HUD/GameServices/Gw2Mumble/PlayerCamera.cs
```diff
@@ -0,0 +1,69 @@
...
+        #endregion
+
+        internal PlayerCamera(Gw2MumbleService service) {
+            _service = service;
+        }
+
+        internal void Update(GameTime gameTime) {
+            _position    = _service.RawClient.CameraPosition.ToXnaVector3();
+            _forward     = _service.RawClient.CameraFront.ToXnaVector3();
+            _fieldOfView = MathHelper.Clamp((float)_service.RawClient.FieldOfView, 0.01f, (float)Math.PI - 0.01f);
```
dlamkins, line 58: Fixes #145!

PR #176, opened 2020-03-27
Reimplemented KeybindingAssigner and KeybindingAssignerWindow controls + KeyboardManager fixes

dlamkins:
Force keys to depress when the keyboard hook is disabled to prevent stuck keys when the game gains and loses focus.  Prevented keys being held down from filling the _keysDown list and from firing keydown when the key was already down.

Added GetFriendlyName with ModifierKeys and Keys params for full KeyBinding display strings which could be referenced both by KeyBindings and the new KeybindingAssinger control.

Added the KeybindingAssigner control back into the repo which allows you to tie a KeyBinding to it to then manipulate it.  When double-clicked, a KeybindingAssignmentWindow is shown which allows the user to update the keybinding.

The new controls can be seen in action, here: https://gfycat.com/smoothpreciousbandicoot

### DIFF
--- a/Blish HUD/Blish HUD.csproj
+++ b/Blish HUD/Blish HUD.csproj
@@ -93,6 +93,8 @@
   <ItemGroup>
     <Compile Include="Common\Gw2\ChatCode.cs" />
     <Compile Include="Common\Gw2\KeyBindings.cs" />
+    <Compile Include="Controls\KeybindingAssignmentWindow.cs" />
+    <Compile Include="Controls\KeybindingAssigner.cs" />
     <Compile Include="Controls\Intern\Mouse.cs" />
     <Compile Include="Controls\ViewContainer.cs" />
     <Compile Include="Controls\_Types\ViewState.cs" />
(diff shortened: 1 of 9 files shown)


Denrage:
Is DoubleClick not something that can be usefull to have in a base class?
Now every control who wants to react on a double click needs to implement it itself and it's not guaranteed that every implementation has the same timing

dlamkins:
@Denrage if you can name any other controls that may need it, it may be worth considering to do, but at this time, this is the only control that I am aware of that requires a double-click to activate.

greaka (commented):
looks good, more in discord: https://discordapp.com/channels/531175899588984842/536970543736291346/693078606947024907 onward

agaertner (dismissed):

dlamkins:
In summary of the discussion in Discord regarding moving double-click support up to the `Control` class:
Per @Denrage and @greaka's request, I have implemented `IsDoubleClick` on `MouseEventArgs` and this is now passed to the `Click` event when necessary and supported at the `Control` level for all controls.

greaka (approved):

PR #177, opened 2020-03-27
Reimplement ColorBox + ColorPicker

Denrage:
Reimplemented ColorBox and ColorPicker. Appearance/Functionality should be the same, just don't know how it behaves previously.

Breaking Changes:

- ColorId on ColorBox is now readonly and the consumer of the ColorBox needs to provide the Gw2Sharp Color-Object for the box. This removes the need of an api call.

Here the control(s) in action: https://drive.google.com/uc?id=1uZsMH836PYzjTTy2X-dWcP4ULg-pmQkO
The position of the cursor seems off bc of YoloMouse

### DIFF
--- a/Blish HUD/Controls/ColorBox.cs
+++ b/Blish HUD/Controls/ColorBox.cs
@@ -20,13 +20,13 @@ namespace Blish_HUD.Controls {
         public event EventHandler<EventArgs> OnColorChanged;
         public event EventHandler<EventArgs> OnSelected;
 
-        private const int DEFAULT_COLOR_SIZE = 32;
-        private const string COLOR_CHANGE_SOUND_NAME = "audio/color-change";
-        private const string DRAW_VARIATION_VERSION_ONE_NAME = "colorpicker/cp-clr-v1";
-        private const string DRAW_VARIATION_VERSION_TWO_NAME = "colorpicker/cp-clr-v2";
+        private const int    DEFAULT_COLOR_SIZE                = 32;
+        private const string COLOR_CHANGE_SOUND_NAME           = "audio/color-change";
+        private const string DRAW_VARIATION_VERSION_ONE_NAME   = "colorpicker/cp-clr-v1";
+        private const string DRAW_VARIATION_VERSION_TWO_NAME   = "colorpicker/cp-clr-v2";
         private const string DRAW_VARIATION_VERSION_THREE_NAME = "colorpicker/cp-clr-v3";
-        private const string DRAW_VARIATION_VERSION_FOUR_NAME = "colorpicker/cp-clr-v4";
-        private const string HIGHLIGHT_NAME = "colorpicker/cp-clr-active";
+        private const string DRAW_VARIATION_VERSION_FOUR_NAME  = "colorpicker/cp-clr-v4";
+        private const string HIGHLIGHT_NAME                    = "colorpicker/cp-clr-active";
 
         private readonly int drawVariation;
 
@@ -38,8 +38,7 @@ namespace Blish_HUD.Controls {
                 if (SetProperty(ref selected, value)) {
                     this.OnSelected?.Invoke(this, EventArgs.Empty);
 
-                    if (this.Visible)
-                        Content.PlaySoundEffectByName(COLOR_CHANGE_SOUND_NAME);
+                    if (this.Visible) Content.PlaySoundEffectByName(COLOR_CHANGE_SOUND_NAME);
                 }
             }
         }
@@ -49,8 +48,7 @@ namespace Blish_HUD.Controls {
         public Gw2Sharp.WebApi.V2.Models.Color Color {
             get => color;
             set {
-                if (color == value)
-                    return;
+                if (color == value) return;
 
                 color = value;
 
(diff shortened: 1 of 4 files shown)


Blish HUD/Controls/ColorBox.cs
```diff
@@ -1,114 +1,111 @@
...
-        //    if (this.MouseOver || this.Selected)
-        //        spriteBatch.Draw(spriteHighlight, bounds, Microsoft.Xna.Framework.Color.White * 0.7f);
-        //}
-
-    //}
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public class ColorBox : Control {
+
+        public event EventHandler<EventArgs> OnColorChanged;
+        public event EventHandler<EventArgs> OnSelected;
```
dlamkins: The TODO above is actually about switching this these to use the style `ColorChanged` and `Selected` instead of using the 'On' prefix before them.  Can knock the TODO out and get this swapped in one fix.

Blish HUD/Controls/ColorBox.cs
```diff
@@ -1,114 +1,111 @@
...
+        public Gw2Sharp.WebApi.V2.Models.Color Color {
+            get => color;
+            set {
+                if (color == value)
+                    return;
+
+                color = value;
+
+                OnPropertyChanged(nameof(this.ColorId));
+                OnPropertyChanged();
```
dlamkins: I think this duplicate is a relic of old code that was commented out at different times.  This can all be replaced with:
```
set {
	if (SetProperty(ref color, value)) {
		this.ColorChanged?.Invoke(this, EventArgs.Empty);
	}
}
```

Blish HUD/Controls/ColorBox.cs
```diff
@@ -1,114 +1,111 @@
...
+        }
+
+        public int ColorId => this.Color?.Id ?? -1;
+
+        private static TextureRegion2D[] _possibleDrawVariations;
+        private static TextureRegion2D _spriteHighlight;
+
+        private static void LoadStatics() {
+            if (_possibleDrawVariations != null)
+                return;
```
dlamkins: This shouldn't be possible since it is set in this call and the static constructor will never get called again anyways.  I think it's another relic of my old code.

Blish HUD/Controls/ColorPicker.cs
```diff
@@ -10,106 +10,120 @@ namespace Blish_HUD.Controls {
 
     // TODO: The ColorPicker needs updates for events, it should probably inherit from FlowPanel,
     // and needs to get reconnected once we have Gorrik.NET included in the project
```
dlamkins: No longer needed and can be cleaned up with the work you've done here.  :)

Blish HUD/Controls/ColorPicker.cs
```diff
@@ -10,106 +10,120 @@ namespace Blish_HUD.Controls {
...
+        private readonly Dictionary<Gw2Sharp.WebApi.V2.Models.Color, ColorBox> colorBoxes;
+
+        private int colorsPerRow;
+
+        private Gw2Sharp.WebApi.V2.Models.Color selectedColor;
+        public Gw2Sharp.WebApi.V2.Models.Color SelectedColor {
+            get => selectedColor;
+            protected set {
+                if (selectedColor != value) {
+                    selectedColor = value;
```
dlamkins: Equality check and assignment can be done with `SetProperty` again.

Blish HUD/Controls/ColorPicker.cs
```diff
@@ -10,106 +10,120 @@ namespace Blish_HUD.Controls {
...
+
+        private ColorBox associatedColorBox;
+        public ColorBox AssociatedColorBox {
+            get => associatedColorBox;
+            set {
+                if (associatedColorBox != value) {
+                    if (associatedColorBox != null)
+                        associatedColorBox.Selected = false;
+
+                    associatedColorBox = value;
```
dlamkins: I believe can be done with `SetProperty` again.

Blish HUD/Controls/ColorPicker.cs
```diff
@@ -10,106 +10,120 @@ namespace Blish_HUD.Controls {
...
+                int verticalPosition = i / colorsPerRow;
+
+                currentBox.Location = new Point(
+                    horizontalPosition * (currentBox.Width + COLOR_PADDING),
+                    verticalPosition * (currentBox.Width + COLOR_PADDING)
+                );
+            }
+        }
+
+        public override void Invalidate() {
```
dlamkins: I believe this can get replaced with an override to `RecalculateLayout()`.  Nearly the same thing, but was introduced to let `Invalidate` handle some internal state stuff.  Technically fine since `Invalidate()` is called below.

Blish HUD/Controls/ColorPicker.cs
```diff
@@ -10,106 +10,120 @@ namespace Blish_HUD.Controls {
...
+            base.Invalidate();
+
+            colorsPerRow = (this.Width -10 ) / (COLOR_SIZE + COLOR_PADDING);
+            this.ContentRegion = new Rectangle(COLOR_PADDING, COLOR_PADDING, (this.Width - 10) - (COLOR_PADDING * 2), this.Height - (COLOR_PADDING * 2));
+        }
+
+        public override void PaintBeforeChildren(SpriteBatch spriteBatch, Rectangle bounds) {
+            // Draw background
+            spriteBatch.DrawOnCtrl(this, Content.GetTexture(BACKGROUND_TEXTURE_NAME), bounds, Color.Black * 0.5f);
```
dlamkins: The content service does cache the textures, but we no longer load textures in the draw loop where we can help it.  I'd assign it somewhere statically as we did with the `ColorBox` control and reference that.  This was just still around from my old code before we made the transition to non-cached UI renders and it didn't matter.

Blish HUD/_Extensions/ColorExtensions.cs
```diff
@@ -0,0 +1,17 @@
...
+namespace Blish_HUD._Extensions
+{
+    public static class ColorExtensions
+    {
+        public static Microsoft.Xna.Framework.Color ToXnaColor(this Gw2Sharp.WebApi.V2.Models.ColorMaterial color)
+        {
+            return  new Color(color.Rgb[0], color.Rgb[1], color.Rgb[2]);
+        }
+    }
+}
```
dlamkins: Would prefer K&R style for consistency.  Sorry for the editorconfig not being the right one!

dlamkins (changes requested):
Just a couple of small requests.  Love the work you were able to put on this. 😄

Blish HUD/Controls/ColorBox.cs
Denrage: It's not in a static constructor. It gets called everytime a ColorBox gets initiated. I could change the `LoadStatic` to a static contructor though

Blish HUD/Controls/ColorBox.cs
Denrage: But `OnPropertyChanged(nameof(this.ColorId));` still needs to be called or not?

dlamkins: Oh, `SetProperty` actually builds that in.

https://github.com/blish-hud/Blish-HUD/blob/e8d9df02f71bfdcd2d8fa118b353138ce1db9889/Blish%20HUD/Controls/Control.cs#L802-L810

Blish HUD/Controls/ColorBox.cs
dlamkins: Oh, you're absolutely right.  I'm so used to them all being in a static ctor now that I didn't even notice.

Blish HUD/Controls/ColorBox.cs
dlamkins: As discussed in Discord, this was for the `ColorId` property if I had been paying attention.  We should be good to actually fully remove that property as it's just redundant.

dlamkins (approved):

dlamkins:
Thanks!

PR #178, opened 2020-03-28
Add BindingChanged event for KeybindingAssigner.

dlamkins:
Adds event `BindingChanged` for when a binding is updated through the `KeybindingAssigner`.  Without this there is not a good way to ensure settings can be saved once the keybinding is updated by the control.


agaertner (approved):
🚂choo choo

PR #180, opened 2020-03-29
Textbox Redux 🎉

dlamkins:
~Drop-in~ Breaking replacement for Textbox proxy control that was used before that doesn't require the significant number of hacks we were using previously.

https://discordapp.com/channels/531175899588984842/536970543736291346/693694521971703808

### DIFF
--- a/Blish HUD/Controls/MultilineTextBox.cs
+++ b/Blish HUD/Controls/MultilineTextBox.cs
@@ -1,0 +1,109 @@
+﻿using System;
+using System.Linq;
+using Blish_HUD.Input;
+using Microsoft.Xna.Framework;
+using Microsoft.Xna.Framework.Graphics;
+
+namespace Blish_HUD.Controls {
+    public class MultilineTextBox : TextInputBase {
+
+        private const int PADDING = 2;
+
+        private const int TEXT_TOPPADDING  = 2;
+        private const int TEXT_LEFTPADDING = 10;
+
+        #region Load Static
+
+        private static readonly Texture2D _textureTextbox;
+
+        static MultilineTextBox() {
+            _textureTextbox = Content.GetTexture("textbox");
+        }
+
+        #endregion
+
+        private int _prevCursorIndex  = 0;
+        private int _horizontalOffset = 0;
+
+        public MultilineTextBox() {
+            _multiline = true;
+        }
+
+        protected override void UpdateScrolling() {
+            
+        }
+
+        public override void DoUpdate(GameTime gameTime) {
+            if (_cursorMoved) {
+
+            }
+
+            base.DoUpdate(gameTime);
+        }
+
+        protected override void Paint(SpriteBatch spriteBatch, Rectangle bounds) {
+            spriteBatch.DrawOnCtrl(this,
+                                   _textureTextbox,
+                                   new Rectangle(Point.Zero, _size - new Point(5, 0)),
+                                   new Rectangle(0,          0, Math.Min(_textureTextbox.Width - 5, _size.X - 5), _textureTextbox.Height));
+
+            spriteBatch.DrawOnCtrl(this, _textureTextbox,
+                                   new Rectangle(_size.X - 5, 0, 5, _size.Y),
+                                   new Rectangle(_textureTextbox.Width - 5, 0,
+                                                 5, _textureTextbox.Height));
+
+            var textBounds = new Rectangle(TEXT_LEFTPADDING - _horizontalOffset,
+                                           TEXT_TOPPADDING,
+                                           _size.X - TEXT_LEFTPADDING * 2,
+                                           _size.Y - TEXT_TOPPADDING  * 2);
+
+            // Draw the Textbox placeholder text
+            if (!_focused && _text.Length == 0) {
+                spriteBatch.DrawStringOnCtrl(this, _placeholderText, _font, textBounds, Color.LightGray, false, false, 0, HorizontalAlignment.Left, VerticalAlignment.Top);
+            }
+
+            // Draw the Textbox text
+            spriteBatch.DrawStringOnCtrl(this, this.Text, _font, textBounds, _foreColor, false, false, 0, HorizontalAlignment.Left, VerticalAlignment.Top);
+
+            int selectionStart  = Math.Min(_selectionStart, _selectionEnd);
+            int selectionLength = Math.Abs(_selectionStart - _selectionEnd);
+
+            if (selectionLength > 0) {
+                float highlightLeftOffset = _font.MeasureString(_text.Substring(0, selectionStart)).Width + textBounds.Left;
+                float highlightWidth      = _font.MeasureString(_text.Substring(selectionStart, selectionLength)).Width;
+
+                spriteBatch.DrawOnCtrl(this,
+                                       ContentService.Textures.Pixel,
+                                       new Rectangle((int)highlightLeftOffset - 1, 3, (int)highlightWidth, _size.Y - 9),
+                                       new Color(92, 80, 103, 150));
+            } else if (_focused && _caretVisible) {
+                int lineIndex = 0;
+                int lineStart = 0;
+
+                for (int n = 0; n < _cursorIndex; n++) {
+                    if (_text[n] == NEWLINE) {
+                        lineIndex++;
+                        lineStart = n;
+                    };
+                }
+
+                var glyphs = _font.GetGlyphs(_text.Substring(lineStart, _cursorIndex - lineStart));
+
+                var offset = Point.Zero;
+
+                if (_cursorIndex > 0) {
+                    var last = glyphs.Last();
+
+                    offset = new Point((int)last.Position.X + (last.FontRegion?.Width ?? 0),
+                                       _font.LineHeight * lineIndex);
+                }
+
+                var   caretOffset = new Rectangle(textBounds.X + offset.X - 2,
+                                                  textBounds.Y + offset.Y,
+                                                  textBounds.Width,
+                                                  textBounds.Height);
+                spriteBatch.DrawStringOnCtrl(this, "|", _font, caretOffset, _foreColor, false, false, 0, HorizontalAlignment.Left, VerticalAlignment.Top);
+            }
+        }
+    }
+}
--- a/Blish HUD/Controls/MultilineTextBox.cs
+++ b/Blish HUD/Controls/MultilineTextBox.cs
@@ -29,8 +29,49 @@ namespace Blish_HUD.Controls {
             _multiline = true;
         }
 
+        protected override void OnClick(MouseEventArgs e) {
+            base.OnClick(e);
+
+            int newIndex = GetCursorIndexFromFromPosition(this.RelativeMousePosition - new Point(TEXT_LEFTPADDING, TEXT_TOPPADDING));
+
+            if (_cursorIndex == newIndex && e.IsDoubleClick) {
+                SelectAll();
+            } else {
+                UserSetCursorIndex(newIndex);
+                ResetSelection();
+            }
+        }
+
+        private int GetCursorIndexFromFromPosition(Point clickPos) {
+            string[] lines = _text.Split(NEWLINE);
+
+            int predictedLine = clickPos.Y / _font.LineHeight;
+
+            if (predictedLine > lines.Length - 1) {
+                return this.Length;
+            }
+
+            var glyphs = _font.GetGlyphs(lines[predictedLine]);
+
+            int charIndex = 0;
+
+            foreach (var glyph in glyphs) {
+                if (glyph.Position.X + glyph.FontRegion.Width > _horizontalOffset + clickPos.X) {
+                    break;
+                }
+
+                charIndex++;
+            }
+
+            for (int i = 0; i < predictedLine; i++) {
+                charIndex += lines[i].Length + 1;
+            }
+
+            return charIndex;
+        }
+
         protected override void UpdateScrolling() {
-            
+
         }
 
         public override void DoUpdate(GameTime gameTime) {
--- a/Blish HUD/Controls/MultilineTextBox.cs
+++ b/Blish HUD/Controls/MultilineTextBox.cs
@@ -55,17 +55,17 @@ namespace Blish_HUD.Controls {
         protected override void OnClick(MouseEventArgs e) {
             base.OnClick(e);
 
-            int newIndex = GetCursorIndexFromFromPosition(this.RelativeMousePosition - new Point(TEXT_LEFTPADDING, TEXT_TOPPADDING));
+            int newIndex = GetCursorIndexFromPosition(this.RelativeMousePosition - new Point(TEXT_LEFTPADDING, TEXT_TOPPADDING));
 
             if (_cursorIndex == newIndex && e.IsDoubleClick) {
                 SelectAll();
             } else {
                 UserSetCursorIndex(newIndex);
-                ResetSelection();
+                UpdateSelectionIfShiftDown();
             }
         }
 
-        private int GetCursorIndexFromFromPosition(Point clickPos) {
+        private int GetCursorIndexFromPosition(Point clickPos) {
             string[] lines = _text.Split(NEWLINE);
 
             int predictedLine = clickPos.Y / _font.LineHeight;
--- a/Blish HUD/Controls/MultilineTextBox.cs
+++ b/Blish HUD/Controls/MultilineTextBox.cs
@@ -66,7 +66,7 @@ namespace Blish_HUD.Controls {
             for (int i = 0; i < index; i++) {
                 charIndex++;
 
-                if (_text[lineIndex] == NEWLINE) {
+                if (_text[i] == NEWLINE) {
                     lineIndex++;
                     charIndex = 0;
                 }
@@ -76,7 +76,58 @@ namespace Blish_HUD.Controls {
         }
 
         private Rectangle[] CalculateHighlightRegions() {
-            return Array.Empty<Rectangle>();
+            int selectionStart  = Math.Min(_selectionStart, _selectionEnd);
+            int selectionLength = Math.Abs(_selectionStart - _selectionEnd);
+
+            if (selectionLength <= 0 || selectionStart + selectionLength > this.Length) return Array.Empty<Rectangle>();
+
+            string[] lines = _text.Split(NEWLINE);
+
+            var startIndex = GetSplitIndex(selectionStart);
+            var endIndex   = GetSplitIndex(selectionStart + selectionLength);
+
+            int lineSpans = endIndex.Line - startIndex.Line;
+
+            var regions = new Rectangle[lineSpans + 1];
+
+            if (lineSpans == 0) {
+                float highlightLeftOffset = MeasureStringWidth(lines[startIndex.Line].Substring(0, startIndex.Character));
+                float highlightWidth      = MeasureStringWidth(lines[startIndex.Line].Substring(startIndex.Character, selectionLength));
+
+                regions[0] = new Rectangle(_textRegion.Left + (int)highlightLeftOffset - 1,
+                                           _textRegion.Top + (startIndex.Line * _font.LineHeight),
+                                           (int)highlightWidth,
+                                           _font.LineHeight - 1);
+            } else {
+                // First line
+                float firstHighlightLeftOffset = MeasureStringWidth(lines[startIndex.Line].Substring(0, startIndex.Character));
+                float firstHighlightWidth      = MeasureStringWidth(lines[startIndex.Line].Substring(startIndex.Character));
+
+                regions[0] = new Rectangle(_textRegion.Left + (int) firstHighlightLeftOffset - 1,
+                                           _textRegion.Top  + (startIndex.Line * _font.LineHeight),
+                                           (int) firstHighlightWidth,
+                                           _font.LineHeight - 1);
+
+                // Middle lines
+                for (int i = startIndex.Line + 1; i < endIndex.Line; i++) {
+                    float fullWidth = MeasureStringWidth(lines[i]);
+
+                    regions[i - startIndex.Line] = new Rectangle(_textRegion.Left - 1,
+                                                                 _textRegion.Top  + (i * _font.LineHeight),
+                                                                 (int) fullWidth,
+                                                                 _font.LineHeight - 1);
+                }
+
+                // Last line
+                float lastHighlightWidth = MeasureStringWidth(lines[endIndex.Line].Substring(0, endIndex.Character));
+
+                regions[lineSpans] = new Rectangle(_textRegion.Left - 1,
+                                                   _textRegion.Top  + (endIndex.Line * _font.LineHeight),
+                                                   (int) lastHighlightWidth,
+                                                   _font.LineHeight - 1);
+            }
+
+            return regions;
         }
 
         private Rectangle CalculateTextRegion() {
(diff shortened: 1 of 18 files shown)


Blish HUD/Controls/TextInputBase.cs
```diff
@@ -0,0 +1,718 @@
...
+            }
+        }
+
+        private void DeleteChars(int index, int length) {
+            if (length <= 0) return;
+
+            SetText(_text.Substring(0, index) + _text.Substring(index + length), true);
+        }
+
+        private string RemoveUnsupportedCharacters(string value) {
```
Denrage: This could possibly result in a StackOverflow if value has a lot of different unsupported characters

```suggestion
        private string RemoveUnsupportedCharacters(string value){
    var result = new StringBuilder();
    foreach (char c in value) {
        (_font.GetCharacterRegion(c) != null || _multiline && c == NEWLINE) {
            result.Append(c);
        }
    }

    return result.ToString();
}
```

Blish HUD/Controls/TextInputBase.cs
```diff
@@ -0,0 +1,718 @@
...
+            foreach (char c in value) {
+                if (_font.GetCharacterRegion(c) == null && !(_multiline && c == NEWLINE)) {
+                    return RemoveUnsupportedCharacters(value.Replace(c.ToString(), string.Empty));
+                }
+            }
+
+            return value;
+        }
+
+        private bool InsertChars(int index, string value, out int length) {
```
Denrage, line 251: We have currently a mix of abbreviation and non-abbreviation "char" and "character". Would be better to have one of them for consistency
You could also call the plural of characters a string

Blish HUD/Controls/TextInputBase.cs
```diff
@@ -0,0 +1,718 @@
...
+        protected int _maxLength = int.MaxValue;
+
+        /// <summary>
+        /// Gets or sets the maximum character length of the control.
+        /// </summary>
+        public int MaxLength {
+            get => _maxLength;
+            set {
+                if (SetProperty(ref _maxLength, value)) {
+                    this.Text = _text.Substring(0, Math.Min(_maxLength, _text.Length));
```
Denrage, line 91: You have a property for "Text" for a reason. Don't know how i feel about it to access the backendfield directly instead using the getter through the property
(this also applies to every other use of a backendfield too)

Blish HUD/Controls/TextInputBase.cs
```diff
@@ -0,0 +1,718 @@
...
+            }
+
+            value = RemoveUnsupportedCharacters(value);
+
+            int startLength = _text.Length;
+
+            if (string.IsNullOrEmpty(_text)) {
+                SetText(value, true);
+            } else {
+                SetText(_text.Substring(0, index) + value + _text.Substring(index), true);
```
Denrage: There is https://docs.microsoft.com/en-us/dotnet/api/system.string.insert?view=netframework-4.8 for that

Blish HUD/Controls/TextInputBase.cs
```diff
@@ -0,0 +1,718 @@
...
+            }
+
+            length = _text.Length - startLength;
+            return true;
+        }
+
+        private bool InsertChar(int index, char value) {
```
Denrage, line 271: No check if Character is unsupported like the one with multiple Characters?

Blish HUD/Controls/TextInputBase.cs
```diff
@@ -0,0 +1,718 @@
...
+                        break;
+                }
+            } finally {
+                _suppressRedoStackReset = false;
+            }
+
+            ResetSelection();
+        }
+
+        protected void UserSetCursorIndex(int newIndex) {
```
Denrage, line 406: Exactly for that is a Setter of a property and i don't see a reason why it shouldn't be there also for the internal setters in this class. I mean this check should also be done if you are inserting characters or smth like that. The only benefit would be that it will be faster but that's a pretty non existing performance boost (premature optimization is the root of all evil and so on)

Blish HUD/Controls/_Types/UndoRedoStack.cs
```diff
@@ -0,0 +1,73 @@
...
+ */
+
+using System.Collections.Generic;
+
+namespace Blish_HUD.Controls {
+    internal class UndoRedoStack {
+
+        private readonly Stack<UndoRedoRecord> _stack = new Stack<UndoRedoRecord>();
+
+        public Stack<UndoRedoRecord> Stack => _stack;
```
Denrage: Maybe return a copy of the stack so the stack cannot be altered outside of this class (you could also just hope that no one in this project will do smth like that)

Blish HUD/Controls/TextInputBase.cs
```diff
@@ -0,0 +1,718 @@
...
+        protected float MeasureStringWidth(string text) {
+            if (string.IsNullOrEmpty(text)) return 0;
+
+            var lastGlyph = _font.GetGlyphs(text).Last();
+
+            return lastGlyph.Position.X + (lastGlyph.FontRegion?.Width ?? 0);
+        }
+
+        private string ProcessText(string value) {
+            value = value?.Replace("\r", string.Empty);
```
Denrage: check if null and return instantly, maybe

Blish HUD/Controls/TextInputBase.cs
```diff
@@ -0,0 +1,718 @@
...
+        protected bool SetText(string value, bool byUser) {
+            string prevText = _text;
+
+            value = ProcessText(value);
+
+            if (!SetProperty(ref _text, value)) return false;
+
+            // TODO: Update formatted text?
+
+            if (!byUser) {
```
Denrage, line 490: You could put that into the setter and use `SetText` inside this class w/o the boolean

Blish HUD/Controls/TextInputBase.cs
```diff
@@ -0,0 +1,718 @@
...
+                this.CursorIndex = _selectionStart = _selectionEnd = 0;
+            }
+
+            if (!_suppressRedoStackReset) {
+                _redoStack.Reset();
+            }
+
+            _cursorMoved = true;
```
Denrage: this will be set anyway in the `OnCursorIndexChanged`, or not?

Blish HUD/Controls/TextInputBase.cs
```diff
@@ -0,0 +1,718 @@
...
+
+        private static readonly Logger Logger = Logger.GetLogger<TextInputBase>();
+
+        protected static readonly char[] WordSeperators = { ' ', '\n', '`', '~', '!', '@', '#', '%', '^', '&', '*', '(', ')', '-', '=', '+', '[', '{', ']', '}', '\\', '|', ';', ':', '\'', '"', ',', '.', '<', '>', '/', '?' };
+
+        #region Load Static
+
+        private static readonly Color _highlightColor;
+
+        static TextInputBase() {
```
Denrage: Why is the static contructor needed here?

Blish HUD/Controls/TextInputBase.cs
```diff
@@ -0,0 +1,718 @@
...
+            _highlightColor = new Color(92, 80, 103, 150);
+        }
+
+        #endregion
+
+        protected const char NEWLINE = '\n';
```
Denrage, line 50: What about `Environment.Newline`?

Blish HUD/Controls/TextInputBase.cs
```diff
@@ -0,0 +1,718 @@
...
+        /// Fires when the <see cref="CursorIndex"/> is changed.
+        /// </summary>
+        public event EventHandler<ValueEventArgs<int>> CursorIndexChanged;
+
+        protected void OnTextChanged(ValueChangedEventArgs<string> e) => TextChanged?.Invoke(this, e);
+
+        protected void OnCursorIndexChanged(ValueEventArgs<int> e) {
+            _cursorMoved = true;
+
+            UpdateScrolling();
```
Denrage, line 67: Why not letting the implementation classes either subscribing to the `CursorIndexChanged`, or let them override `OnCurserIndexChanged`?
With this the `MultilineTextBox` needs to "implement" the `UpdateScrolling` w/o even doing anything in there

Denrage (changes requested):

Blish HUD/Controls/TextInputBase.cs
dlamkins: `Environment.NewLine` returns a `string` instead of `char` and on Windows is `\r\n` instead of just `\n` which is what the multiline control supports in this implementation.

Blish HUD/Controls/TextInputBase.cs
dlamkins: `InsertChar` is only called by `InputChar` which checks for it already.

Blish HUD/Controls/TextInputBase.cs
dlamkins: There is currently:

- InsertChar
- InsertChars
- InputChar
- RemoveUnsupportedCharacters

I can rename the last to `RemoveUnsupportedChars` if you would prefer.

Denrage: I would prefer Characters because i personally dislike abbreviations (C/C++ nightmares) but i don't have any objective arguments for that matter so it's your decision.

Blish HUD/Controls/TextInputBase.cs
dlamkins: > You have a property for "Text" for a reason.

That reason is primarily public access from other code.  In the case of the `Text` property, to ensure that when text is set the undo/redo buffers are updated, cursor position is handled properly, invalid characters are moved, etc.

To avoid a jumble of using both backend and properties within the owning class, I've generally kept to a pattern of only using the public properties if the additional logic in the getter or setter is actually needed when accessing or setting them internally.

This is especially important with setters as all of these setters will fire an event and several will invalidate the control layout if updated using the public setter.  You can see an example of this within `DeleteSection` where we use the property/setter just for `SelectionEnd` to ensure that the control isn't invalidated until it-itself is set and doesn't get fired twice (which would happen if `CursorIndex` was used instead of `_cursorIndex`).

Blish HUD/Controls/TextInputBase.cs
dlamkins: `MultilineTextBox` will actually need to implement `UpdateScrolling` at some point, but the exact behavior hasn't been decided on.

Blish HUD/Controls/TextInputBase.cs
dlamkins: I've updated it to standardize to "Char(s)" simply to help match what is "upstream" in the Myra implementation in case we wish to pull future improvements to it into our own implementation since it'll be easier to follow the function names if they still match.

Blish HUD/Controls/TextInputBase.cs
dlamkins: This is again to match the original implementation.

https://github.com/rds1983/Myra/blob/a9dbf7a1ceedc19f9e416c754eaf38e89a89a746/src/Myra/Graphics2D/UI/TextBox.cs#L545-L558

It's not for an optimization.

Blish HUD/Controls/TextInputBase.cs
dlamkins: This is maintaining the convention used in Myra:
https://github.com/rds1983/Myra/blob/a9dbf7a1ceedc19f9e416c754eaf38e89a89a746/src/Myra/Graphics2D/UI/TextBox.cs#L860-L867

Also doesn't matter as a whole and letting `SetText` accept the bool lets implementing classes provide it as well.

Blish HUD/Controls/TextInputBase.cs
dlamkins: Resolved.

Blish HUD/Controls/TextInputBase.cs
dlamkins: Yep!  Removed it.

Blish HUD/Controls/TextInputBase.cs
dlamkins: Not a bad idea.  Now immediately returns `string.Empty` if it sees it is null.  Removed the null checks in the remainder of the function.

Blish HUD/Controls/_Types/UndoRedoStack.cs
dlamkins: The `UndoRedo` method in `TextInputBase` needed access to directly pop it.  It's been marked as internal and nothing else within the assembly should be poking at it too much anyways (I'd hope we'd notice anything weird in a PR).

That being said, I realized the calls in `UndoRedo` that need direct access to the stack could probably get reimplemented directly into `UndoRedoStack` and exposed with a single function call, so I've gone ahead and done that and removed the property for the Stack.

Blish HUD/Controls/TextInputBase.cs
dlamkins: Fixed!

Blish HUD/Controls/TextInputBase.cs
dlamkins: Updated!

Blish HUD/Controls/TextInputBase.cs
dlamkins: Copied from when this was originally brought up.  It'll get handled in a separate PR, but something to look over in case you have thoughts.

> - Decide on how to support text that extends beyond the width of the control (word wrap?  horizontal scroll bar? something else?) - I am not looking forward to implementing whatever is decided here.  The game wraps lines that extend beyond the end of the control.
> - Decide on how to support text that extends beyond the bottom of the control (vertical scroll bar? something else?) - The game lets this happen, but does not show a vertical scrollbar, so scrolling up and down when this happens is weird.

Blish HUD/Controls/TextInputBase.cs
Denrage: Can be resolved

Blish HUD/Controls/TextInputBase.cs
Denrage: Can be resolved

Blish HUD/Controls/_Types/UndoRedoStack.cs
Denrage: Looks good!

Blish HUD/Controls/TextInputBase.cs
Denrage: Implementing classes can still use `SetText` and the setter like the base class, so it won't matter there. But i cannot say anything against the first argument.

Blish HUD/Controls/TextInputBase.cs
Denrage: Can be resolved

Blish HUD/Controls/TextInputBase.cs
Denrage: I've no problem if no one will not implement it in the future, so can be resolved

Blish HUD/Controls/MultilineTextBox.cs
```diff
@@ -0,0 +1,237 @@
...
+
+            if (_cursorIndex == newIndex && e.IsDoubleClick) {
+                SelectAll();
+            } else {
+                UserSetCursorIndex(newIndex);
+                ResetSelection();
+            }
+        }
+
+        private int GetCursorIndexFromFromPosition(Point clickPos) {
```
Denrage: FromFrom?

Blish HUD/Controls/MultilineTextBox.cs
```diff
@@ -0,0 +1,237 @@
...
+        protected override void OnClick(MouseEventArgs e) {
+            base.OnClick(e);
+
+            int newIndex = GetCursorIndexFromFromPosition(this.RelativeMousePosition - new Point(TEXT_LEFTPADDING, TEXT_TOPPADDING));
+
+            if (_cursorIndex == newIndex && e.IsDoubleClick) {
+                SelectAll();
+            } else {
+                UserSetCursorIndex(newIndex);
+                ResetSelection();
```
Denrage: Not handling the Shift-Key Selection here?

Blish HUD/Controls/Textbox.cs
```diff
@@ -1,233 +1,136 @@
...
-        #region Load Static
+        private const int TEXT_HORIZONTALPADDING = 10;
 
-        private static MouseInterceptor           _sharedInterceptor;
-        private static System.Windows.Forms.Label _sharedUnfocusLabel;
+        #region Load Static
 
         private static readonly Texture2D _textureTextbox;
 
         static TextBox() {
```
Denrage, line 18: Static contructor still needed?

Blish HUD/Controls/Textbox.cs
```diff
@@ -1,233 +1,136 @@
...
-                ShortcutsEnabled         = true,
-                TabStop                  = false
-            };
-
-            _sharedInterceptor.LeftMouseButtonReleased += delegate {
-                if (_sharedInterceptor.ActiveControl == this) {
-                    Textbox_LeftMouseButtonReleased(null, null);
-                }
-            };
+        protected override void MoveLine(int delta) { /* NOOP */ }
```
Denrage: Shouldn't this set the cursor position at the beginning and at the end?

Blish HUD/Controls/Textbox.cs
```diff
@@ -1,233 +1,136 @@
...
-
-        protected override void OnMouseEntered(MouseEventArgs e) {
-            bool restoreFocus = _mttb.Focused;
-
-            _sharedInterceptor.Show(this);
-
-            if (restoreFocus) {
-                Textbox_LeftMouseButtonReleased(null, null);
+            if (_cursorIndex == newIndex && e.IsDoubleClick) {
+                SelectAll();
```
Denrage: Shouldn't this select the word instead of the hole text?

Blish HUD/Controls/Textbox.cs
```diff
@@ -1,233 +1,136 @@
...
             }
-
-            base.OnMouseEntered(e);
         }
 
-        protected override void OnMouseLeft(MouseEventArgs e) {
-            if (_sharedInterceptor.ActiveControl == this) {
-                _sharedInterceptor.Hide();
-            }
+        private int GetCursorIndexFromFromPosition(int x) {
```
Denrage: FromFrom?

Blish HUD/Controls/_Types/KeyRepeatState.cs
```diff
@@ -0,0 +1,35 @@
...
+            _lastInterval   = gameTime.TotalGameTime;
+            _repeatableArgs = repeatableArgs;
+        }
+
+        public void HandleUpdate(GameTime gameTime, EventHandler<KeyboardEventArgs> handler) {
+            int minDelay = _hasDelayed
+                               // https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.systeminformation.keyboardspeed?view=netframework-4.8#remarks
+                               ? -12 * SystemInformation.KeyboardSpeed + 400
+                               // https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.systeminformation.keyboarddelay?view=netframework-4.8#remarks
+                               : (SystemInformation.KeyboardDelay + 1) * 250;
```
Denrage: Constants would be nice, so someone can see at one glance for what those numbers are

Blish HUD/Controls/_Types/UndoRedoStack.cs
```diff
@@ -0,0 +1,77 @@
...
+namespace Blish_HUD.Controls {
+    internal class UndoRedoStack {
+
+        private readonly Stack<UndoRedoRecord> _stack = new Stack<UndoRedoRecord>();
+
+        public void Reset() {
+            _stack.Clear();
+        }
+
+        public UndoRedoRecord Pop() {
```
Denrage, line 38: Idk, if you should return an `UndoRedoRecord`-Object bc you are also not using those as inputs in the `Make`-Methods. Feels a little bit inconsistent that way. But don't ask me of a better way (bc the consumer needs to know the OperationType at least). Can also stay at is, just wanted to note.

Blish HUD/Controls/MultilineTextBox.cs
```diff
@@ -0,0 +1,237 @@
...
+            UpdateSelectionIfShiftDown();
+        }
+
+        protected override void OnClick(MouseEventArgs e) {
+            base.OnClick(e);
+
+            int newIndex = GetCursorIndexFromFromPosition(this.RelativeMousePosition - new Point(TEXT_LEFTPADDING, TEXT_TOPPADDING));
+
+            if (_cursorIndex == newIndex && e.IsDoubleClick) {
+                SelectAll();
```
Denrage: Shouldn't this select the word instead of the hole text?

Blish HUD/Controls/MultilineTextBox.cs
dlamkins: I hadn't even thought of that.  Fixed by replacing it with a call to `UpdateSelectionIfShiftDown()` instead of `ResetSelection()`.

Blish HUD/Controls/Textbox.cs
dlamkins: For the purpose of maintaining consistency with the rest of the project, yes.  I dislike assigning textures inline.

Blish HUD/Controls/Textbox.cs
dlamkins: Fixed by checking the delta and calling either `HandleHome(false)` or `HandleEnd(false)`.

Blish HUD/Controls/_Types/UndoRedoStack.cs
dlamkins: That's fair.  I don't expect this portion of the work to get a lot of "traffic" in terms of future changes, so I probably will leave it as is now even if only to simply the implementation on the TextInputBase side of things.

Blish HUD/Controls/Textbox.cs
dlamkins: The answer to this is _yes_, but I don't think it will be included in this PR.  Should just be a matter of searching in both directions for the word boundary and updating the selection and would probably be a method on `TextInputBase` since both would use it so I'll make note to have it added in the future.

Blish HUD/Controls/MultilineTextBox.cs
dlamkins: Copied from above:

The answer to this is _yes_, but I don't think it will be included in this PR.  Should just be a matter of searching in both directions for the word boundary and updating the selection and would probably be a method on `TextInputBase` since both would use it so I'll make note to have it added in the future.

Blish HUD/Controls/Textbox.cs
dlamkins: Just kidding.  I went ahead and got it implemented.  🙂

Blish HUD/Controls/MultilineTextBox.cs
dlamkins: Just kidding.  I went ahead and got it implemented.  🙂

dlamkins:
Pulled more of the mouse click handling into `TextInputBase` since it was heavily duplicated between both controls.

Blish HUD/Controls/Textbox.cs
Denrage: Can be resolved then

Blish HUD/Controls/_Types/UndoRedoStack.cs
Denrage: Can be resolved then ^-^

dlamkins: Thanks!

dlamkins:
Items brought up in Discord have been addressed.  🙂 

https://discordapp.com/channels/531175899588984842/536970543736291346/701041505074413588

Denrage (approved):

PR #182, opened 2020-03-30
Large class cleanup

dlamkins:
Worked to resolve a significant number of simple issues based on SonarCloud recommendations.

https://sonarcloud.io/dashboard?branch=dev&id=blish-hud_Blish-HUD

Focused on changes that:
- Did not change functionality
- Helped improve code readability by reducing code complexity
- Improved consistency
- Improved performance


Blish HUD/Controls/Control.cs
```diff
@@ -793,8 +786,6 @@ public static class StandardColors {
                     this.DisposeControl();
                 }
 
-                // TODO: free unmanaged resources (unmanaged objects) and override a finalizer below.
-                // TODO: set large fields to null.
                 _disposedValue = true;
```
greaka, line 789: can you give a quick comment why this is not needed anymore?

Blish HUD/Controls/Panel.cs
```diff
@@ -197,7 +197,7 @@ public class Panel : Container, IAccordion {
         }
 
         private void UpdateContentRegionBounds(object sender, EventArgs e) {
-            //UpdateScrollbar();
+            UpdateScrollbar();
```
greaka, line 200: did you test that this doesn't have any ugly side effects? do you remember why it was commented out in the first place?

Blish HUD/Controls/Scrollbar.cs
```diff
@@ -86,29 +86,33 @@ public class Scrollbar:Control {
...
-            Input.Mouse.LeftMouseButtonReleased += delegate { if (Scrolling) { Scrolling = false; /* Invalidate(); */ } };
+            Input.Mouse.LeftMouseButtonReleased += MouseOnLeftMouseButtonReleased;
 
             _associatedContainer.MouseWheelScrolled += HandleWheelScroll;
         }
 
+        private void MouseOnLeftMouseButtonReleased(object sender, MouseEventArgs e) {
+            _scrolling = false;
+        }
+
```
greaka, line 115: Sometimes you use `Input_EventInPascalCase` and sometimes you use `InputEventInPascalCase`. Since this is a cleanup pr, it would be a great opportunity to get this straight.

greaka (commented):
nothing major, i just want to make sure that nothing of importance is missed here

Blish HUD/Controls/Control.cs
dlamkins: It was a misunderstanding of mine at the time that it would assist with GC to clear some of the larger fields, but this of course is done automatically.

The reference to freeing unmanaged resources is only relevant to the TextBox control (and soon won't be with the new work being done on the TextBox control) and the CachedRender stuff.  Both of these already handle freeing unmanaged resources when disposed as described.

Blish HUD/Controls/Scrollbar.cs
dlamkins: Do you mind linking an example of both so I can ensure I'm fully on the same page.  I agree that this is likely a good time to get whatever it is consistent.  🙂

greaka: i sent you all occurences of both on discord
here they are again:
https://cdn.discordapp.com/attachments/558957192493924364/694437814951215144/message.txt
https://cdn.discordapp.com/attachments/558957192493924364/694438320448602142/message.txt

greaka: i just saw that the latter one is missing 3 functions, because my regex didnt account for numbers in the function name

dlamkins: Thank you for sending those!  That is very odd.  I haven't noticed that.  Looks like VS tries to auto-complete it two different ways.  Likely a consequence of having a style specified in ReSharper that maybe doesn't match the Visual Studio default.

https://gfycat.com/wealthydecisivearrowworm

You can see the first tooltip show it with an `_` and the dropdown from the IntelliSense showing it without.

dlamkins: Okay, the `_` is apparently the default in the VS provided method and ReSharper was providing its own.  After some review, I've opted to go with the convention ReSharper uses.  Fixed the ones you linked in 3d34091 & 949e518.

Blish HUD/Controls/Panel.cs
dlamkins: Yes, per the git blame, it looks like I commented it out 9 months ago.  It covers the case where a panel has `CanScroll` enabled but does not have enough content to necessitate it scrolling.  If a new control is added that is either larger than or extends passed the content region necessitating a scroll bar, a scroll bar is given the chance to be instanced.

greaka (approved):

PR #190, opened 2020-04-17
Update Gw2Sharp to 0.9.4 and fixes

dlamkins:
- IsMapOpen is now used to toggle 3D entity visibility instead of requiring ArcDps.
- ContentService Render Service queries now use Gw2Sharp and responses are cached (in memory still).
- Removed unused fields and methods from Gw2ApiService.
- Removed old render service methods for using darth's CDN.


agaertner (approved):
Gw2Sharp is being updated to 0.9.5 soonish. 
[Gw2Sharp CHANGELOG.md](https://github.com/Archomeda/Gw2Sharp/blob/master/CHANGELOG.md#095)

PR #191, opened 2020-04-21
Fix visual on pressed

agaertner:
Branch can be deleted after merge.


agaertner:
* This makes the icons small when they are pressed not when just hovered.

dlamkins:
Yep, this looks fine.  Thanks!

dlamkins (approved):

PR #196, opened 2020-04-28
WIP: Improvements for Markers & Paths

dlamkins:
This marks the beginning of some changes to the `PathingService` and other supporting classes needed for the Markers & Paths module.

To reduce the amount of abstraction on abstraction on abstraction on abstraction, we'll be pulling the pathing specific details out of the core codebase and into the Markers & Paths module.  This should reduce complexity, allow for faster interactive changes to the Markers & Paths module, and allow for a simplification to entities needed in the core codebase to help facilitate multiple worlds which have their own transforms (such as `World` and `MiniMap`).

This change does break the Markers & Paths module (as will many upcoming changes to the pathing service).  These first several changes are primarily to help with the addition of several new attributes (`Festival`, `Profession`, `Specialization`, `Race`, and potentially `IsCommander`).

### DIFF
--- a/Blish HUD/GameServices/Gw2Mumble/PlayerCharacter.cs
+++ b/Blish HUD/GameServices/Gw2Mumble/PlayerCharacter.cs
@@ -15,13 +15,23 @@ namespace Blish_HUD.Gw2Mumble {
         /// </summary>
         public event EventHandler<ValueEventArgs<string>> NameChanged;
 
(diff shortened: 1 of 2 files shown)


dlamkins:
I think I'm going to pull all of this into the Markers & Paths module first and then delete the pathing service instead of trying to do both at the same time.

PR #203, opened 2020-05-02
Updated MonoGame to 3.7.1 (for real this time)

dlamkins:
Updated MonoGame to 3.7.1 (for real this time) and forced all textures to load in premultiplied.

Also removed the many DLLs no longer needed.
- MonoGame / SharpDX / etc. are now included from NuGet.
- DiscordRPC.dll is bundled with the Discord RPC module.
- GW2NET is no longer relevant as we are using Gw2Sharp entirely.


greaka (approved):
as said, wasnt able to test, but looks good

PR #205, opened 2020-05-02
Migrate to new .csproj style and support exporting as NuGet package

Archomeda:
The following things have been updated:
- .gitignore because why not
- .csproj to .NET Core style (but supports .NET Framework targeting)
- Forward slashes in ref.dat references instead of backward slashes (new zipping method apparently requires that)
- Various config files are no longer needed
- Build scripts have been migrated to .csproj targets
- Support to export as NuGet package for easier consumption when developing modules
- Added a custom .targets file for custom build steps when the NuGet package gets consumed by a module

The new .csproj style might change the way how msbuild works. I don't think msbuild can handle this style, so you'll probably have to use dotnet instead:
- `dotnet build` to build (with optional `-c <CONFIGURATION>`)
- `dotnet pack -p:Version=x.y.z` to create a NuGet package with a specific version
- `dotnet nuget push` to push the NuGet package

The .csproj contains all relevant parameters in order to create a NuGet package. I've configured it so that a symbol package is created as .snupkg and I believe `dotnet nuget push` will push the snupkg as well.
You might want to check all parameters in the .csproj to see if they are set to what you want (e.g. author, description, versions, etc...)

You'll most likely have to update the CI build scripts. A couple of things to look out for:
- Versioning; previously this was handled in AssemblyInfo.cs, but the new .csproj style supports those tags directly in the project file (it's possible to revert this though)
- Building; as mentioned above, the build command should most likely be `dotnet build` now
- Output folder; the new csproj forces the output directory to be changed, files are put in the net472 subfolder now instead
- Packing NuGet; if you want to change the package id (I just picked BlishHUD), you'll have to rename `Blish HUD/build/BlishHUD.targets` as well (the .targets file requires the same filename as the package id)

--- 

For modules, once the NuGet package is up, you'll only have to include this (in a new style .csproj):
```xml
  <ItemGroup>
    <PackageReference Include="BlishHUD" Version="x.y.z" IncludeAssets="build;compile" />
  </ItemGroup>

  <ItemGroup>
    <Reference Include="System.ComponentModel.Composition" />
  </ItemGroup>
```
This will make sure that Blish HUD is used as reference, the .targets file is used while building the module, but is not copied to the output folder. `System.ComponentModel.Composition` is required for the module attributes. Without it, I don't believe it will build (but a code fixer should notify you about importing it).

---

Also a small note about the package references. I made sure that the references that NuGet lists, are the same as the current template. This means: Gw2Sharp, MonoGame.Framework.WindowsDX and Newtonsoft.Json. I'm not sure if any other package should be made available to modules, but if needed, you can remove the `PrivateAssets="all"` part, because this is responsible for "hiding" it when it's referenced.
At the same time, the ones that are not private, are now locked to a specific version (note the brackets). This will make sure that when developing against a specific Blish HUD version, someone doesn't accidentally reference a different version.


Blish HUD/build/BlishHUD.targets
```diff
@@ -0,0 +1,43 @@
+<Project>
+
+  <Target Name="CheckBlishHUDModuleRequiredFiles" BeforeTargets="Build">
+    <!-- Check for manifest.json -->
+    <Error Condition="!Exists('$(ProjectDir)manifest.json')" Text="The file &quot;$(ProjectDir)manifest.json&quot; was not found. This file is required for creating a BlishHUD module." />
```
dlamkins: Could you add a space between "Blish" and "HUD" here (just being pedantic, but there is a space in it where we can) and could we potentially include a link to the manifest repo/doc so it's easy to know where to go next?  🙂 

https://github.com/blish-hud/manifest.json/blob/master/manifest-v1.md

dlamkins: ("BlishHUD" is of course fine for anything internal where a space would be obnoxious, but for messaging my brain likes the consistency.)

Blish HUD/Blish HUD.csproj
```diff
@@ -1,766 +1,226 @@
...
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+    </PackageReference>
+    <PackageReference Include="Gw2Sharp" Version="[0.9.6]" />
+    <PackageReference Include="Humanizer.Core.de" Version="2.6.2" PrivateAssets="all" />
+    <PackageReference Include="Humanizer.Core.es" Version="2.6.2" PrivateAssets="all" />
+    <PackageReference Include="Humanizer.Core.fr" Version="2.6.2" PrivateAssets="all" />
+    <PackageReference Include="JsonSubTypes" Version="1.6.0" PrivateAssets="all" />
+    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="all" />
+    <PackageReference Include="MonoGame.Content.Builder" Version="3.7.0.9" PrivateAssets="all" />
+    <PackageReference Include="MonoGame.Extended" Version="3.7.0" PrivateAssets="all" />
```
dlamkins: Could we get the following included with modules as well?
- MonoGame.Extended (but not MonoGame.Extended.Content.Pipeline)
- ~Fody & Costura (assuming we don't go with ILMerge for modules)~
_(Never mind: #207 enables assembly resolving for modules)_
- AsyncClipboardService

dlamkins (changes requested):

Blish HUD/build/BlishHUD.targets
dlamkins: > Packing NuGet; if you want to change the package id (I just picked BlishHUD), you'll have to rename Blish HUD/build/BlishHUD.targets as well (the .targets file requires the same filename as the package id)

And BlishHUD is certainly fine here as well.

Archomeda: > could we potentially include a link to the manifest repo/doc so it's easy to know where to go next?

I thought about that, and sure I can put it in the message, but it's not clickable nor copyable. So it seemed a bit useless to me. Of course, I can still add it if you want.

Blish HUD/Blish HUD.csproj
Archomeda: Done :)

Blish HUD/build/BlishHUD.targets
Archomeda: I updated Blish HUD in a few places :)

dlamkins: That's fair.  I just want them to know at least where to look so they have next steps if necessary.  🙂

dlamkins: I'll just make sure that the docs for new modules makes it clear.  🙂

dlamkins (approved):
These are some truly fantastic changes and improvements and I truly appreciate the contribution.  This is a huge leap forward to making module development significantly easier.

Thanks!  👍

PR #214, opened 2020-05-07
set working dir to application executable path

bennieboj:
set working dir to application executable path


dlamkins:
Nice and simple - I love it.  Thanks!  👍

dlamkins (approved):

PR #219, opened 2020-05-11
GitHub actions

greaka:
@Archomeda what is still needed to merge this?


Archomeda:
You'd probably want to check the first version that's used in GitVersion.yml. For the rest, possibly maybe different filenames for the artifacts, but it's not required. Or maybe advanced pipelines. But the base is essentially ready I think.

Archomeda:
@dlamkins @greaka Just to let you know that this is ready as far as I can tell, unless you want to customize the workflow files some more of course.

Archomeda:
Right, one thing to add is probably SonarCloud still. But I'm unsure if it's possible for me to configure it. Here's a reference to the workflow file I'm using for Gw2Sharp: https://github.com/Archomeda/Gw2Sharp/blob/master/.github/workflows/sonarcloud.yml

dlamkins:
Okay, I think I just need to review how the versioning works with this and the artifact names.  I won't worry about SonarCloud for now.

GitVersion.yml
```diff
@@ -0,0 +1,3 @@
+next-version: 0.5.2
```
dlamkins, line 1: So as commits come in, does it update this itself, or does it use this as a base and use the commit history to generate the version each time?

Archomeda: This is used as a base and it will look in the history to resolve the correct version. If you decide to use tags at some point, those will take priority, and next-version doesn't get used I believe. At least in Gw2Sharp I don't have this defined and the tags are used instead.

dlamkins (approved):

PR #220, opened 2020-05-13
Add debug input hooks for improving debugging experience

Archomeda:
A whole bunch of code to implement a separate process that gets launched that handles the global input hooks from Windows. This is currently only used on `--debug`.

No more annoying delays when debugging Blish HUD or modules, while still keeping the input hooks enabled 🎉 

This also changes the C# langversion to 8 since I've used a few statements that can only be used in there. This should not have any breaking changes.

For the detailed implementation, see [this comment](https://github.com/blish-hud/Blish-HUD/pull/220#issuecomment-633052370).

If there's anything that needs to be changed, let me know.


Blish HUD/GameServices/Input/Keyboard/KeyboardHandler.cs
```diff
@@ -102,25 +91,30 @@ public class KeyboardManager : InputManager {
 
             UpdateStates();
 
-            foreach (var key in passingKeys) {
-                OnKeyStateChanged(new KeyboardEventArgs(KeyboardEventType.KeyUp, key));
-            }
+            foreach (Keys key in passingKeys) OnKeyStateChanged(new KeyboardEventArgs(KeyboardEventType.KeyUp, key));
```
dlamkins, line 94: Is it the editor config / ReSharper config that is consolidating all of these down?

Blish HUD/GameServices/Input/Mouse/MouseHandler.cs
```diff
@@ -52,124 +51,118 @@ public class MouseManager : InputManager {
...
-                #if !NOMOUSEHOOK
-                return newEvent.EventType != MouseEventType.LeftMouseButtonReleased;
-                #endif
-            } else if (newEvent.EventType == MouseEventType.RightMouseButtonPressed) {
-                _cameraDragging = true;
-            }
+            else if (_hudFocused && !_hookOverride) {
+                _mouseEvent = mouseEventArgs;
+                return mouseEventArgs.EventType != MouseEventType.LeftMouseButtonReleased;
+            } else if (mouseEventArgs.EventType == MouseEventType.RightMouseButtonPressed) _cameraDragging = true;
```
dlamkins, line 136: I'm not a fan of skipping on braces on single liners outside of early exist / shortcircuit stuff at the top, but the mixing of having braces and not braces for some of these if blocks is a bit much.

I've kind of just stuck with https://stackoverflow.com/a/2125078/595437 to make life a bit easier.  

Was this something changed by the editorconfig or was this manually updated?

Blish HUD/Blish HUD.csproj
```diff
@@ -5,7 +5,7 @@
     <Platforms>x64</Platforms>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <OutputType>WinExe</OutputType>
-    <LangVersion>7.3</LangVersion>
+    <LangVersion>8</LangVersion>
```
dlamkins, line 8: Nothing wrong with this per se, but should probably at least have a mention in the actual PR itself.

dlamkins:
Probably going to take me a bit to get through.  Looks like there are a number of breaking changes - especially with some of the classes getting renamed.

Looks great so far, though!  🙂

dlamkins:
Would probably be good to get a quick write up of how it works at a high level.  Something that can be referenced easily later.

Archomeda:
Alright, it's been a while, and let's see if I can get everything right still.

The way it works on high level is as follows:
Depending on the launch mode, it will either use the already available one in Blish HUD, or the new debug hooks one. They are implemented through newly made interfaces so they are easily interchangeable.

The debug hooks are made in a separate project that creates a separate executable that will always live next to Blish HUD (maybe we can change that so it's included in the Blish HUD executable, but for now I've left it at this). When Blish HUD is running in this mode, it will start (and maintain) the debug hooks process. The debug hooks process will hook the global input events, just like Blish HUD normally does outside this mode. Blish HUD communicates with this process through the stdin and stdout streams.

This communication protocol is as follows:
- Blish HUD periodically sends a keep-alive message (or a ping) to keep the debug hooks process "active"
- As long as the debug hooks are active, it will intercept and forward the global input events
- Blish HUD can act on these events it receives from the debug hooks, and may send a response back
- If Blish HUD doesn't send a response back, the debug hooks will time this out shorter than the default Windows timeout to maintain responsiveness

The following configuration has been applied:
- The debug hooks process waits for 10ms before timing out a forwarded input event message and reporting back to Windows
- The debug hooks process has an internal timer that resets itself every time a ping message is received; if there's no ping for 50ms, the debug hooks process will stop the global hooks until a new ping has been received (this is to prevent timeouts due to Blish HUD being breakpointed; since the pings are sent from Blish HUD, they aren't sent when the process is paused)
- Blish HUD pings the debug hooks process every 10ms in order to be well under the 50ms timeout from above

When Blish HUD is closed, the debug hooks process will shut itself down as well:
- Either Blish HUD kills it when it's unloading
- Or the debug hooks process kills itself if it no longer detects that Blish HUD is running (its process id is sent as a command line parameter when it's started)

This is to prevent zombie processes that no one wants.

---

If there's anything more you would want to know, you know where to find me 😁

dlamkins (changes requested):

Blish HUD/GameServices/Input/Keyboard/KeyboardHandler.cs
Archomeda: I believe so yes. I ran ReSharper on all the files I created (maybe the ones that I have changed too).

Blish HUD/GameServices/Input/Mouse/MouseHandler.cs
Archomeda: Same as the other comment. I ran ReSharper, so it must have changed this too.

Blish HUD/Blish HUD.csproj
Archomeda: Ah. You're right. I've used some statements that can only be used in version 8.

Blish HUD/GameServices/Input/Keyboard/KeyboardHandler.cs
dlamkins: Thanks!

dlamkins (approved):

dlamkins:
Looks like a number of breaking changes.  Will likely mark as v0.6.

PR #221, opened 2020-05-14
Flowpanel splitting padding types

greaka:
Breaking change ahead:

I'd like to get some feedback to this approach.


dlamkins:
Went with the changes in #234 to avoid breaking changes.

PR #232, opened 2020-05-25
Chat Interactions

agaertner:
I am using these functions in several modules including KillProof.
I am adding them to core because I am annoyed to dupe them and they might come in handy for other modules as well.

I rewrote it a little bit and I am unsure if I should `await GetTextAsync()` in order to backup the old clipboard content
right before `await SetTextAsync()` to set the custom message in order to be able to use the Ctrl-V shortcut. According to the [AsyncClipboard documentation](https://www.nuget.org/packages/AsyncClipboardService/) this is set up alright, hence why I did not have any issues during testing. I even spammed chat codes without problem. I am still worried and wondering if I should nest another `.ContinueWith`

(branch can be deleted after merge)


Blish HUD/Common/Gw2/Chat.cs
```diff
@@ -0,0 +1,110 @@
...
+                        Keyboard.Press(VirtualKeyShort.LCONTROL, true);
+                        Keyboard.Stroke(VirtualKeyShort.KEY_A, true);
+                        Keyboard.Stroke(VirtualKeyShort.KEY_C, true);
+                        Thread.Sleep(50);
+                        Keyboard.Release(VirtualKeyShort.LCONTROL, true);
+                        Unfocus();
+            });
+            var currentInput = await ClipboardUtil.WindowsClipboardService.GetTextAsync();
+            if (prevClipboardContent != null)
+                await ClipboardUtil.WindowsClipboardService.SetTextAsync(prevClipboardContent);
```
dlamkins: It looks like we nicely restore the users clipboard text which is good, but what about other clipboard data types like images?

dlamkins (changes requested):

Blish HUD/Common/Gw2/Chat.cs
agaertner: Ah yes. Now pulling bytes.
https://github.com/blish-hud/Blish-HUD/pull/232/commits/3364a957cb8fbbf138aae2f2ebda69cf0bda62b1

agaertner: Restoring every content is not feasable though.
Google: "backup and restore clipboard C#" and you will find plenty of answers why you shouldn't try to backup a complete clipboard.
For ex.
[#1 stackoverflow.com](https://stackoverflow.com/questions/2578900/how-do-i-backup-and-restore-the-system-clipboard-in-c)

Blish HUD/_Utils/ClipboardUtil.cs
```diff
@@ -19,6 +26,43 @@ public static class ClipboardUtil {
...
+            } catch (ThreadStateException e) {
+                Logger.Error("The current thread is not in single-threaded apartment (STA) mode and the MessageLoop property value is true. Add the STAThreadAttribute to your application's Main method. " + e.Message);
+            }
+            return null;
+        }
+        /// <summary>
+        /// Sets the clipboard's content by assigning it a new DataObject with the given content.
+        /// </summary>
+        /// <param name="content">Dictionary of objects indexed by their DataFormat (System.Windows.Forms.DataFormats).</param>
+        public static void SetClipboardContent(Dictionary<string, object> content) {
```
dlamkins: The async clipboard library we're using for this should actually be handling this for us so that we don't have to interface with the clipboard directly ourselves.  https://github.com/undergroundwires/AsyncWindowsClipboard

Blish HUD/Common/Gw2/Chat.cs
```diff
@@ -0,0 +1,120 @@
...
+using Blish_HUD.Controls.Extern;
+using Blish_HUD.Controls.Intern;
+using System.Linq;
+
+namespace Blish_HUD.Common.Gw2 {
+
+    /// <summary>
+    /// Methods related to interaction with the in-game chat.
+    /// </summary>
+    public static class Chat {
```
dlamkins: I think this would end up being better exposed under the GameIntegration service.  Something like `GameService.GameIntegration.Chat`.  Especially since it already depends on so much from it.

Blish HUD/Common/Gw2/Chat.cs
```diff
@@ -0,0 +1,120 @@
...
+                Keyboard.Stroke(VirtualKeyShort.BACK);
+                Unfocus();
+            });
+        }
+        private static void Focus() {
+            Unfocus();
+            Keyboard.Stroke(VirtualKeyShort.RETURN);
+        }
+        private static void Unfocus() {
+            GameService.GameIntegration.FocusGw2();
```
dlamkins: It GW2 actually needing focus at this point for it to work?  We used to need to set and unset focus primarily due to the way the old textbox implementation worked, but with that no longer in the way, the game should already have focus if the overlay is active.  

May be worth checking to see if the game is active (and not just running) so that we don't interrupt the user if they are doing something on another screen (not that we should be sending messages in chat on their behalf if they aren't actively in the game).

dlamkins (changes requested):

Blish HUD/_Utils/ClipboardUtil.cs
agaertner: these were test functions to backup the clipboard. The asyncwindowclipboard doesn't support all DataFormats and I am assuming this is because you shouldn't try to restore a clipboard except string type (which asyncclipboard sole support lies.)
Will be removing these functions again.

Blish HUD/Common/Gw2/Chat.cs
agaertner: Noted.

Blish HUD/Common/Gw2/Chat.cs
agaertner: I will try swapping the shortcut pressed with window messages like I did in Musician.
(WM_PASTE, WM_COPY etc) etc so I don't need to make these calls at all. Probably have to make a change or two in PInvoke somewhere.
If that won't work I will check for the Activity of the window.

Blish HUD/Common/Gw2/Chat.cs
agaertner: Review: https://github.com/blish-hud/Blish-HUD/pull/232/commits/f9ce5a737eec4052bf3b0d0e6b5af9962a9578f8

``GameService.GameIntegration.Chat``

agaertner:
> 
> 
> Looks like an extra semi-colon here which CodeFactor is gonna block for.

Uhm, you oversaw the commit here: https://github.com/blish-hud/Blish-HUD/pull/232/commits/0c6481b90fc7e932983fb19565bb4b2f7b7bd9b0
It's already gone :)

dlamkins:
Oh, you're right!  Thanks!

Blish HUD/_Utils/ClipboardUtil.cs
```diff
@@ -8,6 +13,8 @@ namespace Blish_HUD {
     /// </summary>
     public static class ClipboardUtil {
 
+        private static readonly Logger Logger = Logger.GetLogger(typeof(ClipboardUtil));
+
```
dlamkins: Logger doesn't appear to be needed anymore.

Blish HUD/_Utils/ClipboardUtil.cs
```diff
@@ -1,4 +1,9 @@
 ﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.InteropServices;
+using System.Threading;
+using System.Windows.Forms;
```
dlamkins: These usings aren't needed anymore, right?

dlamkins (changes requested):

Blish HUD/_Utils/ClipboardUtil.cs
agaertner: removed, logger, cleaned using: 
https://github.com/blish-hud/Blish-HUD/pull/232/commits/309af67318815a886ce42f53a11eec84ae377c9b

dlamkins (approved):

PR #234, opened 2020-05-27
Enable: Registering API Keys

dlamkins:
Allow users to register API keys through a new tab under settings.

1 API key can be stored per account (if you register a new one for the same account, it will overwrite the previous).  The Gw2ApiService will maintain which is the active token for consuming modules and services based on the character you are currently playing as.

Tokens are checked to ensure they are valid.  Information about them is shown once confirmed.

Also included:
- Deprecated `PadTopBeforeControl` and `PadLeftBeforeControl` and replaced them with `OuterControlPadding`.
- Cherry picked fading improvements for the `ViewContainer` from the MVP branch.
- Removed Presenter generic from Views as they can operate without touching any associated presenters (reduces complexity).
- Added support for undefining a setting by key name.


agaertner (approved):
Couldn't find anything except in my own translations in the resx file lol.
As a result I've 100% german now on POeditor with some fixes.

PR #237, opened 2020-05-31
include_wikimedia_lib_to_support_wiki_searching_in_universal_search

bennieboj:
include_wikimedia_lib_to_support_wiki_searching_in_universal_search
needed for this PR: https://github.com/blish-hud/Community-Module-Pack/pull/66

### DIFF
--- a/Blish HUD/Blish HUD.csproj
+++ b/Blish HUD/Blish HUD.csproj
@@ -176,6 +176,7 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
+    <PackageReference Include="CXuesong.MW.WikiClientLibrary" Version="0.7.1" />
     <PackageReference Include="DynamicLanguageRuntime" Version="1.2.2" PrivateAssets="all" />
     <PackageReference Include="EntryPoint" Version="1.3.0" PrivateAssets="all" />
     <PackageReference Include="Flurl.Http" Version="2.4.2" PrivateAssets="all" />


dlamkins:
Thanks!  Is there a reason why the reference can't get included directly in that module instead of needing to include it directly here in core?

bennieboj:
I did include it in the module, but that didn't work for some reason when running blish. Should this work? I noticed that other dependencies were also both in the module and in the core? I can remove it again and test if your want.

dlamkins:
Some of the dependencies are still in core since the modules used to actually get compiled as just classes directly within the main assembly.  We're slowly getting some of that out.

If the module project is still using Fody for dependency merging, that may be the issue.  Modules now support including DLL dependencies from NuGet or otherwise which can be loaded from their .bhm.  Those should get packaged up automatically if compiled normally.

I'm not going to make you try and guess and find out what it is, though.  I'll go over and poke at it some and see if I can get it to include the dependency correctly.  🙂

bennieboj:
Great, thanks for the help 😁

dlamkins:
Alright, between the semi-recent improvements to dependency loading for modules and https://github.com/blish-hud/Blish-HUD/pull/345 which is currently pending approval, the change here should no longer be necessary and should be able to get fully included in https://github.com/blish-hud/Community-Module-Pack/pull/66/commits/01a36b16becaa0e7e9deeb4af848937f2d30c7c4

bennieboj:
K, I subbed to #345  and will try to test once it's merged to develop

dlamkins:
Sorry for keeping this open for so long.  Was keeping it open as a reminder.  Gonna go ahead and close and review it again on the module side.  Thanks again!

PR #239, opened 2020-05-31
Multibox-friendly arcdps service

greaka:
With the 0.3.0 update of the arcdps-bhud bridge and this PR, Blish HUD will only receive arcdps events from the correct gw2 instance.

Testing is very much appreciated, as it is hard to test certain edge cases - like a randomly collapsing socket.
CI for the 0.3.0 bridge for testing is available here: https://ci.appveyor.com/project/greaka/arcdps-bhud-46l35/builds/33228186


greaka:
Ahh, dang it
I forgot, that the CI will only finish at thursday, because of a bug fix reaching the stable channel of rust then. In the meantime, I can give builds to whoever wishes to test in the discord.

greaka:
A build can be found here: https://discordapp.com/channels/531175899588984842/536970543736291346/716753991307886612

Pls do not merge until thursday, so I can release the new version of the bridge together with this dev build

Blish HUD/GameServices/ArcDpsService.cs
```diff
@@ -101,47 +90,56 @@ private void DispatchSkillSubscriptions(object sender, RawCombatEventArgs eventH
...
-            _server.Start(new IPEndPoint(IPAddress.Loopback, 8214));
+        protected override void Load() {
+            Gw2Mumble.Info.ProcessIdChanged += Start;
             _stopwatch.Start();
         }
 
-        protected override void Unload()
-        {
+        private void Start(object sender, ValueEventArgs<uint> value) {
+            if (this.Loaded) _server.Start(new IPEndPoint(IPAddress.Loopback, GetPort(value.Value)));
```
dlamkins, line 109: Do we stop listening if there was already a listener going?

For instance, if Blish HUD is left open between sessions and has multiple PIDs in the lifetime, will it properly stop the first listener?

Blish HUD/GameServices/ArcDps/SocketListener/SocketListener.cs
```diff
@@ -3,141 +3,96 @@
...
             };
-            _listenSocket.Bind(localEndPoint);
-            _listenSocket.Listen(_maxConnectionCount);
-            StartAccept(null);
+
+            _socketAsyncReceiveEventArgs.AcceptSocket   = _listenSocket;
+            _socketAsyncReceiveEventArgs.RemoteEndPoint = localEndPoint;
+
+            try {
+                if (!_listenSocket.ConnectAsync(_socketAsyncReceiveEventArgs)) ProcessConnect(_socketAsyncReceiveEventArgs);
```
dlamkins, line 40: Looks like we're just silently failing if the port isn't available for now. Is that correct?

greaka: Yes, that's right. Should I add a log entry? And which log level?

Blish HUD/GameServices/ArcDpsService.cs
greaka: For now, the gc will handle it. Between the runs of gw2 the socket would just wait for something to happen, which will never happen. When gw2 then starts again, the new socket will be initialized and the old one will be taken care of by gc.
I could add some handling here: https://github.com/blish-hud/Blish-HUD/blob/e7c443ea6703c7ee23befbb2637fbea089473c0f/Blish%20HUD/GameServices/ArcDps/SocketListener/SocketListener.cs#L59-L68
There is a disconnect enum variant that could either stop the listener altogether or try to reconnect a few times, etc.

Blish HUD/GameServices/ArcDps/SocketListener/SocketListener.cs
dlamkins: Yeah, a warning level log entry would be appropriate, I think.

Blish HUD/GameServices/ArcDpsService.cs
dlamkins: That makes sense.  Could maybe kill the listener when we detect the game closing (event in GameIntegration service), but not a big deal.

Thanks for clarifying!

Blish HUD/GameServices/ArcDps/SocketListener/SocketListener.cs
greaka: I added the log, but would like to mention that this will put out a warning every time someone starts gw2 without having the arcdps bridge installed

dlamkins: In that case, it may be appropriate to lower the log level to `Info` with something like "Unable to connect to Arcdps-BHUD bridge", but otherwise, that should be fine as it'll help indicate in logs, too, if the bridge was available or not during the session.

dlamkins (changes requested):

dlamkins (approved):

PR #240, opened 2020-06-01
DirectoryUtil update

agaertner:
Screens and music path.

There aren't any more paths that the game client creates in the documents folder. Might as well cover those here.


dlamkins (approved):
Thanks for completing the requested changes!

PR #241, opened 2020-06-01
Setting Renderers, FlowPanel Improvements, and Setting/About Views

dlamkins:
Introduces proper setting renderer views which can automate the display and wiring of supported setting types.  Currently those include:

- Bool
- String
- KeyBinding
- Enum (any)

Improvements have been made to the Container and FlowPanel controls to allow for these new renderers.

Containers can now specify a sizing method for both width and height.  This allows them to auto-size based on their child contents, fill the remaining space, or the normal behavior.

FlowPanel has additional flow rules allowing for "TopToBottom" now as well as single width column flows and single height row flows.

These features have been used to replace the static builders that were previously being used for the Overlay Settings panel and About panel.  These have now been replaced with actual views.


Blish HUD/Strings/GameServices/OverlayService.Designer.cs
```diff
@@ -144,7 +156,7 @@ internal class OverlayService {
         }
         
         /// <summary>
-        ///   Looks up a localized string similar to Minimize to tray when Guild Wars 2 Closes.
+        ///   Looks up a localized string similar to Minimize to tray when Guild Wars 2 Closes..
```
greaka, line 159: Is this intended?

Blish HUD/GameServices/Settings/UI/Views/SettingsView.cs
```diff
@@ -0,0 +1,87 @@
...
+                _lockBounds = value;
+
+                UpdateBoundsLocking(_lockBounds);
+            }
+        }
+
+        private ViewContainer _lastSettingContainer;
+
+        public SettingsView(SettingCollection settings, int definedWidth = -1)
+            : base(new SettingEntry<SettingCollection>() { Value = settings }, definedWidth) {
```
greaka: Can't you use `: this(...` here and make it a noop?

greaka (approved):

Blish HUD/Strings/GameServices/OverlayService.Designer.cs
dlamkins: It definitely looks odd, but it's autogenerated.  The actual string is "Minimize to tray when Guild Wars 2 Closes." which I did purposefully add a period to.  Looks like the autogenerated comment just uses `Looks up a localized string similar to {0}.`

Blish HUD/GameServices/Settings/UI/Views/SettingsView.cs
dlamkins: Good thinking.  Fixed!

Blish HUD/Strings/GameServices/OverlayService.Designer.cs
greaka: Alright, just wanted to make sure

greaka (approved):

PR #244, opened 2020-06-05
SettingRender fixes, scrollbar fixes, and locale event wiring.

dlamkins:
Addresses a number of issues brought up on 6/4/2020

- Settings are no longer rendered unless something in that session has ensured they are defined (prevents old and unused settings from rendering).
- Avoid condition where a view unloads/loads overlapping and causes scrollbar bindings to modify the collection while being enumerated.
- Wired UserLocaleChanged event.
- Ensured default CurrentUICulture matches on all new threads automatically.


agaertner (approved):

agaertner:
looks ok to me. 👍

PR #246, opened 2020-06-12
Fix slow mouse input when hooked through WinAPI

Archomeda:
Fixes a regression introduced in #220.
The input was accidentally hooked on the main thread instead of the dedicated thread, causing a significant slowdown when a high volume of messages are coming through (e.g. when moving the mouse while holding a button) while Blish HUD is also rendering stuff at the same time.


dlamkins (approved):
Thanks for the quick changes!

PR #247, opened 2020-06-13
Multibox Support

dlamkins:
- Modifies the application mutex to include the mumble target name if provided to enable multi-instance.
- Detects the process ID from mumble link where possible.
- Lets you launch Guild Wars 2 using that mumble link name using the tray icon if one was provided to Blish HUD.

_Multiple instances of Blish HUD should each have their own settings directory specified._


Blish HUD/Program.cs
```diff
@@ -44,20 +46,23 @@ public static class Program {
...
+                                                       ? $"{APP_GUID}:{ApplicationSettings.Instance.MumbleMapName}"
+                                                       : $"{APP_GUID}");
+
+            if (!_singleInstanceMutex.WaitOne(TimeSpan.Zero, true)) {
                 Logger.Warn("Blish HUD is already running!");
                 return;
             }
 
-            // Needed by textboxes to enable CTRL + A selection
-            Application.EnableVisualStyles();
```
greaka, line 53: Why was this removed?

dlamkins: It was unrelated, but noticed while updating the mutex stuff. 

This was needed by the old WinForms textbox which is no longer used.

greaka (approved):

PR #248, opened 2020-06-16
gw2sharp [0.9.6] => [0.10.0]

agaertner:
(branch can be deleted after merge)
(also delete branch 'MoreFlowDirections')

### DIFF
--- a/Blish HUD/Blish HUD.csproj
+++ b/Blish HUD/Blish HUD.csproj
@@ -183,7 +183,7 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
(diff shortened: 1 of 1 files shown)


dlamkins:
@Archomeda could you remind me what the exact version match (`[0.9.6]`) for 0.9.6 was for again?

Archomeda:
That's to make sure that the NuGet variant of Blish HUD requires that exact version. Meaning that modules are required to use that specific version or it wouldn't otherwise be accepted when restoring packages.

Basically to prevent accidental incompatible version references, since Blish HUD includes (before this PR) version 0.9.6 and no other version, it wouldn't make sense for a module to reference a different version.

https://docs.microsoft.com/en-us/nuget/concepts/package-versioning#version-ranges

dlamkins (approved):
Thank you both!  @Archomeda @TybaIt

PR #253, opened 2020-06-19
added idea folder to gitignore

greaka:


dlamkins:
I like this and may end up trying Rider.

After learning about the idea folder [from the support site](https://rider-support.jetbrains.com/hc/en-us/articles/207097529-What-is-the-idea-folder-) it looks like there might be a couple of exceptions we'd want to make in there if we ever start taking advantage of some of the other Rider features.  Mind updating this with the more explicit items under the .idea folder provided [in this example .gitignore](https://github.com/JetBrains/resharper-rider-samples/blob/master/.gitignore)?

.gitignore
```diff
@@ -358,3 +358,6 @@ MigrationBackup/
 
 # Fody - auto-generated XML schema
 FodyWeavers.xsd
+
+# IntelliJ
+.idea/
```
dlamkins: For reference:
https://rider-support.jetbrains.com/hc/en-us/articles/207097529-What-is-the-idea-folder-
https://github.com/JetBrains/resharper-rider-samples/blob/master/.gitignore

dlamkins (changes requested):

greaka:
No problem, since we didn't check in any vs files so far, I opted for keeping them out of the tree

dlamkins:
Thanks!  Looks good.  Gonna give rider a try soon enough.  🙂

dlamkins (approved):

PR #265, opened 2020-06-30
Refresh Gw2Process on exception

### LINKED ISSUE #260: Process handle not refreshed and becomes stale
**Greaka at 6/25/2020 6:24:12 AM +00:00**
> okay, regarding <#534492012263636992>:
> the main window had a faulty handle, i assumed that bhud catched the login handle and didnt update when the game started
> so my temporary fix was changing `public IntPtr Gw2WindowHandle { get; private set; }` to
> ```cs
>         public IntPtr Gw2WindowHandle 
>         {
>             get
>             {
>                 this.Gw2Process.Refresh();
>                 return this.Gw2Process.MainWindowHandle;
>             }
>         }
> ```
>
> it refreshes the cached process object before accessing the windowhandle
> 
> before merging this into the dev branch, we should proabably think about a better way as there are 8 references, sometimes in controls that use this handle

**Source discussion from the #💢help channel:** https://discordapp.com/channels/531175899588984842/534492012263636992/725566715466874892

**Logs:**
[blishhud.20200624-211059.log](https://github.com/blish-hud/Blish-HUD/files/4829546/blishhud.20200624-211059.log)

> 21:23:45.7555|WARN|Blish_HUD.WindowUtil|GetClientRect failed with error code 1400.

REF: https://discordapp.com/channels/531175899588984842/536970543736291346/725597194371203092

dlamkins:
Fixes #260

### DIFF
--- a/Blish HUD/GameServices/GameIntegrationService.cs
+++ b/Blish HUD/GameServices/GameIntegrationService.cs
@@ -331,7 +331,9 @@ namespace Blish_HUD {
             this.IsInGame = Gw2Mumble.TimeSinceTick.TotalSeconds <= 0.5;
 
             if (this.Gw2IsRunning) {
-                switch (WindowUtil.UpdateOverlay(BlishHud.FormHandle, this.Gw2WindowHandle, this.Gw2HasFocus)) {
+                var updateResult = WindowUtil.UpdateOverlay(BlishHud.FormHandle, this.Gw2WindowHandle, this.Gw2HasFocus);
+
+                switch (updateResult.Response) {
                     case WindowUtil.OverlayUpdateResponse.WithFocus:
                         this.Gw2HasFocus = true;
                         break;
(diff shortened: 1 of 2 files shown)


Blish HUD/GameServices/GameIntegrationService.cs
```diff
@@ -341,7 +343,21 @@ public class GameIntegrationService : GameService {
...
-                        this.Gw2Process = null;
+                        switch (updateResult.ErrorCode) {
+                            case 1400:
+                                this.Gw2Process.Refresh();
+
+                                if (this.Gw2Process.MainWindowHandle == IntPtr.Zero) {
+                                    // Guild Wars 2 most likely closed
+                                    goto case -1;
+                                }
+
```
greaka, line 352: This shouldn't work. I believe the user was just lucky when they tested this. Either `Gw2WindowHandle` has to be set here again, or `Gw2Process` needs to be reset here again, or `Gw2WindowHandle` should be modified to point to `Gw2Process.MainWindowHandle`.

greaka (changes requested):

Blish HUD/GameServices/GameIntegrationService.cs
dlamkins: Good catch.  Updated Gw2WindowHandle to point to `Gw2Process.MainWindowHandle`.

greaka (approved):

PR #268, opened 2020-06-30
Make hook handlers thread-safe.

### LINKED ISSUE #262: Rare race condition in input hook managers.
**Freesnöw at 6/24/2020 9:09:29 AM +00:00**
> <@!95699952390385664> No rush since it's only happened once and it's low impact by the looks of it (I assume it's debug only), but on one of my launches, `DebugHelperMouseHookManager` threw the exception `Collection was modified; enumeration operation may not execute.` on line 18 (https://github.com/blish-hud/Blish-HUD/blob/7ad64de6d29d960ff4fca2985a5b6980ab70ba11/Blish%20HUD/GameServices/Input/Mouse/DebugHelperMouseHookManager.cs#L18).  I suspect it just needs locked while the handlers are initially added or something.  Wanted to see if you had any initial feedback or if I should dig into it some later and potentially stick a lock or something on `Handlers`.

REF: https://discordapp.com/channels/531175899588984842/536970543736291346/725276404689666079

dlamkins:
Fixes #262

### DIFF
--- a/Blish HUD/GameServices/Input/DebugHelperInputHookManager.cs
+++ b/Blish HUD/GameServices/Input/DebugHelperInputHookManager.cs
@@ -46,13 +46,9 @@ namespace Blish_HUD.Input {
             isEnabled = false;
         }
 
-        public virtual void RegisterHandler(THandlerDelegate handleInputCallback) {
-            this.Handlers.Add(handleInputCallback);
-        }
+        public virtual void RegisterHandler(THandlerDelegate handleInputCallback) { this.Handlers.Add(handleInputCallback); }
 
-        public virtual void UnregisterHandler(THandlerDelegate handleInputCallback) {
-            this.Handlers.Remove(handleInputCallback);
-        }
+        public virtual void UnregisterHandler(THandlerDelegate handleInputCallback) { this.Handlers.Remove(handleInputCallback); }
 
         protected abstract void HookCallback(TEventMessage message);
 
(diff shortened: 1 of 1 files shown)


Archomeda (changes requested):
The following still snippet throws `System.InvalidOperationException: 'Collection was modified; enumeration operation may not execute.'`.

```cs
using System;
using System.Collections.Generic;
using System.Threading;

namespace TestConsole
{
    internal static class Program
    {
        private static SynchronizedCollection<int> collection = new SynchronizedCollection<int>();

        private static void Main(string[] args)
        {
            new Thread(RunThreadA).Start();
            new Thread(RunThreadB).Start();
        }

        private static void RunThreadA()
        {
            for (var i = 0; i < 20; i++)
            {
                collection.Add(i);
                Console.WriteLine($"Added {i}");
                Thread.Sleep(500);
            }
        }

        private static void RunThreadB()
        {
            Thread.Sleep(1000);
            foreach (var item in collection)
            {
                Thread.Sleep(1000);
                Console.WriteLine($"Read {item}");
            }
        }
    }
}
```

~~Using a [`BlockingCollection<>`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.blockingcollection-1?view=netframework-4.7.2) instead would fix it (note the `.GetConsumingEnumerable()` method.~~
Okay, maybe I was thinking too quickly, that one might not be viable if it doesn't adhire the current support of removing hooks. I'll search a bit more. And also definitely not when `.GetConsumingEnumerable()` actually takes stuff out.

Last edit: Right. You can put a lock on `((ICollection)collection).SyncRoot` when iterating. This property is used when adding and removing items in `SynchronizedCollection` under the hood too, so you'll be fine there.
Also, don't forget to change the WinApi counterpart `WinApiInputHookManager` since it may happen there as well 😄

dlamkins:
Closing in favor of #270

PR #277, opened 2020-07-13
Module views

dlamkins:
- Adds new views for managing modules to replace the static builders.
- Fixed DrawUtil `WrapText` function (now correctly wraps text that already contains newlines).
- New module views introduce:
  - Clear module settings option.
  - Open registered directories from module screen.
  - Better module status (loading, unloading, faulted, loaded, disabled) accuracy.
  - Dependency checking before enabling.
    - Users can manually override this with a setting listed on the Dependencies view (Ignore module dependencies)
  - API permission consent.
  - Modify module permissions on the module's page.
- New setting renderers for both `float` and `int`.
- Settings now support "compliance requisites" which essentially allow us to add restraints to settings for when they're displayed in the UI (for instance, limit the minimum and maximum value of an `int` based setting).
  - `SettingEntry<int>` has `.SetRange(int minValue, int maxValue)`
  - `SettingEntry<float>` has `.SetRange(float minValue, float floatValue)`
  - In the future, enums will have have the option to limit which enum members are visible for the setting.
- `SettingCollection`s have a new bool property `RenderInUi` (default false) which determines if that SettingCollection will automatically render when it is a subcollection of a rendered SettingCollection.  If it is the root collection being rendered, this setting is ignored.
- `ContextMenuStrip` now has overloads `AddMenuItem(ContextMenuStripItem item)` and new method `AddMenuItem(IEnumerable<ContextMenuStrip> items)`.
- Setting saves are now much less aggressive (they were saving on every property change immediately) and will now save no more than once every 3 seconds.
- When not enabled, `ContextMenuStripItem` the bullet and dropdown arrow (if applicable) are now tinted grey like the text.
- `Scrollbar` now invalidates a bit less aggressively and also avoids a StackOverflowException caused by Invalidate getting called recursively in specific rare circumstances.
- `TrackBar` control has been revised.
  - Now appears much closer to the look of the actual Guild Wars 2 control.
  - **[BREAKING]** `MinValue` and `MaxValue` values have had their value types changed to floats (previously int).
  - **[BREAKING]** `RoundedValue` has been removed (in favor of the `SmallStep` property described below).
  - **[BREAKING]** The `ValueChanged` event has had its event arg type changed to `ValueEventArgs<float>`.
     - These breaking changes should have minimal to no impact - I am not aware of the `TrackBar` control being used by any modules at this time.
  - Added `SmallStep` bool property (default false).
     - When `true`, the TrackBar will report non-whole values (float precision) between the min and max values.
     - When `false`, the TrackBar will snap to whole integers between the min and max values.
  - Holding CTRL while dragging the TrackBar nub will force it to snap to whole integers (as if `SmallStep` was disabled) for the duration that CTRL is held.

See the module settings view in action (this was recorded prior to the final layout changes which can be seen in the images further down): https://gfycat.com/ShowyHopefulItaliangreyhound

Disabled module with full set of API Permissions listed and a module listed as a dependency:
image

Enabled module with no permissions requested, dependent Blish HUD version listed, and a single float setting automatically displayed.
image

Similar to the prior with 3 float settings automatically displayed (with compliance values set in the module using `.SetRange(..)`.
image

Module settings dropdown ("Clear Settings" is disabled if the module is active):
image


Blish HUD/GameServices/Modules/UI/Presenters/ManageModulePresenter.cs
```diff
@@ -0,0 +1,230 @@
...
+        private string GetModuleAuthor() {
+            if (this.Model.Manifest.Contributors?.Count > 0) {
+                return string.Join(", ", this.Model.Manifest.Contributors.Select(c => c.Name));
+            }
+
+            if (this.Model.Manifest.Author != null) {
+                return this.Model.Manifest.Author.Name;
+            }
+
+            return Strings.Common.Unknown;
```
greaka: If you want to, then you could change this to 
```cs
return this.Model.Manifest.Author?.Name ?? Strings.Common.Unknown;
```

Blish HUD/GameServices/Modules/UI/Presenters/ManageModulePresenter.cs
```diff
@@ -0,0 +1,230 @@
...
+            SubscribeToModuleRunState();
+        }
+
+        private void ViewOnDisableModuleClicked(object sender, EventArgs e) {
+            this.Model.Enabled = false;
+        }
+
+        private void SubscribeToModuleRunState() {
+            if (this.Model.ModuleInstance != null) {
+                this.Model.ModuleInstance.ModuleRunStateChanged += ModuleInstanceOnModuleRunStateChanged;
```
greaka, line 175: You run the risk of subscribing multiple times to the same event here.
Maybe change it to
```cs
this.Model.ModuleInstance.ModuleRunStateChanged -= ModuleInstanceOnModuleRunStateChanged;
this.Model.ModuleInstance.ModuleRunStateChanged += ModuleInstanceOnModuleRunStateChanged;
```

dlamkins: Oh, good catch!  Yeah, that would cause some wonky problems really quick.

Blish HUD/GameServices/Modules/UI/Presenters/ManageModulePresenter.cs
dlamkins: Revised.

Blish HUD/GameServices/Modules/UI/Presenters/ManageModulePresenter.cs
dlamkins: Revised.

dlamkins:
These were resolved, but GitHub has failed to update the PR with the changes from my local branch, so waiting to fix that first.

dlamkins:
GitHub has caught up and is now showing the correct status so it should be ready for review again. @greaka

greaka (approved):

PR #287, opened 2020-07-25
Feat/audio cscore

dlamkins:
Uses the game audio levels to attempt to match our audio output levels.


Blish HUD/GameServices/GameIntegration/AudioIntegration.cs
```diff
@@ -0,0 +1,117 @@
...
+        public float AverageGameVolume {
+            get {
+                float total = 0;
+                for (int i = 0; i < _audioPeakBuffer.InternalBuffer.Length; i++) {
+                    total += _audioPeakBuffer.InternalBuffer[i];
+                }
+
+                return MathHelper.Clamp(total / _audioPeakBuffer.InternalBuffer.Length, 0, MAX_VOLUME);
+            }
+        }
```
greaka: I would only let this compute the average if there is a need. So have a backing field that gets populated if null.
You could reset it to null at line 83. Especially in debug runs where this is called each frame, we could save some cycles 🙂

Blish HUD/GameServices/GameIntegration/AudioIntegration.cs
```diff
@@ -0,0 +1,117 @@
...
+            _timeSinceCheck += gameTime.ElapsedGameTime.TotalMilliseconds;
+            
+            if (_timeSinceCheck > CHECK_INTERVAL) {
+                _timeSinceCheck -= CHECK_INTERVAL;
+
+                _audioPeakBuffer.PushValue(_processMeterInformation.GetPeakValue());
+            }
+        }
+
+        private IEnumerable<(Process Process, AudioMeterInformation MeterInformation)> GetProcessMeters() {
```
greaka, line 99: Please add a doc comment to this function that `Process` is going to be discarded when advancing. This function is private, but we would forget about this in the future. When saving `Process` to use for after iterating over the meters, you could end up with a null exception.

greaka (changes requested):

Blish HUD/GameServices/GameIntegration/AudioIntegration.cs
dlamkins: Good eye.

greaka (approved):

PR #288, opened 2020-07-25
Begin modularizing GameIntegrationService

dlamkins:
This PR begins the tedious task of cleaning up the `GameIntegrationService`.

Some services have grown to be a bit too monolithic / bloated.  `GameIntegrationService` and `OverlayService` are our most heavily used offenders.  This PR works to clean some of that up by splitting their functionality into smaller modules that they manage.  ServiceModules can become more fully integrated with services in the future, but for now, this:
- Splits TacO integration (scooting over our icon when TacO is running) into it's own class (removing it from the `OverlayService` update loop).
- Puts the notify icon and form wrapping code (used to hide Blish HUD from the taskbar) into its own class.

Future work can move the logic used to identify the correct GW2 process to its own sub-module as well as moving the logic used to overlay GW2 to its own sub-module.  This will help make it more clear where certain functionality is defined, separate the logic, and make it easier to expand on individual components in the future.


Blish HUD/GameServices/OverlayService.cs
```diff
@@ -229,18 +223,10 @@ public class OverlayService : GameService {
...
-                    } else {
-                        CornerIcon.LeftOffset = _clientType == Gw2ClientContext.ClientType.Chinese ? 36 : 0;
-                    }
-
-                    _lastTacoCheckTime = 0;
+                if (GameIntegration.TacO.TacOIsRunning) {
+                    CornerIcon.LeftOffset = 36 * (_clientType == Gw2ClientContext.ClientType.Chinese ? 2 : 1);
+                } else {
+                    CornerIcon.LeftOffset = _clientType == Gw2ClientContext.ClientType.Chinese ? 36 : 0;
                 }
```
greaka: I need some help here, why is it 36 and not `CornerIcon.ICON_SIZE`?
In addition, following https://discordapp.com/channels/384735285197537290/603928533584183310/736854069028585472 I would like to simplify this code.

What do you think about this:
```cs
var offset = _clientType == Gw2ClientContext.ClientType.Chinese ? 1 : 0;
offset += GameIntegration.TacO.TacOIsRunning ? 1 : 0;

CornerIcon.LeftOffset = offset * 36;
```

greaka (commented):
Great idea! It got silently more bloated over time. I like the `ServiceModule` approach.

Blish HUD/GameServices/OverlayService.cs
dlamkins: We found early on that the TacO icon did not look good with our standard static offset.  It needed a manual adjustment. 

I'll look at simplifying that section.

dlamkins: Done!

greaka (approved):

greaka (commented):
test

greaka (commented):
For some reason, Github broke approving. I am sorry 🙁 
image

EDIT: It's just too early in the morning 🙈 
Didn't catch the merge.

PR #290, opened 2020-08-04
Add PositionRaw

greaka:
After a lot of testing, I can confidently say that neither the change implemented here, nor switching over to full mouse input handling via the system events would have any reasonable impact for us. Every modification changes processing times within 1 microsecond with an accuracy of 100 nanoseconds.

For comparison: Moving the mouse across a border of a control spikes the time to process this input to over 200 microseconds.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -8,6 +8,7 @@ tab_width=4
 indent_style=space
 indent_size=4
 tab_width=4
+csharp_prefer_braces=true:hint
 
 [*.{appxmanifest,build,config,csproj,dbml,discomap,dtd,json,jsproj,lsproj,njsproj,nuspec,proj,props,resjson,resw,resx,StyleCop,targets,tasks,vbproj,xml,xsd}]
 indent_style=space
--- a/Blish HUD/GameServices/DebugService.cs
+++ b/Blish HUD/GameServices/DebugService.cs
@@ -160,6 +160,9 @@ namespace Blish_HUD {
         #endregion
 
         #region Debug Overlay
+        
+        private ConcurrentDictionary<string, Func<GameTime, string>> _overlayTexts;
+        public ConcurrentDictionary<string, Func<GameTime, string>> OverlayTexts => _overlayTexts;
 
         public void DrawDebugOverlay(SpriteBatch spriteBatch, GameTime gameTime) {
             int debugLeft = Graphics.WindowWidth - 600;
@@ -171,10 +174,9 @@ namespace Blish_HUD {
                 spriteBatch.DrawString(Content.DefaultFont14, $"{timedFuncPair.Key} {Math.Round(timedFuncPair.Value.GetAverage())} ms", new Vector2(debugLeft, 50 + (i++ * 25)), Color.Orange);
             }
 
-            spriteBatch.DrawString(Content.DefaultFont14, $"3D Entities Displayed: {Graphics.World.Entities.Count}",              new Vector2(debugLeft, 50 + (i++ * 25)), Color.Yellow);
-            spriteBatch.DrawString(Content.DefaultFont14, "Render Late: "            + (gameTime.IsRunningSlowly ? "Yes" : "No"), new Vector2(debugLeft, 50 + (i++ * 25)), Color.Yellow);
-            spriteBatch.DrawString(Content.DefaultFont14, "ArcDPS Bridge: "          + (ArcDps.RenderPresent ? "Yes" : "No"),     new Vector2(debugLeft, 50 + (i++ * 25)), Color.Yellow);
-            spriteBatch.DrawString(Content.DefaultFont14, "Average In-Game Volume: " + GameIntegration.Audio.AverageGameVolume,   new Vector2(debugLeft, 50 + (i++ * 25)), Color.Yellow);
+            foreach (Func<GameTime, string> func in _overlayTexts.Values) {
+                spriteBatch.DrawString(Content.DefaultFont14, func(gameTime), new Vector2(debugLeft, 50 + (i++ * 25)), Color.Yellow);
+            }
         }
 
 #endregion
@@ -191,6 +193,12 @@ namespace Blish_HUD {
 
         protected override void Load() {
             _funcTimes = new ConcurrentDictionary<string, DebugCounter>();
+
+            _overlayTexts = new ConcurrentDictionary<string, Func<GameTime, string>>();
+            _overlayTexts.TryAdd("entityCount", (_) => $"3D Entities Displayed: {Graphics.World.Entities.Count}");
+            _overlayTexts.TryAdd("renderLate", (gameTime) => "Render Late: " + (gameTime.IsRunningSlowly ? "Yes" : "No"));
+            _overlayTexts.TryAdd("arcDps", (_) => "ArcDPS Bridge: " + (ArcDps.RenderPresent ? "Yes" : "No"));
+            _overlayTexts.TryAdd("volume", (_) => "Average In-Game Volume: " + GameIntegration.Audio.AverageGameVolume);
         }
 
         protected override void Update(GameTime gameTime) {
--- a/Blish HUD/GameServices/DebugService.cs
+++ b/Blish HUD/GameServices/DebugService.cs
@@ -163,7 +163,7 @@ namespace Blish_HUD {
 
         #region Debug Overlay
 
-        public ConcurrentDictionary<string, Func<GameTime, string>> OverlayTexts { get; private set; }
+        public OverlayStrings OverlayTexts { get; private set; }
 
         public void DrawDebugOverlay(SpriteBatch spriteBatch, GameTime gameTime) {
             int debugLeft = Graphics.WindowWidth - 600;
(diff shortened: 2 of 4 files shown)


greaka:
To add a comment to the debug service change: you can simply use it like this now:
```cs
var niceText = $"not so nice text {localVar}";
GameService.Debug.OverlayTexts["foo"] = _ => niceText;

GameService.Debug.OverlayTexts["elapsed"] = gameTime => $"elapsed time: {gameTime.ElapsedMilliseconds}";

GameService.Debug.OverlayTexts["stateVar"] = _ => $"ArcBridge: {GameService.Arcdps.Renderpresent}";
```

Blish HUD/GameServices/Input/Mouse/MouseHandler.cs
```diff
@@ -74,8 +76,8 @@ public class MouseHandler : IInputHandler {
             MouseState rawMouseState = Mouse.GetState();
 
             this.State = new MouseState(
-                                        (int)(rawMouseState.X / GameService.Graphics.UIScaleMultiplier),
-                                        (int)(rawMouseState.Y / GameService.Graphics.UIScaleMultiplier),
+                                        (int) (rawMouseState.X / GameService.Graphics.UIScaleMultiplier),
+                                        (int) (rawMouseState.Y / GameService.Graphics.UIScaleMultiplier),
```
dlamkins: Any reason to continue to use the mouse position we're polling for if we've got the raw position?

Blish HUD/GameServices/Input/Mouse/MouseHandler.cs
```diff
@@ -25,6 +25,8 @@ public class MouseHandler : IInputHandler {
         /// </summary>
         public Point Position => this.State.Position;
 
+        public Point PositionRaw { get; private set; }
```
dlamkins: Seems redundant to have an accurate position and a non-accurate position.  Should we just use this as the base of our positions?

`Position` needs to be scaled value as that is used for UI positioning.

If not, at the least I'd like to have `PositionRaw` flipped around to `RawPosition` or better yet `ScreenPosition` or `DesktopPosition`.

Blish HUD/GameServices/DebugService.cs
```diff
@@ -161,23 +164,26 @@ public class DebugService:GameService {
 
         #region Debug Overlay
 
+        private ConcurrentDictionary<string, Func<GameTime, string>> _overlayTexts;
+        public  ConcurrentDictionary<string, Func<GameTime, string>> OverlayTexts => _overlayTexts;
```
dlamkins: I think we'd want to have functions within the DebugService for adding and removing entries, if we want it to be publicly accessible.  If you don't want to go that far, I don't see any reason to keep the public member for now (since if we do add methods for adding/removing later, we'd need to deprecate this member anyways).

Blish HUD/GameServices/Input/Mouse/MouseHandler.cs
```diff
@@ -126,9 +128,12 @@ public class MouseHandler : IInputHandler {
...
 
         public bool HandleInput(MouseEventArgs mouseEventArgs) {
-            if (mouseEventArgs.EventType == MouseEventType.MouseMoved) return false;
+            if (mouseEventArgs.EventType == MouseEventType.MouseMoved) {
+                PositionRaw = new Point(mouseEventArgs.PointX, mouseEventArgs.PointY);
+                return false;
+            }
 
-            if (_cameraDragging && (mouseEventArgs.EventType == MouseEventType.RightMouseButtonReleased))
+            if (_cameraDragging && mouseEventArgs.EventType == MouseEventType.RightMouseButtonReleased)
```
dlamkins: I know this doesn't originate in this PR (it originated in Archomeda's when he revamped the InputService), but could you put the braces back on this conditional?  I dislike skipping on the braces for one liners unless it's for an early exit at the top, but this also has an else if that does have braces which makes it exceptionally bleck.

dlamkins (changes requested):

Blish HUD/GameServices/Input/Mouse/MouseHandler.cs
greaka: I did not remove them by hand, so I assume it is a setting in editorconfig? I can put them back in ofc.

Blish HUD/GameServices/DebugService.cs
greaka: I thought about it when implementing it, as I wanted to have extra functions too. In the end I just went with KISS, since all those functions would have been one liners anyway. Why would you want to remove direct access to the dictionary? Is there a reason why we should restrict getting the values from the dictionary?

How would you feel about a wrapper type? So that indexing still works and we can expose whatever we want.

Blish HUD/GameServices/Input/Mouse/MouseHandler.cs
greaka: I had a look at our current method of handling input and saw 3 options:

1. dump `GetState()` and use the windows input event handling system (my original plan)
To keep it non-breaking, I would have had to make a bigger change to the whole handling, since in its current form, this would result in a lot of funky behavior. In the update loop, we only process the last event sent to us, which is fine, since we filter for mouse clicks currently and they are spaced out wide enough for us to catch every single one of them. Mouse move events would shadow clicks easily this way.
After evaluating the impact of `GetState()`, I figured if it would be faster at all, the change wouldn't be noticeable.
2. dump `GetState()`, but just ignore everything except left mouse click, right mouse click and scroll wheel. This is the breaking change I mentioned earlier.
3. Let everything be what it currently is and just expose the raw values for anyone who needs them.

I am interested in your opinion here.

Blish HUD/GameServices/Input/Mouse/MouseHandler.cs
greaka: See here https://github.com/blish-hud/Blish-HUD/pull/290#discussion_r465554001

Blish HUD/GameServices/Input/Mouse/MouseHandler.cs
dlamkins: Oh, the braces, not the parentheses.

Blish HUD/GameServices/DebugService.cs
dlamkins: Ultimately I didn't want the backend implementation directly exposed in case we wanted to make changes later since those would be breaking.  If it's abstracted by a function and/or a wrapper type, we can modify it down the road without actually causing a breaking change.

greaka:
Now this is a little messy all in one PR, but I ran the Jetbrains code cleanup and this is what came out of it.
The braces were introduced, but a lot of other stuff was changed too.

.editorconfig
```diff
@@ -8,6 +8,7 @@ tab_width=4
 indent_style=space
 indent_size=4
 tab_width=4
+csharp_prefer_braces=true:hint
```
dlamkins, line 11: Ahh, thank you for this!

greaka: It has its flaws:
https://github.com/blish-hud/Blish-HUD/pull/290/commits/1656d70e9b39b3119117460934d8602aa2c746d8#diff-867794a57e040a77f0b88399e707dbbbR104-R106

Blish HUD/GameServices/Input/Mouse/MouseHandler.cs
dlamkins: I dislike item 3 simply because it adds an extra value to decide if it should be used and trying to explain when it would be best to use one or the other seems like a hard line to draw.  I imagine most folks would want the more raw and accurate value.

For instance, is there something specifically wrong with using:

```cs
this.State = new MouseState((int)(this.PositionRaw.X / GameService.Graphics.UIScaleMultiplier),
							(int)(this.PositionRaw.Y / GameService.Graphics.UIScaleMultiplier),
							_mouseEvent?.WheelDelta ?? 0,
							rawMouseState.LeftButton,
							rawMouseState.MiddleButton,
							rawMouseState.RightButton,
							rawMouseState.XButton1,
							rawMouseState.XButton2);
```

Perhaps I am missing where the breaking change is that you're mentioning.

.editorconfig
dlamkins: 😕 Well, I suppose that's collateral damage that can maybe be addressed with further tweaking of the config.  In any case, I still appreciate you throwing that in.  👍

Blish HUD/GameServices/Input/Mouse/MouseHandler.cs
greaka: Where does rawMouseState come from? The event does not provide the full mouse state, but only the fraction that changed. So a mouse move event only contains the new position of the mouse.

dlamkins: It's defined a few lines above the `this.State` assignment.

greaka: I am a little confused. What I see is this:
```cs
            MouseState prevMouseState = this.State;

            MouseState rawMouseState = Mouse.GetState();
```
which gets `rawMouseState` from the `GetState()` method. Did I understand you correctly that you want to get rid of `Mouse.GetState()`?

dlamkins: Not necessarily the entire `GetState()` since we're using it for those other button states.  Ideally we would have only one source and reference for the mouse position, so if we have a more accurate or "raw" mouse position, I would think it would be more ideal to use that instead of also using the polled value from `GetState()`.  That way we don't need `.PositionRaw` and can instead use `.Position` (which is just `this.State.Position`) for everything.

greaka: Oh okay, that's no problem.

dlamkins (approved):
Looking good!

PR #294, opened 2020-08-16
Use Cubic Hermite Interpolation for paths

kennipj:
Recreates the trail path by sampling from the Hermite interpolated path, resulting in a smooth trail. Should also fix weird behaviour that occurs when points are too close to each other, not 100% confirmed.

`numPoints` decides the number of points to sample in between trail points. May want to make this depend on the distance between points.
`alpha` can be changed to change smoothness of the curve, 0.5 seems like the best fit for now.


dlamkins (approved):

dlamkins:
Looks good — Thanks for all of your intense work on this!

PR #295, opened 2020-08-16
updated to gw2sharp 0.11.0

greaka:


dlamkins (approved):
I tested the PR locally and confirmed that the rate-limiting still works without issue.  I was surprised to see that v0.11.0 introduced far fewer breaking changes (that we would have to account for) than I anticipated.  Happy to see that!

dlamkins:
I've opened #297 to remind us to switch the rate-limiting method over to using a middleware when we have time.

PR #298, opened 2020-08-17
Improved effect management

dlamkins:
Makes several improvements:
- Improves EntityEffect class to reduce duplicate code.
- Moves trail effect details into an Effect class to keep a single instance of the effect.
- Adds tint support to both markers and trails (matching TacO implementation).
- Adds custom title color for markers.
- Removed unused Mesh behavior.

Moving to TrailEffect and MarkerEffect is part of an effort to later extract certain matrix details so that they're passed via the Graphics service during the draw loop instead of pulled directly from the Gw2Mumble service.  This will allow us to later render objects to multiple "worlds" (i.e. main and mini-map) by just passing a different view matrix.


kennipj (approved):
Everything looks fine to me

PR #302, opened 2020-08-29
Adds post processing support to trails

kennipj:
You can now set the post processing functions to be used when initialising the trail, by setting the `PostProcessFunctions` property of `ScrollingTrail` and `ScrollingTrailSection`.

Example usage:
```cs
this.ManagedEntity.PostProcessFunctions = new List<Func<List<Vector3>, List<Vector3>>>() { x => x.SetResolution(30f) };
```

Also moved `SetTrailResolution` to `Vector3Extensions` and renamed it to `SetResolution`.


kennipj:
Needs more changes, post processing does not implement nicely, as there is no way to change the post processing method after initialisation, without applying it ontop of the already post-processed trail.

dlamkins (approved):

Issue #181, opened 2020-03-29
--pid launch option


BlishHUD:
**QuitarHero at 3/29/2020 1:40:02 PM +00:00**
> So going through the Launch Paramaters, and for the most part they work. The only issue I'm having is as such:
> - `--process` doesn't really work all that well for programs that have multiple instances of themselves open, such as Discord & Chrome. This is probably because the paramater defaults to a PID that the user can't specify, so while it most likely works, you just don't *see*  it, making it kinda worthless.
> Since you're recommending AutoHotkey for these kinds of launch paramaters, you may as well include a `--pid` (optional) paramater as well, to specify the exact instance of the program you want to overlay Blish HUD on.

REF: https://discordapp.com/channels/531175899588984842/534492173362528287/693816745273589780

PR #307, opened 2020-10-01
Add --pid launch option

TheMrMilchmann:
Adds the "--pid" launch option as described in #181. Additionally, I set the shortcut to 'P' (`p` is used by `process`).

Motivation: When multiple instances of the game are already open, the file-handle used for MumbleLink is not necessarily known. Further, all instances use the same process name. The `--pid` launch option could be used to launch Blish-HUD for a specific process.

Process discovery uses the following precedence:
1. PID reported by MumbleLink (if available)
2. PID specified by --pid (if available)
3. ProcessName-based lookup

I tried to follow the code-style and formatting used in the project. Please let me know what you think of the implementation. Also, I couldn't find any tests for this part of the logic and assumed it is tested manually. If there are automated tests, please point me in the right direction.

(If this is merged, it should probably also be added to https://github.com/blish-hud/Blish-HUD/wiki/Launch-Options. I would have done it myself but there does not seem to be a way to open PRs against wikis.)

### DIFF
--- a/Blish HUD/ApplicationSettings.cs
+++ b/Blish HUD/ApplicationSettings.cs
@@ -37,6 +37,12 @@ namespace Blish_HUD {
         #region Game Integration
 
         [
+            OptionParameter("pid", 'P'),
+            Help("The PID of the process to overlay.")
+        ]
+        public int ProcessId { get; private set; } = 0;
+
+        [
             OptionParameter("process", 'p'),
             Help("The name of the process to overlay (without '.exe').")
         ]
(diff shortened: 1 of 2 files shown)


Blish HUD/ApplicationSettings.cs
```diff
@@ -36,6 +36,11 @@ public class ApplicationSettings : BaseCliArguments {
 
         #region Game Integration
 
+        [
+            OptionParameter("pid", 'P'),
+            Help("The PID of the process to overlay.")
+        ]
+        public int ProcessId { get; private set; } = 0;
```
dlamkins, line 43: Minor, but could we get a newline after this property line?

dlamkins (changes requested):

Blish HUD/ApplicationSettings.cs
TheMrMilchmann: Sure thing, done!

dlamkins (approved):

PR #319, opened 2020-10-15
Add PowerShell to launchsettings for easier overlay debugging

Archomeda:
Just noticed that this was not yet added. Having this launch setting shared makes it easier to debug the overlay without having GW2 open.


dlamkins (approved):
Yep.  Seems good to me.  👍

PR #320, opened 2020-10-15
Use GitVersion action instead of a custom PowerShell script

Archomeda:
Because [GitHub started deprecating set-env in actions](https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/), let's try out using the dedicated GitVersion action instead of using a custom PowerShell script to version Blish HUD.

This cleans up the workflow yaml files a bit and makes it more readable.

Update: Because msbuild also used add-path, I also had to update this action to v1.0.2 in order to fix that too.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,36 +13,40 @@ jobs:
     runs-on: windows-latest
     steps:
       - uses: actions/checkout@v2
-      - run: git fetch --prune --unshallow
-      - uses: microsoft/setup-msbuild@v1.0.0
+        with:
+          fetch-depth: 0
+      - uses: microsoft/setup-msbuild@v1.0.2
 
-      - name: Patch nightly version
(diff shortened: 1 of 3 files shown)


Archomeda:
It seems that this PR is currently blocked by GitTools/actions#220.
Additionally, the GitVersion action itself is also using set-env, which is what I was hoping to move away from. But since it's already been reported in GitTools/actions#213, I assume it will be fixed in an upcoming version.

Will revisit later.

Archomeda:
I copied over the GitHub actions code from Archomeda/Gw2Sharp#75 in order to manually create the AssemblyVersionInfo.cs file while we wait (for 2.5 months now already) for GitTools to reply on GitTools/actions#220...

This should solve the broken GitHub actions that happened earlier this month when they removed support for set-env and add-path after only a month of deprecation I believe (it happened too fast imo).

dlamkins (approved):

Issue #327, opened 2020-10-19
SpaceTamer's Bug Report [767534112709148703]


BlishHUD:
**SpaceTamer at 10/18/2020 11:46:32 PM +00:00**
> <@!164832069401182210> I am getting this with the recent commits: https://i.imgur.com/LioGpjR.png

https://imgur.com/LioGpjR

REF: https://discordapp.com/channels/531175899588984842/534492173362528287/767534112709148703

PR #328, opened 2020-10-19
Fix null reference on GetSessionEnumerator()

dlamkins:
Prevent race condition when getting audio device manager allowing it to be null before the first time it is needed.  Also, early exit if the device manager is null.


dlamkins:
Waiting on @TybaIt to confirm if this resolves the intermittent issues he was seeing and that a volume level is still reported every launch.

dlamkins:
Tybalt confirmed that the changes did resolve the issue for him.

nickcode122 (approved):

dlamkins:
Fixes #327

PR #330, opened 2020-10-20
ArcDps documentation

agaertner:


Blish HUD/GameServices/ArcDps/SocketListener/RawCombatEventArgs.cs
```diff
@@ -7,7 +7,13 @@ public class RawCombatEventArgs : EventArgs {
 
         public enum CombatEventType {
 
+            /// <summary>
+            /// ArcDps calculations and display values.
+            /// </summary>
             Area,
+            /// <summary>
+            /// Local recording of just the own character similar to the in-game combat log in the chat window.
```
dlamkins: > Local recording of just the own character

This is oddly worded.  Is it meant to say "Local recording of just _your_ own character..."

Otherwise, everything else looks great.  👍

dlamkins (changes requested):

TybaIt:
Fixed and added some missing doc.

Blish HUD/GameServices/ArcDps/SocketListener/RawCombatEventArgs.cs
agaertner: Fixed and added some missing doc.

Blish HUD/GameServices/ArcDps/Models/Ag.cs
```diff
@@ -5,11 +5,29 @@
     /// </summary>
     public class Ag {
 
+        /// <summary>
+        /// Agent name. May be null. Valid only at time of event.
```
greaka: Due to us copying the data over, the agent name is valid for as long as you need it. I would suggest removing the last sentence.

Blish HUD/GameServices/ArcDps/Models/CombatEvent.cs
```diff
@@ -5,11 +5,29 @@
     /// </summary>
     public class CombatEvent {
 
+        /// <summary>
+        /// The current combat event.
```
greaka: Just a minor nitpick, I would call it `The event data` or similar, due to the delay, current feels out of touch to me.

Blish HUD/GameServices/ArcDps/Models/Ev.cs
```diff
@@ -1,36 +1,117 @@
 ﻿namespace Blish_HUD.ArcDps.Models {
 
     /// <summary>
-    ///     Infos and data about the combat event. Have a look at the arcdps plugin documentation for how to use it
+    ///     Infos and data about the combat event. For more information see the arcdps plugin documentation.
```
greaka: We could maybe add a link to the docs here: https://deltaconnected.com/arcdps/api/

Blish HUD/GameServices/ArcDps/Models/Ev.cs
```diff
@@ -1,36 +1,117 @@
...
     /// </summary>
     public class Ev {
 
+        /// <summary>
+        /// Time when the event was registered.
```
greaka, line 12: We should note here that the time is some windows specific time since system boot, not the actual time in a known format. You can compare and convert it by looking up the windows docs for that timestamp.

Blish HUD/GameServices/ArcDps/Models/Ev.cs
```diff
@@ -1,36 +1,117 @@
...
+        /// <summary>
+        /// Current affinity of SrcAgent and DstAgent (friend = 0, foe = 1 or unknown = 2).
+        /// </summary>
         public byte   Iff             { get; }
+        /// <summary>
+        /// TRUE if buff was applied, removed or damaging. Otherwise FALSE.
+        /// </summary>
         public bool   Buff            { get; }
+        /// <summary>
+        /// Physical Hit Result (normal hit = 0, was critical = 1, was glance =  2, was blocked = 3, was evaded = 4, interrupted the target = 5, was absorbed = 6, missed = 7, killed the target = 8, downned the target = 9).
```
greaka: little spelling mistake here: `downned`

Blish HUD/GameServices/ArcDps/Models/Ev.cs
```diff
@@ -1,36 +1,117 @@
...
+        /// <summary>
+        /// TRUE if DstAgent is below 50% health. Otherwise FALSE.
+        /// </summary>
         public bool   IsFifty         { get; }
+        /// <summary>
+        /// TRUE if SrcAgent is moving at time of event. Otherwise FALSE.
+        /// </summary>
         public bool   IsMoving        { get; }
+        /// <summary>
+        /// TRUE if a state change occured (SrcAgent is now alive, dead, downed and other ambigious stuff ex. when SrcAgent is self, DstAgent is a reward id and Value is a reward type such as a wiggly box.). Otherwise FALSE.
```
greaka: little spelling mistake: `ambigious`

Blish HUD/GameServices/ArcDps/Models/Ev.cs
```diff
@@ -1,36 +1,117 @@
...
+        /// <summary>
+        /// TRUE if SrcAgent is flanking DstAgent at time of event. Otherwise FALSE.
+        /// </summary>
         public bool   IsFlanking      { get; }
+        /// <summary>
+        /// TRUE if all or part of damage was VS. barrier/shield. Otherwise FALSE.
+        /// </summary>
         public bool   IsShields       { get; }
+        /// <summary>
+        /// FALSE if buff damage happened during tick. Otherwise TRUE.
```
greaka: Minor nitpick: All the other summaries start with what's up when the field is true, this is the only one inverted. Would be nice to have consistency there.

Blish HUD/GameServices/ArcDps/SocketListener/RawCombatEventArgs.cs
```diff
@@ -7,7 +7,13 @@ public class RawCombatEventArgs : EventArgs {
 
         public enum CombatEventType {
 
+            /// <summary>
+            /// ArcDps calculations and display values.
+            /// </summary>
             Area,
+            /// <summary>
+            /// Local recording of the player character bound to the signed-in account of this GW2 instance similar to the in-game combat log in the chat window.
+            /// </summary>
```
greaka: Me being picky again.
I'd say it'd be better to distinguish the difference between the two better: Area is calculated and shows displayed estimates, Local is an exact recording only available for the player's character.

greaka (dismissed):
Thank you for documenting it! I really like some of the wordings. I marked a few things that could be improved.

In addition, what I didn't mark: There is a lot of info in parenthesis that could be moved to the `<remarks>` tag. Basically everything in parenthesis, but this is personal taste. @dlamkins What is your stance on that?

dlamkins:
I agree with @greaka's further review.  Tackle those quickly and we should be good to merge!  🙂

agaertner:
I am not sure exactly what the ``Ag.Team`` or ``Revision`` fields do and/or if the other ``Ev.Pad`` fields are also garbage such as ``Pad64``.
**Ag.Team**
I know it is a unique identifier but the documentation just states a patch date of GW2 and I barely remember which september it was so I could look it up what feature this was.
**Revision**
I know it's a header value. Not sure exactly why we have this field exactly but yeah. It's from the logging stuff. Probably the length?
**Pad fields**
The official docs and mechanics plugin have only a note at the ``Pad64`` field that this one is for internal tracking and thus garbage. Not sure about the other pad fields. Might be the case for them as well.

dlamkins:
> I am not sure exactly what the `Ag.Team` or `Revision` fields do and/or if the other `Ev.Pad` fields are also garbage such as `Pad64`.
> **Ag.Team**
> I know it is a unique identifier but the documentation just states a patch date of GW2 and I barely remember which september it was so I could look it up what feature this was.
> **Revision**
> I know it's a header value. Not sure exactly why we have this field exactly but yeah. It's from the logging stuff. Probably the length?
> **Pad fields**
> The official docs and mechanics plugin have only a note at the `Pad64` field that this one is for internal tracking and thus garbage. Not sure about the other pad fields. Might be the case for them as well.

@greaka we need anything else before merge?

Blish HUD/GameServices/ArcDps/Models/Ev.cs
```diff
@@ -42,63 +57,78 @@ public class Ev {
...
         /// </summary>
         public bool   IsShields       { get; }
         /// <summary>
-        /// FALSE if buff damage happened during tick. Otherwise TRUE.
+        /// TRUE if no buff damage happened during tick. Otherwise FALSE.
         /// </summary>
         public bool   IsOffCycle      { get; }
         /// <summary>
-        /// Buff instance id of buff applied. Zero if buff damage happened during tick, non-zero otherwise.
+        /// Buff instance id of buff applied. Non-zero if no buff damage happened during tick. Otherwise zero.
```
greaka: no buff damage during tick sounds a little confusing to me.
What do you think about `TRUE if buff event did not happen as part of the tick` or something along the lines, not really sure tbh.

greaka (commented):
Looks good to me otherwise.

greaka:
I have no idea anymore what revision was for.

The pad fields are fields to pad to 64 byte objects in the client, they weren't relevant at the time of writing the bridge. However, they are reserved for future use.

Blish HUD/GameServices/ArcDps/Models/Ev.cs
agaertner: Yeah it's because of the double negation. (``Non-zero`` and ``no damage``)  I thought it sounded weird but it's the change you wanted.
I am not sure if ``TRUE if buff event did not happen as part of the tick`` is the same as ``no buff damage happening`` because a buff can cause 0 damage and be relevant.

agaertner:
> 
> 
> I have no idea anymore what revision was for.
> 
> The pad fields are fields to pad to 64 byte objects in the client, they weren't relevant at the time of writing the bridge. However, they are reserved for future use.

~~Oh now that I think of it. Revision may just be an indicator if the data is valid and all operations succeeded. It seems to be just a validation byte. 
Here's a tiny vague note from the *GW2-ArcDPS-Mechanics-Log* plugin
``/* combat event logging (revision 1, when header[12] == 1) */``
at 
[arcdps_datastructures.h#L102](https://github.com/MarsEdge/GW2-ArcDPS-Mechanics-Log/blob/d068f8113109893d9d9c06796a7ff04297313ac2/src/arcdps_datastructures.h#L102)~~
Revision is actually the version of the format of the data structure and thus static unless ArcDps changes the format. It has been ``1`` for 2 years now.

Blish HUD/GameServices/ArcDps/Models/Ev.cs
agaertner: This is the only thing left but no clue how to phrase it better. I do not find your suggestion particular convincing either, because the phrase from the official doc states "buff damage" specifically.

dlamkins (approved):

PR #333, opened 2020-10-27
GameIntegration.IsInGameChanged event

agaertner:


Blish HUD/GameServices/Gw2MumbleService.cs
```diff
@@ -50,6 +50,21 @@ public class Gw2MumbleService : GameService {
...
         
+        #region Events
+        /// <summary>
+        /// Fires when the mumble availability changes.
+        /// </summary>
+        public event EventHandler<ValueEventArgs<bool>> IsAvailableChanged;
+
+        private void OnIsAvailableChanged(ValueEventArgs<bool> e) => IsAvailableChanged?.Invoke(this, e);
+
+        private bool   _prevIsAvailable               = false;
```
dlamkins: Get rid of the extra spacing here and put a newline before the `#endregion` and after the `#region Events` portions and we should be good.  👍

Blish HUD/GameServices/Gw2MumbleService.cs
```diff
@@ -73,6 +88,8 @@ public class Gw2MumbleService : GameService {
         }
 
         protected override void Update(GameTime gameTime) {
+            HandleEvents();
```
dlamkins: Move this to after the other update calls in this method are made.  That way when the event is called, the fresh values are already available.

dlamkins (changes requested):

greaka (approved):

agaertner:
Doesn't seem to fire yet. Any clue? The implementation is equal to the others.
May be just my module. But it's like two lines and breakpoints aren't hit.

dlamkins:
The Mumble Link API is technically available outside of when it is being updated. I believe `IsAvailable` is only checking to see if the memory mapped file exists to be read which could even be after the game has closed. 

Perhaps the properties on the integration service about being in game would do better for your needs since they check the time since the last tick, I believe?

agaertner:
You mean the ``IsInGame``?

Oh what. I am actually just noticing IsInGame uses Gw2Mumble.
Why is it in GameIntegration?

agaertner:
ok I've removed ``Gw2Mumble.IsAvailableChanged``
in favor of ``GameIntegration.IsInGameChanged``

works well.

Not sure though why this field isn't under mumble as 
``Gw2Mumble.IsUpdating`` or ``Gw2Mumble.IsActive``

GameIntegration could keep the ``IsInGame`` field as an expression: 

```csharp
GameIntegration.IsInGame => Gw2Mumble.IsUpdating
```

dlamkins:
It's under the `GameIntegrationService` because it has roots in other features that were added early on.

dlamkins (approved):

Issue #69, opened 2019-07-21
Global volume settings


BlishHUD:
**majσR at 1/25/2019 2:01:28 AM +00:00**
> Small feature request: Volume control from within Blish HUD. Windows volume mixer works but it would be much easier in app.

REF: https://discordapp.com/channels/531175899588984842/534490472224391169/538176549241749536

dlamkins:
Sound loading and playing, in general, needs an overhaul. I think we can use the same library Andy did for the musician module as it definitely appeared to be flexible whereas the MonoGame audio support is fairly limited. 

First steps would be wrapping that library with our `ContentService` and 
`ContentsManager` to allow loading audio from the ref.dat or the module ref. 

I think when playing audio, it would be best to provide a "sound channel" for each volume slider available in GW2. For now, the volume of each channel would probably be the same and tied to a single volume slider, but it allows us to be flexible in the future if options for better set the volume to match the game settings become available.

dlamkins:
While this could still be nice to have, the importance is now much more diminished with the merging of #287 now allowing for the master audio level to be managed automatically.

dlamkins:
Closed by #335

PR #335, opened 2020-11-04
Extend AudioIntegration

Denrage:
This fixes mainly #69 but more changes were made to make Blish HUD more immersive:
* The user can now choose if Blish HUD should adjust to GW2 Volume or set his own Volume
* Blish HUD will now change it's Audio Device when GW2 changes it which the user can turn on and off
* SoundEffects in general now uses CSCore for Audio Output and SoundEffects are no longer in the ContentPipeline


Blish HUD/GameServices/ContentService.cs
```diff
@@ -101,21 +107,39 @@ public enum FontStyle {
...
 
         public Microsoft.Xna.Framework.Content.ContentManager ContentManager => Blish_HUD.BlishHud.ActiveContentManager;
-        
+
         protected override void Initialize() { /* NOOP */ }
 
         protected override void Load() {
             Textures.Load();
+            var zipArchiveReader = new ZipArchiveReader(RefPath);
+            _audioDataReader = zipArchiveReader.GetSubPath("audio");
```
dlamkins, line 116: What was the reason for limiting it all down to the audio subpath and replacing the audio paths in all of the other places instead of it all still being relative to root?  Just figured it best to lock it down to all be under /audio?

I'm not against it, just wanted to know if there was anything else behind the decision.  👍

Blish HUD/GameServices/GameIntegration/AudioIntegration.cs
```diff
@@ -1,38 +1,66 @@
...
         /// <summary>
-        /// Provides an estimated volume level for the application
-        /// based on the volumes levels exhibited by the game.
+        /// This either provides  an estimated volume level for the application
+        /// based on the volumes levels exhibited by the game
+        /// or
+        /// the set volume in settings.
         /// </summary>
-        public float AverageGameVolume => _averageGameVolume ??= CalculateAverageVolume();
+        public float Volume => _volume ??= GetVolume();
```
dlamkins, line 50: Noting a breaking change here.

Blish HUD/GameServices/GameIntegration/AudioIntegration.cs
```diff
@@ -1,38 +1,66 @@
...
+        /// or the selected one in the settings.
+        /// </summary>
+        public MMDevice AudioDevice { get; private set; }
 
         public AudioIntegration(GameIntegrationService service) : base(service) {
+            var audioSettings = GameService.Settings.RegisterRootSettingCollection(APPLICATION_SETTINGS);
+            _useGameVolume = audioSettings.DefineSetting(USEGAMEVOLUME_SETTINGS, true, Strings.GameServices.OverlayService.Setting_UseGameVolume_DisplayName, Strings.GameServices.OverlayService.Setting_UseGameVolume_Description);
+            _volumeSetting = audioSettings.DefineSetting(VOLUME_SETTINGS, MAX_VOLUME / 2, Strings.GameServices.OverlayService.Setting_Volume_DisplayName, Strings.GameServices.OverlayService.Setting_Volume_Description);
+            _volumeSetting.SetRange(0.0f, MAX_VOLUME);
+            _deviceSetting = audioSettings.DefineSetting(DEVICE_SETTINGS, Devices.Gw2OutputDevice, Strings.GameServices.OverlayService.Setting_AudioDevice_DisplayName, Strings.GameServices.OverlayService.Setting_AudioDevice_Description);
```
dlamkins, line 63: Ownership for some for these changes may not be appropriate here.  It's hard to say.  We don't have a dedicated audio service, but this particular ServiceModule was specifically for providing the integration to pull from the game's audio levels with its estimates which is why it's owned by the `GameIntegrationService`.

It may be that we need to have an audio service or provide one as a ServiceModule under a rewritten `ContentService`.  🤔

In the meantime, it's probably fine to keep here until that work is done.

dlamkins:
@greaka If you have some time in the next couple of days, do you think you could give this a quick look over?

Blish HUD/GameServices/GameIntegration/AudioIntegration.cs
dlamkins: #336 for the primary continuation of this.

Blish HUD/GameServices/GameIntegration/AudioIntegration.cs
dlamkins: I don't believe any modules currently rely on this at the moment.  The first one that comes to mind relies on its own master volume multiplier.

https://github.com/OpNop/UndaDaSea-BlishHUD/blob/master/UndaDaSea_BlishHUD.cs

Blish HUD/GameServices/ContentService.cs
Denrage: Otherwise you always have a redundancy for the audio path and something that wants to access audio shouldn't care in which folder it resides in and let it handle from the called function. It makes more sense when the function is called "PlayAudioFromName" that you actually only need to put in a name and not a path. tldr: the caller of that function shouldn't care about a "filesystem" behind playing the audio.

Blish HUD/GameServices/GameIntegration/AudioIntegration.cs
Denrage: Hmmm, yeah, didn't think about it, normally things available to dependencies outside of the main applicaton have an interface. Thought the naming would be more appropriate now but it isn't smth that would break the code if the name would still be `AverageGameVolume`

Blish HUD/GameServices/GameIntegration/AudioIntegration.cs
Denrage: So the `AudioIntegration` purpose was mainly to get the ingame audio average volume and for nothing else? Would make sense bc before the "AudioService" from MonoGame handled the playing of the sound. Would also move out the whole audio playing in `ContentService` into that AudioService and move the whole thing i did out of the `AudioIntegration` into that service too tbh

Blish HUD/GameServices/ContentService.cs
```diff
@@ -101,21 +107,39 @@ public enum FontStyle {
...
         }
 
         public void PlaySoundEffectByName(string soundName) {
-            if (!_loadedSoundEffects.ContainsKey(soundName)) {
-                _loadedSoundEffects.TryAdd(soundName, Blish_HUD.BlishHud.ActiveContentManager.Load<SoundEffect>($"{soundName}"));
-            }
+            const string SOUND_EFFECT_FILE_EXTENSION = ".wav";
+            var filePath = soundName + SOUND_EFFECT_FILE_EXTENSION;
+            if (_audioDataReader.FileExists(filePath)) {
```
greaka, line 122: It might be worth to invert the if here and throw a debug level message to the logger.

Blish HUD/GameServices/ContentService.cs
```diff
@@ -101,21 +107,39 @@ public enum FontStyle {
...
+                soundOutput.Initialize(waveSource);
+                soundOutput.Volume = GameService.GameIntegration.Audio.Volume;
+
+                Task.Run(() => {
+                    soundOutput.Play();
+                    soundOutput.WaitForStopped();
+
+                    // This will dispose the sound effect stream and the wave source too
+                    soundOutput.Dispose();
+                });
```
greaka, line 136: Does `using var soundOutput` not work with `Task.Run` here? If it does, then I would opt for using the `using` keyword here.

Blish HUD/GameServices/GameIntegration/AudioIntegration.cs
greaka: Given that the meaning of it changed, I would go for `Volume` here too.

Blish HUD/GameServices/GameIntegration/AudioIntegration.cs
greaka: I agree that this doesn't belong here. However, I'm fine with having it here for the time being. I see that the follow-up issue already exists 🙂

greaka (commented):
Overall a very good contribution 🙂

Blish HUD/GameServices/ContentService.cs
Denrage: Because the object in `soundOutput` is instanciated in another Task than the Task.Run-Task, the method could return before that task reaches it's end. Therefore it could be disposed when the Task.Run wants to access the object. Probably the creation of the soundOutput should also be in that task so the ownership won't shift (and the `using`-Keyword can be used)

Blish HUD/GameServices/GameIntegration/AudioIntegration.cs
dlamkins: Yes, that's fine.  Was just noting that it's a breaking change.

Blish HUD/GameServices/ContentService.cs
dlamkins: I agree, but we can keep that for #336 since these changes here are mostly for compatibility with the changes made elsewhere.

Blish HUD/GameServices/ContentService.cs
dlamkins: I believe everything has been addressed elsewhere.  Were there any other changes needed here?

greaka (approved):

dlamkins (approved):

PR #338, opened 2020-11-20
:tada: TyrianTimeUtil 

agaertner:
:tada: TyrianTimeChanged
** thanks to Thaumanova for the improved conversion function


dlamkins:
It's a nice to have, but I don't see enough use for it for us to include it at this time in the core overlay.  I'd recommend just putting it in the modules that you need it in.

PR #342, opened 2020-12-01
Module repo support

dlamkins:
Adds module repo support.

- IPkgRepoProvider allows for alternative module repo sources.
- Repo URL is stored in settings and points to the GitHub repository acting as the repo (which GitHubPkgRepoProvider can be used to load).
- Modules can be set to any version exposed in the repo allowing users to install, upgrade, and downgrade their modules entirely from within the UI of the overlay.
- Listed modules update dynamically under settings.

Changes needed before PR is finalized:
- Resolve logic issue when a module is set to multiple versions back and forth in the same session (installed version does not match listed version).
  - ANSWER: Modules are fully deleted from the modules folder instead of relying on it happening at next launch.
- Notify user if the assembly has already been loaded and thus a restart of the overlay will be required.
  - ANSWER: A warning is shown both on the package repo page as well as the individual module's page as well.
- Finalize UI details.


dlamkins:
Under consideration with this PR:

Splitting module state management and repo stuff into their own ServiceModules to reduce the complexity of the ModuleService.

Blish HUD/GameServices/Modules/Pkgs/GitHubPkgRepoProvider.cs
```diff
@@ -0,0 +1,127 @@
...
+        public GitHubPkgRepoProvider(string repoUrl) {
+            this.RepoUrl = repoUrl;
+        }
+
+        public async Task<bool> Load(IProgress<string> progress) {
+            if (!_pkgCache.ContainsKey(this.RepoUrl)) {
+                _pkgCache[this.RepoUrl] = await LoadRepo(progress);
+            }
+
+            return _pkgCache != null;
```
greaka: `_pkgCache` can't be null

Blish HUD/GameServices/Modules/Pkgs/GitHubPkgRepoProvider.cs
```diff
@@ -0,0 +1,127 @@
...
+                if (manifests.Exception == null) {
+                    return manifests.PkgManifests;
+                } else {
+                    progress?.Report($"{Strings.GameServices.ModulesService.PkgManagement_Progress_FailedToReadOrParseRepoManifest}\r\n{manifests.Exception.Message}");
+                }
+            } else {
+                progress?.Report($"{Strings.GameServices.ModulesService.PkgManagement_Progress_FailedToGetReleases}\r\n{releaseResult.Exception.Message}");
+            }
+
+            return null;
```
greaka: If you want, then you could flatten this construct by inverting all the ifs. IMO this would help readability.
e.g.
```cs
if (releaseResult.Exception != null) {
    progress?.Report("");
    return null;
}
...
if (manifests.Exception != null) {
    progress?.Report("");
    return null;
}

return manifests.PkgManifests;
```

Blish HUD/GameServices/Modules/Pkgs/GitHubPkgRepoProvider.cs
```diff
@@ -0,0 +1,127 @@
...
+            return (null, lastException);
+        }
+
+        public IEnumerable<PkgManifest> GetPkgManifests(params Func<PkgManifest, bool>[] filters) {
+            foreach (var pkg in from pkg in _pkgCache[this.RepoUrl]
+                                let include = filters.All(filter => filter(pkg))
+                                where include
+                                select pkg) {
+                yield return pkg;
+            }
```
greaka: IMO, the query syntax is harder to read than the function syntax, even though the function syntax just feels like boilerplate when writing the code.
The whole function could be rewritten as:
```cs
        public IEnumerable<PkgManifest> GetPkgManifests(params Func<PkgManifest, bool>[] filters) {
            return _pkgCache[this.RepoUrl].Where(pkg => filters.All(filter => filter(pkg)));
        }
```
I'm pretty sure you could even change the `IEnumerable` to `IQueryable` without any problem

dlamkins: I didn't actually mean to leave this query syntax in.  I find it quite more difficult to read, too.  ReSharper must have been getting antsy.  😁

Blish HUD/GameServices/Modules/Pkgs/GitHubPkgRepoProvider.cs
dlamkins: Having the same 
```cs
    progress?.Report("");
    return null;
```
written twice seems to really bother me.  I'll think it over / see how it looks.

Blish HUD/GameServices/Modules/Pkgs/GitHubPkgRepoProvider.cs
dlamkins: Good catch.  👍

dlamkins:
The most recent commits have this PR feature ready and it is now just pending some testing and review.

Blish HUD/GameServices/ModuleService.cs
```diff
@@ -128,7 +137,28 @@ public class ModuleService : GameService {
             return RegisterModule(new ZipArchiveReader(modulePath));
         }
 
-        private ModuleManager LoadModuleFromUnpackedBhm(string moduleDir) {
+        /// <summary>
+        /// Unregisters the module.
+        /// </summary>
+        /// <returns>Returns if the assembly was ever loaded or not (thus indicating if it has been fully unloaded or not).</returns>
+        public void UnregisterModule(ModuleManager moduleManager) {
```
greaka: Is the returns wanted here? The method returns void.

Blish HUD/GameServices/Modules/Pkgs/GitHubPkgRepoProvider.cs
```diff
@@ -0,0 +1,153 @@
...
+            return _pkgCache[this.RepoUrl].Where(pkg => _activeFilters.All(filter => filter(pkg)));
+        }
+
+        public IEnumerable<(string OptionName, Action<bool> OptionAction, bool IsToggle)> GetExtraOptions() {
+            yield return ("Force Reload Repository", async (toggleState) => { await LoadRepo(); }, false);
+
+            yield return ("Only Show Supported Versions", (toggleState) => ToggleFilter(FilterShowOnlySupportedVersion, toggleState), true);
+            yield return ("Only Show Modules With Updates", (toggleState) => ToggleFilter(FilterShowOnlyUpdates,        toggleState), true);
+            yield return ("Only Show Installed Modules", (toggleState) => ToggleFilter(FilterShowOnlyInstalled,         toggleState), true);
+            yield return ("Only Show Modules Not Installed", (toggleState) => ToggleFilter(FilterShowOnlyNotInstalled,  toggleState), true);
```
greaka: Not a blocker, but in the future this should support localization

Blish HUD/GameServices/Modules/UI/Views/ManagePkgView.cs
```diff
@@ -0,0 +1,220 @@
...
+                case PkgVersionRelationship.CanUpdate:
+                    _statusImage.Texture          = GameService.Content.GetTexture("common/157397");
+                    _statusImage.BasicTooltipText = "A newer version of this module is available.";
+                    _statusImage.Visible          = true;
+                    break;
+                case PkgVersionRelationship.CurrentVersion:
+                    _statusImage.Texture          = GameService.Content.GetTexture("common/157330");
+                    _statusImage.BasicTooltipText = "This module is installed and up to date.";
+                    _statusImage.Visible          = true;
+                    break;
```
greaka: localization support is missing here too

Blish HUD/GameServices/Modules/UI/Views/ManagePkgView.cs
```diff
@@ -0,0 +1,220 @@
...
+                Top     = 5,
+                Parent  = buildPanel
+            };
+
+            _statusImage = new Image(GameService.Content.GetTexture("common/157397")) {
+                Visible          = false,
+                Size             = new Point(16, 16),
+                Top              = _versionDropdown.Height / 2 - 8 + _versionDropdown.Top,
+                Right            = _versionDropdown.Left           - 8,
+                BasicTooltipText = "A newer version of this module is available.",
```
greaka: localization support is missing here too

greaka (approved):
The comments are not blockers. Overall solid PR, I didn't test it yet though.

For the future we might not want to forget to support multiple pkg repo urls, but that involves a greater change including UI and doesn't need to be part of this PR.

Blish HUD/GameServices/Modules/UI/Views/ManagePkgView.cs
dlamkins: 256177e Thanks!

Blish HUD/GameServices/Modules/UI/Views/ManagePkgView.cs
dlamkins: 256177e Thanks!

Blish HUD/GameServices/Modules/Pkgs/GitHubPkgRepoProvider.cs
dlamkins: e1ee816 Thanks!

Blish HUD/GameServices/ModuleService.cs
dlamkins: Whoops!  
36bd1fe

dlamkins:
> For the future we might not want to forget to support multiple pkg repo urls, but that involves a greater change including UI and doesn't need to be part of this PR.

Agreed!  For now, as a stop gap, I've used bhud-pkgs as the default value which is actually added to the settings.json.  It can be overridden by manually editing the settings.json file, for now.  :)

We'll discuss it more in the future, but I tried to abstract things enough so that we can have other types of repo providers as necessary (or modules could even technically add support for other repos, I suppose).

greaka (approved):

PR #343, opened 2020-12-01
:sparkles: ArcDPS enums

agaertner:
I was going todo the entirety myself, then I noticed Elite Insights had already copied them over nicely.
I went ahead and asked them to use the file. I was kindly notified by the MIT License in their repo. Source is in the docs.

Wired up and ready.
Might want to exclude that file from codefactor if we were to just keep it up to date with theirs. It's a nice resource and they are a group working on finding the ids and stuff


dlamkins:
> Does Gw2-Elite-Insights Parser have a NuGet package or anything that it deploys that would expose this same stuff to us? That would be preferred over manually copying over the enums and whatnot, but not entirely sure what all they have set up or if it's just a UI application with the API stuff not split out.

greaka (approved):
@dlamkins can you have a look too please?

dlamkins (approved):

dlamkins:
Looks good to me.  Thanks for revising the formating!

PR #370, opened 2021-05-30
Control improvements

dlamkins:
Containers should get rid of their child controls if explicitly disposed of.

Resolved CMS memory leak by centralizing the code used to close them when they're clicked off of.

Also allowed the menus to re-orient themselves when child items are dynamically added or removed from the menu while it is open.


greaka (approved):
Tough to spot, good job!

PR #373, opened 2021-06-07
Entity changes for v0.8.0

dlamkins:
All new IEntity and render path.  Renamed EntityEffect to SharedEffect (to replace base for ControlEffect).  Made working with cameras easier by passing them through the render path.


greaka (approved):
I really like where this is going!

PR #389, opened 2021-07-15
add logos for access marking

agaertner:
A lof of things require specific access. A ton of achievements such as dailies, crafting recipes, legendary equipment etc
I prepared these 512px images to help mark access requirements.


dlamkins:
Keeping accessory textures to modules for now unless they are also utilized by core.

PR #391, opened 2021-07-18
Hack to fix scroll issues in settings

manlaan:


dlamkins:
While I dislike how slow the larger panels now scroll, I think this certainly fixes more issues than it causes.  Thank you for adjusting the view as well to make sure that the scrollbar wasn't covered up by the parent container.

We'll track the scrollbar speed considerations in #392.

dlamkins (approved):

Issue #399, opened 2021-07-21
Setting display names and descriptions are never recalculated


dlamkins:
When a setting is defined, so too is its DisplayName and Description which are used when rendering the setting in the UI.  This causes issues with localization as it assumes that views will have a chance to reference the localized strings every time the entry is built.

When defining settings, the option to define using a Func<string> delegate will allow the displayname and description to update themselves when the locale does.

![image](https://user-images.githubusercontent.com/1950594/126421920-0ab543fd-2015-42af-9852-bfec09a004c1.png)

PR #400, opened 2021-07-21
Allow setting displaynames and descriptions to update when language is changed

dlamkins:
Introduced `func<string>` based delegates for `SettingEntry` display name and description.  These can be called by UI when views are built to get the proper string translation.

Also removes most of the old setting renderer methods.  Obsolete delegate left only to keep backwards compatibility with modules still calling the old `DefineSetting` function.  Technically a breaking change, but nothing should be left using the items which were removed.


dlamkins:
Resolves #399

nickcode122 (approved):

PR #401, opened 2021-07-21
WIP: Self Update Support

dlamkins:
Introduces support for a self updater:
- About page shows the option to select a version (this means that users can downgrade their version if they would like).
- Available versions are pulled from `https://versions.blishhud.com/all.json`.
- Option to open changelog and to toggle showing prerelease versions (WIP).

Self update process works as follows:
- When a version is selected, Blish HUD unloads the module (so that modules can cleanly unload) and input service (so that the hooks are disabled and the debug hook application closed, if in a debug session).
- Blish HUD window is closed and made unavailable.  A loading spinner is shown in the top left for the duration of the package download.
- The package is downloaded to "unpack.zip" in the application directory.
- Once downloaded, the checksum is compared to what was listed in the original manifest.
  - If checksum is invalid then the user is told the update failed and Blish HUD restarts automatically.
- Blish HUD moves its own exe to rename it in the same folder to `Blish HUD.exe__bak`.
- The new `Blish HUD.exe` is extracted from `unpack.zip` and placed in the folder.
- Blish HUD restarts passing the `--parentpid <current pid>` to the new Blish HUD executable.

When Blish HUD launches:
- If `unpack.zip` is found in the same directory, it is assumed that we are mid-upgrade.
- We delete the `Blish HUD.exe__bak`.
- We wait a max of 10 seconds for the previous Blish HUD instance to finish exiting (we do not force close it, we want it to exit cleanly) by checking for parentpid to stop existing.
  - If after 10 seconds the parentpid is still running, we will notify the user and the application exits.
  - Any future attempts to launch Blish HUD will attempt to complete the update so it is able to usually recover.
- We remove the old content, language dlls, hook helper exe, ref.dat, etc.
- We begin exporting the files found in `unpack.zip` to the directory.
- We delete `unpack.zip`.
- Blish HUD continues launching as it normally would from this point forward.

Remaining items:
- [ ] Finalize settings context menu with "Open changelogs" and "Toggle prerequisites" support.
- [ ] Enable filtering of prerequisites in version dropdown.
- [ ] Show update arrow on the settings tab if an update is available.

### DIFF
--- a/Blish HUD/ApplicationSettings.cs
+++ b/Blish HUD/ApplicationSettings.cs
@@ -23,7 +23,7 @@ namespace Blish_HUD {
         }
 
         public override void OnUserFacingException(UserFacingException e, string message) {
-            MessageBox.Show("Invalid launch option(s) specified.  See --help for available options.", "Failed to launch Blish HUD", MessageBoxButtons.OK);
+            MessageBox.Show($"Invalid launch option(s) specified.  See --help for available options.\r\n\r\n{e.Message}", "Failed to launch Blish HUD", MessageBoxButtons.OK);
         }
 
         public override void OnHelpInvoked(string helpText) {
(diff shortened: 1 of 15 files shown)


dlamkins:
With the direction we're taking in .NET, the embedding of the debug helper, and more, much of this work may need to be redone to account for the changes, so for now I'm going to go ahead and close and we can review again in the future.

PR #402, opened 2021-07-27
Lock container children

dlamkins:
Introduces a new `ControlCollection` list which handles ownership of controls in a thread-safe way.  Written with the expectation that most actions will be read instead of write.

This fixes a ton of different edge cases which can be caused by quick UI changes.  For example: an edge case where children are cleared mid-enumeration (e.g. control is mid-disposal) and thus an exception is thrown.  **Lock prevents enumerations and clears from occurring at the same time.**

This has been done in a fairly non-breaking way, so in my testing none of the modules had any immediate issues with it.  That being said, anything that had custom containers may have issue if they manipulated the children in a more direct way, thus it is marked a breaking change.

Enumeration is now safe from being modified mid-enumeration and two new functions allow to get safe non-locking lists/arrays which can be used to avoid dead-locking the application (e.g. avoiding locking when a child control decides to change its parent as part of its update in the update loop).

Also includes a minor fix to `Control.AbsoluteBounds` which more than doubles UI performance.  I should have done this a long time ago.


entrhopi:
As stated in #405 an `else` should be added to `Control.cs` like this:
```
        public Rectangle AbsoluteBounds {
            get {
                if (_parent == null) return this.LocalBounds;
                else {
                    var parentBounds        = _parent.AbsoluteBounds;
                    var parentContentRegion = _parent.ContentRegion;

                    // Clean this up
                    // This is really the absolute bounds of the ContentRegion currently because mouse
                    // input is currently using this to determine if the click was within the region
                    return new Rectangle(parentBounds.X + parentContentRegion.X + _location.X - _parent.HorizontalScrollOffset,
                                         parentBounds.Y + parentContentRegion.Y + _location.Y - _parent.VerticalScrollOffset,
                                         _size.X,
                                         _size.Y);
                }
            }
        }
```

greaka (changes requested):
See Discord :slightly_smiling_face:

dlamkins:
> See Discord 🙂

_Maybe_ all set now.

Issue #5, opened 2019-05-12
Mult-lined Textbox Control


greaka:
Missing features:

- [x] Basic writing
- [x] Control keys
- [x] Shift key

TybaIt:
This should be very high up in priority.

dlamkins:
Closed in #404!

PR #404, opened 2021-07-29
Support for modifiers in KeyboardHandler

entrhopi:
AFAIK it has only 2 minor issues atm:
- number keys do not respect keyboard layout (it always uses US layout)
- holding shift for too long will kind of lock it for the next key (does not impact normal writing)

### DIFF
--- a/Blish HUD/Controls/TextInputBase.cs
+++ b/Blish HUD/Controls/TextInputBase.cs
@@ -43,7 +43,7 @@ namespace Blish_HUD.Controls {
 
         private static readonly Logger Logger = Logger.GetLogger<TextInputBase>();
 
-        protected static readonly char[] WordSeperators  = { ' ', '\n', '`', '~', '!', '@', '#', '%', '^', '&', '*', '(', ')', '-', '=', '+', '[', '{', ']', '}', '\\', '|', ';', ':', '\'', '"', ',', '.', '<', '>', '/', '?' };
+        protected static readonly char[] WordSeperators = { ' ', '\n', '`', '~', '!', '@', '#', '%', '^', '&', '*', '(', ')', '-', '=', '+', '[', '{', ']', '}', '\\', '|', ';', ':', '\'', '"', ',', '.', '<', '>', '/', '?' };
 
         private static readonly Color _highlightColor = new Color(92, 80, 103, 150);
 
@@ -209,34 +209,8 @@ namespace Blish_HUD.Controls {
         }
 
         private void OnTextInput(string value) {
-            bool ctrlDown = this.IsCtrlDown;
-
             foreach (char c in value) {
-                if (char.IsControl(c)) continue;
                 if (_font.GetCharacterRegion(c) == null) continue;
-
-                if (ctrlDown) {
-                    switch (c) {
-                        case 'c':
-                            HandleCopy();
-                            return;
-                        case 'x':
-                            HandleCut();
-                            return;
-                        case 'v':
-                            HandlePaste();
-                            return;
-                        case 'z':
-                            HandleUndo();
-                            return;
-                        case 'y':
-                            HandleRedo();
-                            return;
-                        case 'a':
-                            SelectAll();
-                            return;
-                    }
-                }
 
                 InputChar(c);
             }
@@ -553,6 +527,24 @@ namespace Blish_HUD.Controls {
                 case Keys.Enter:
                     HandleEnter();
                     break;
+                case Keys.C:
+                    if (this.IsCtrlDown) HandleCopy();
+                    return;
+                case Keys.X:
+                    if (this.IsCtrlDown) HandleCut();
+                    return;
+                case Keys.V:
+                    if (this.IsCtrlDown) HandlePaste();
+                    return;
+                case Keys.Z:
+                    if (this.IsCtrlDown) HandleUndo();
+                    return;
+                case Keys.Y:
+                    if (this.IsCtrlDown) HandleRedo();
+                    return;
+                case Keys.A:
+                    if (this.IsCtrlDown) SelectAll();
+                    return;
                 default:
                     // Skip key repeat state
                     return;
@@ -568,10 +560,10 @@ namespace Blish_HUD.Controls {
 
                 ClipboardUtil.WindowsClipboardService.SetTextAsync(clipboardText)
                              .ContinueWith((clipboardResult) => {
-                                  if (clipboardResult.IsFaulted) {
-                                      Logger.Warn(clipboardResult.Exception, "Failed to set clipboard text to {clipboardText}!", clipboardText);
-                                  }
-                              });
+                                 if (clipboardResult.IsFaulted) {
+                                     Logger.Warn(clipboardResult.Exception, "Failed to set clipboard text to {clipboardText}!", clipboardText);
+                                 }
+                             });
             }
         }
 
@@ -583,14 +575,14 @@ namespace Blish_HUD.Controls {
         protected virtual void HandlePaste() {
             ClipboardUtil.WindowsClipboardService.GetTextAsync()
                          .ContinueWith((clipboardTask) => {
-                              if (!clipboardTask.IsFaulted) {
-                                  if (!string.IsNullOrEmpty(clipboardTask.Result)) {
-                                      Paste(clipboardTask.Result);
-                                  }
-                              } else {
+                             if (!clipboardTask.IsFaulted) {
+                                 if (!string.IsNullOrEmpty(clipboardTask.Result)) {
+                                     Paste(clipboardTask.Result);
+                                 }
+                             } else {
                                  Logger.Warn(clipboardTask.Exception, "Failed to read clipboard text from system clipboard!");
-                              }
-                          });
+                             }
+                         });
         }
 
         protected virtual void HandleUndo() {
--- a/Blish HUD/GameServices/Input/Keyboard/KeyboardHandler.cs
+++ b/Blish HUD/GameServices/Input/Keyboard/KeyboardHandler.cs
@@ -61,8 +61,8 @@ namespace Blish_HUD.Input {
             Keys.LeftWindows,  // TODO: let the OS handle all key presses that are made during the Windows-key is pressed
             Keys.RightWindows, //       i.e. pressing win+space does not result in keyboard layout changes but just opens the start menu
             Keys.CapsLock,
-            //Keys.LeftControl,  // Hack to get CTRL+C/X/V/Y/A working again
-            //Keys.RightControl, //
+            Keys.LeftControl,
+            Keys.RightControl,
             Keys.LeftShift,
             Keys.RightShift
         };
--- a/Blish HUD/GameServices/Input/Keyboard/KeyboardHandler.cs
+++ b/Blish HUD/GameServices/Input/Keyboard/KeyboardHandler.cs
@@ -126,7 +126,7 @@ namespace Blish_HUD.Input {
 
         private bool ShouldBlockKeyEvent(Keys key) {
             // Prevent blocking shift for input capitalization
-            // if (key == Keys.LeftShift || key == Keys.RightShift) return false; // "SHIFT" support temporarily disabled
+            if (_keysDown.Contains(Keys.LeftShift) || _keysDown.Contains(Keys.RightShift)) return false; // "SHIFT" support temporarily disabled
 
             // Skip keys that we wish to explicitly ignore
             if (_hookIgnoredKeys.Contains(key)) return false;
@@ -141,7 +141,7 @@ namespace Blish_HUD.Input {
             _inputBuffer.Enqueue(new KeyboardEventArgs(eventType, key));
 
             if (_textInputDelegate != null) {
-                string chars = TypedInputUtil.VkCodeToString((uint)key, eventType == KeyboardEventType.KeyDown);
+                string chars = TypedInputUtil.VkCodeToString((uint)key, eventType == KeyboardEventType.KeyDown, _keysDown.Contains(Keys.LeftShift) || _keysDown.Contains(Keys.RightShift));
                 _textInputDelegate?.BeginInvoke(chars, EndTextInputAsyncInvoke, null);
                 return ShouldBlockKeyEvent(key);
             }
--- a/Blish HUD/GameServices/Input/Keyboard/KeyboardHandler.cs
+++ b/Blish HUD/GameServices/Input/Keyboard/KeyboardHandler.cs
@@ -126,7 +126,7 @@ namespace Blish_HUD.Input {
 
         private bool ShouldBlockKeyEvent(Keys key) {
             // Prevent blocking shift for input capitalization
-            //if (_keysDown.Contains(Keys.LeftShift) || _keysDown.Contains(Keys.RightShift)) return false; // "SHIFT" support temporarily disabled
+            if (_keysDown.Contains(Keys.LeftShift) || _keysDown.Contains(Keys.RightShift)) return false; // "SHIFT" support temporarily disabled
 
             // Skip keys that we wish to explicitly ignore
             if (_hookIgnoredKeys.Contains(key)) return false;
@@ -141,7 +141,7 @@ namespace Blish_HUD.Input {
             _inputBuffer.Enqueue(new KeyboardEventArgs(eventType, key));
 
             if (_textInputDelegate != null) {
-                string chars = TypedInputUtil.VkCodeToString((uint)key, eventType == KeyboardEventType.KeyDown, _keysDown.Contains(Keys.LeftShift) || _keysDown.Contains(Keys.RightShift));
+                string chars = TypedInputUtil.VirtKeyCodeToString((uint)key, eventType == KeyboardEventType.KeyDown);
                 _textInputDelegate?.BeginInvoke(chars, EndTextInputAsyncInvoke, null);
                 return ShouldBlockKeyEvent(key);
             }
--- a/Blish HUD/GameServices/Input/Keyboard/KeyboardHandler.cs
+++ b/Blish HUD/GameServices/Input/Keyboard/KeyboardHandler.cs
@@ -58,8 +58,8 @@ namespace Blish_HUD.Input {
             Keys.LeftAlt,
             Keys.RightAlt,
             Keys.NumLock,
-            Keys.LeftWindows,
-            Keys.RightWindows,
+            Keys.LeftWindows,  // TODO: let the OS handle all key presses that are made during the Windows-key is pressed
+            Keys.RightWindows, //       i.e. pressing win+space does not result in keyboard layout changes but just opens the start menu
             Keys.CapsLock
         };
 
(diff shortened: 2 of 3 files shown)


Blish HUD/GameServices/Input/Keyboard/TypedInputUtil.cs
```diff
@@ -110,6 +110,9 @@ internal static class TypedInputUtil {
                 // Single character in buffer
                 case 1:
                     ret = sbString[0].ToString();
+                    if (shift) {
```
dlamkins: Can we mark this with a `//TODO: ` or something to indicate that we need to review in the future?  I'll create an issue to track it since certainly we'll want to aim for better keyboard layout support in the future, but I think this could be a fine solution for now.  Better than no support.

dlamkins (changes requested):

greaka:
Just leaving this here for the future: https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-tounicode

Blish HUD/GameServices/Input/Keyboard/TypedInputUtil.cs
entrhopi: I've committed a patch handling all the issues I've found so far.
But I'll mark potential places in the future.

entrhopi:
> Just leaving this here for the future: https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-tounicode

Isn't the ToUnicodeEx we're using here 'better', because it handles keyboard layouts?
My problem was that the shift-modifier wasn't working and the `ToUpper` way of capitalizing chars was not.

We'll probably have to rework the entire keyboard handling anyway as I would consider my patch only a temporary workaround to get basic things working now.

agaertner:
> Just leaving this here for the future: https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-tounicode

Also this https://stackoverflow.com/questions/17386766/how-do-i-get-a-char-obtained-by-holding-shift

Glad I was able to help with this.

I know TacO uses the ToASCIIEx function but I just saw the ToUnicodeEx function also takes an array where you could pass the shift key pressed state in.

Blish HUD/GameServices/Input/Keyboard/TypedInputUtil.cs
```diff
@@ -151,5 +170,28 @@ internal static class TypedInputUtil {
...
+            // vkKeyScanResult & 0xff is the base key, without any modifiers
+            uint code = (uint)vkKeyScanResult & 0xff;
+
+            // set shift key pressed
+            byte[] b = new byte[256];
+            b[0x10] = 0x80;
+
+            uint r;
+            // return value of 1 expected (1 character copied to r)
+            if (1 != ToAscii(code, code, b, out r, 0))
```
agaertner: Worth trying ToUnicode to reflect the old implementation.

entrhopi: I would like to revisit the whole keyboard handling in the near future but I think it's more important to have a quick solution for now that is working since the shift key is't working at all and somehow important for the functionality of at least the KillProof module.

agaertner: I was trying out your fork and it appears sometimes the bug that the shift toggle becomes active at the second or third letter still happens. There seems to be still something off with the state.

entrhopi: I think I've found what's causing that. I'm currently reworking the whole function to be more rubust and use `ToUnicodeEx`.
Please consider my PR as on-hold until the next commit.

entrhopi: I just noticed that accent and diacritic characters aren't working in either implementation so I'll fix that too.

entrhopi: So I've made some modifications but as @greaka pointed out setting the key state for shift manually is not the most elegant solution and I agree.
However, the problem seems that using `GetKeyboardState` and passing it directly to `ToUnicode` isn't working because it seems to set the low-order bit for the keys which is for 'toggle keys' whereas `ToUnicode` expects the high-oder bit set.
Here is what I've tried:
```cs
byte[] keyState = new byte[256];
GetKeyboardState(keyState);
ToUnicode(virtKeyCode, scanCode, keyState, output, (int)5, (uint)0);
```
Even getting some key states before calling `GetKeyboardState` as mentioned here ( https://stackoverflow.com/a/53713024 ) does not work (for me at least).

entrhopi: New insights:
Adding `Keys.(Left|Right)Shift` and `Keys.(Left|Right)Control` to `_hookIgnoredKeys` is necessary for the systems `User32.dll` to know of the states of these keys.
`GetKeyboardState` will then work if a `GetKeyState` is called before as stated in the SO post above.

This leaves us with all modified keys working.

Dead key implementation still pending.

entrhopi: Dead-key characters are now implemented too.

I think we're golden now.

agaertner: Tested. The bug that caps begins to apply to the second or third letter appears to be gone. 🚀

However, the current implementation seems to be blocking the key-combos in the [TextInputBase](https://github.com/blish-hud/Blish-HUD/blob/0e82ee8668b1e8fcb835795c4735e1ab360947e9/Blish%20HUD/Controls/TextInputBase.cs#L214-L243) or that control base has to be updated. Mainly I tried Ctrl+A (select all), Ctrl+C (copy), Ctrl+V (paste)

Blish HUD/GameServices/Input/Keyboard/TypedInputUtil.cs
```diff
@@ -14,142 +9,90 @@ internal static class TypedInputUtil {
...
-        private static extern uint GetCurrentThreadId();
-
-        private static uint   _lastVkCode   = 0;
-        private static uint   _lastScanCode = 0;
-        private static byte[] _lastKeyState = new byte[255];
-        private static bool   _lastIsDead   = false;
+        private static uint _lastVirtKeyCode = 0;
+        private static uint _lastScanCode = 0;
+        private static byte[] _lastKeyState = new byte[256];
+        private static bool _lastIsDead = false;
```
agaertner: redundant ``= false``

Blish HUD/GameServices/Input/Keyboard/TypedInputUtil.cs
```diff
@@ -14,142 +9,90 @@ internal static class TypedInputUtil {
...
         /// </summary>
-        /// <param name="vkCode">VKCode</param>
-        /// <param name="isKeyDown">Is the key down event?</param>
-        /// <returns>String representing single unicode character.</returns>
-        public static string VkCodeToString(uint vkCode, bool isKeyDown) {
+        /// <param name="vkCode"></param>
+        /// <param name="isKeyDown"></param>
+        /// <returns></returns>
+        public static string VirtKeyCodeToString(uint vkCode, bool isKeyDown) {
             // ToUnicodeEx needs StringBuilder, it populates that during execution.
```
agaertner: ``ToUnicode``

Blish HUD/GameServices/Input/Keyboard/TypedInputUtil.cs
```diff
@@ -14,142 +9,90 @@ internal static class TypedInputUtil {
         // because of this behavior, "^" is called dead key)
 
         [DllImport("user32.dll")]
-        private static extern int ToUnicodeEx(uint wVirtKey, uint wScanCode, byte[] lpKeyState, [Out, MarshalAs(UnmanagedType.LPWStr)] System.Text.StringBuilder pwszBuff, int cchBuff, uint wFlags, IntPtr dwhkl);
+        public static extern int ToUnicode(uint vkCode, uint scanCode, byte[] keyboardState, [Out, MarshalAs(UnmanagedType.LPWStr, SizeConst = 64)] System.Text.StringBuilder receivingBuffer, int bufferSize, uint flags);
```
agaertner: can be made private as all references appear to be in TypedInputUtil.cs.

Blish HUD/GameServices/Input/Keyboard/TypedInputUtil.cs
```diff
@@ -14,142 +9,90 @@ internal static class TypedInputUtil {
...
+        /// Convert virtual-key code and keyboard state to unicode string
         /// </summary>
-        /// <param name="vkCode">VKCode</param>
-        /// <param name="isKeyDown">Is the key down event?</param>
-        /// <returns>String representing single unicode character.</returns>
-        public static string VkCodeToString(uint vkCode, bool isKeyDown) {
+        /// <param name="vkCode"></param>
+        /// <param name="isKeyDown"></param>
+        /// <returns></returns>
+        public static string VirtKeyCodeToString(uint vkCode, bool isKeyDown) {
```
agaertner: internal scope might be more appropriate since we are dealing with keyboard state

Blish HUD/GameServices/Input/Keyboard/TypedInputUtil.cs
```diff
@@ -14,142 +9,90 @@ internal static class TypedInputUtil {
...
 
-        [DllImport("kernel32.dll")]
-        private static extern uint GetCurrentThreadId();
-
-        private static uint   _lastVkCode   = 0;
-        private static uint   _lastScanCode = 0;
-        private static byte[] _lastKeyState = new byte[255];
-        private static bool   _lastIsDead   = false;
+        private static uint _lastVirtKeyCode = 0;
+        private static uint _lastScanCode = 0;
```
agaertner: redundant ``= 0``

Blish HUD/GameServices/Input/Keyboard/TypedInputUtil.cs
```diff
@@ -14,142 +9,90 @@ internal static class TypedInputUtil {
...
-            // This programs Thread ID
-            uint thisProgramThreadId = GetCurrentThreadId();
+            byte[] keyState = new byte[256];
+            uint scanCode = MapVirtualKey(vkCode, 0);
 
-            // Attach to active thread so we can get that keyboard state
-            if (AttachThreadInput(thisProgramThreadId, currentWindowThreadId, true)) {
-                // Current state of the modifiers in keyboard
-                bKeyStateStatus = GetKeyboardState(bKeyState);
+            // aparrently GetKeyboardState() has a problem returning the correct states
```
agaertner: *apparently

agaertner (approved):
Tested. Everything appears to be working now. 👍 
Consider scopes and typos before merging.

entrhopi:
> Tested. Everything appears to be working now. 👍
> Consider scopes and typos before merging.

Thank you, will refactor.

Blish HUD/GameServices/Input/Keyboard/TypedInputUtil.cs
```diff
@@ -14,142 +9,89 @@ internal static class TypedInputUtil {
...
-            } else {
-                // Could not attach, perhaps it is this process?
-                bKeyStateStatus = GetKeyboardState(bKeyState);
-            }
-
-            // On failure we return empty string.
-            if (!bKeyStateStatus)
+        /// <param name="vkCode"></param>
+        /// <param name="isKeyDown"></param>
+        /// <returns></returns>
```
agaertner: empty doc tags can be removed.

Blish HUD/Controls/TextInputBase.cs
```diff
@@ -568,10 +560,10 @@ public abstract class TextInputBase : Control {
 
                 ClipboardUtil.WindowsClipboardService.SetTextAsync(clipboardText)
                              .ContinueWith((clipboardResult) => {
-                                  if (clipboardResult.IsFaulted) {
-                                      Logger.Warn(clipboardResult.Exception, "Failed to set clipboard text to {clipboardText}!", clipboardText);
-                                  }
-                              });
+                                 if (clipboardResult.IsFaulted) {
```
agaertner, line 563: It might be better to use
```csharp
(!clipboardResult.IsCompleted)
```
instead in a ContinueWith since IsCompleted covers 
Canceled, Faulted and RanToCompletion according to [docs](https://docs.microsoft.com/de-de/dotnet/api/system.threading.tasks.task.iscompleted?redirectedfrom=MSDN&view=net-5.0#System_Threading_Tasks_Task_IsCompleted).

Blish HUD/Controls/TextInputBase.cs
```diff
@@ -583,14 +575,14 @@ public abstract class TextInputBase : Control {
         protected virtual void HandlePaste() {
             ClipboardUtil.WindowsClipboardService.GetTextAsync()
                          .ContinueWith((clipboardTask) => {
-                              if (!clipboardTask.IsFaulted) {
-                                  if (!string.IsNullOrEmpty(clipboardTask.Result)) {
-                                      Paste(clipboardTask.Result);
-                                  }
-                              } else {
+                             if (!clipboardTask.IsFaulted) {
```
agaertner, line 578: It might be better to use
```csharp
(!clipboardResult.IsCompleted)
```
instead in a ContinueWith since IsCompleted covers 
Canceled, Faulted and RanToCompletion according to [docs](https://docs.microsoft.com/de-de/dotnet/api/system.threading.tasks.task.iscompleted?redirectedfrom=MSDN&view=net-5.0#System_Threading_Tasks_Task_IsCompleted).

Blish HUD/Controls/TextInputBase.cs
```diff
@@ -583,14 +575,14 @@ public abstract class TextInputBase : Control {
...
+                                 if (!string.IsNullOrEmpty(clipboardTask.Result)) {
+                                     Paste(clipboardTask.Result);
+                                 }
```
agaertner, line 581: redundant parenthesis

agaertner, line 581: try reduce redundant paranthesis in single-line nested IF-statements using indent instead. Might look better.

greaka (approved):
I am against the open proposed changes by Andy and would merge this as is after hearing back from the testers.

Blish HUD/Controls/TextInputBase.cs
dlamkins: This wasn't part of the original PR - it was just a formatting thing that got included.  We also only want to log an entry if it strictly faulted.  If it gets canceled or otherwise, no reason to log it because there won't be an exception included.

Blish HUD/Controls/TextInputBase.cs
dlamkins: Not really part of this PR - appears to be an artifact of some minor formatting that got applied to the whole file.  👍

Blish HUD/GameServices/Input/Keyboard/KeyboardHandler.cs
```diff
@@ -125,23 +130,19 @@ public class KeyboardHandler : IInputHandler {
...
-
             return true;
         }
 
         private bool ProcessInput(KeyboardEventType eventType, Keys key) {
             _inputBuffer.Enqueue(new KeyboardEventArgs(eventType, key));
 
             if (_textInputDelegate != null) {
-                string chars = TypedInputUtil.VkCodeToString((uint)key, eventType == KeyboardEventType.KeyDown);
+                string chars = TypedInputUtil.vkCodeToString((uint)key, eventType == KeyboardEventType.KeyDown);
```
dlamkins: Since this function is accessible outside of the class.  Let's keep the name as `VkCodeToString` (capital `V`).

Blish HUD/GameServices/Input/Keyboard/TypedInputUtil.cs
```diff
@@ -14,142 +9,86 @@ internal static class TypedInputUtil {
...
-        private static uint   _lastVkCode   = 0;
-        private static uint   _lastScanCode = 0;
-        private static byte[] _lastKeyState = new byte[255];
-        private static bool   _lastIsDead   = false;
+        private static uint _lastVirtKeyCode;
+        private static uint _lastScanCode;
+        private static byte[] _lastKeyState = new byte[256];
+        private static bool _lastIsDead;
+        private static byte VK_SHIFT = 0x10;
+        private static byte VK_MENU = 0x12;
```
dlamkins: These are walking and talking like consts.  Let's make 'em consts and place them near the top of the file (either before or after the DLLImports is fine).

Blish HUD/GameServices/Input/Keyboard/TypedInputUtil.cs
```diff
@@ -14,142 +9,86 @@ internal static class TypedInputUtil {
         // because of this behavior, "^" is called dead key)
 
         [DllImport("user32.dll")]
-        private static extern int ToUnicodeEx(uint wVirtKey, uint wScanCode, byte[] lpKeyState, [Out, MarshalAs(UnmanagedType.LPWStr)] System.Text.StringBuilder pwszBuff, int cchBuff, uint wFlags, IntPtr dwhkl);
+        private static extern int ToUnicode(uint vkCode, uint scanCode, byte[] keyState, [Out, MarshalAs(UnmanagedType.LPWStr, SizeConst = 64)] System.Text.StringBuilder receivingBuffer, int bufferSize, uint flags);
```
dlamkins, line 12: I might have missed it.  Was there any reason in particular we decided to go for `ToUnicode` instead of `ToUnicodeEx` for this?

dlamkins (changes requested):

dlamkins:
This is looking absolutely fantastic.  I posted my final thoughts prior to approval.

Assuming Ni has no issues when he tests it, I suspect we're nearly ready to merge!  🎉

Blish HUD/GameServices/Input/Keyboard/TypedInputUtil.cs
dlamkins: (this is not a blocker)

agaertner: It may have had something todo with letting the OS handle the keyboard layout and not ourselfs since ``ToUnicodeEx`` expects a local identifier of a keyboard layout used in translating the code ie. it appears to be just a means to force a different keyboard layout used in translation if you'd want that. ``ToUnicode`` uses the current active one.

Blish HUD/GameServices/Input/Keyboard/KeyboardHandler.cs
agaertner: Good catch.

Blish HUD/GameServices/Input/Keyboard/TypedInputUtil.cs
entrhopi: Yes. @greaka and I were discussing this. Since we don't need to inject our own layout at any point we ditched the `ToUnicodeEx`. Reading the current layout and injecting it would have been unnecessary and redundant code.

greaka: `ToUnicodeEx` requires us to handle keyboard layouts ourselves which `ToUnicode` does on its own. When a user switches layouts, we either have to poll layouts again every update or we don't support switching layouts. We just don't want or need the control `ToUnicodeEx` provides

dlamkins: Alright, thanks for the explanations!  That makes perfect sense.

dlamkins (approved):

dlamkins:
Closes #5!

PR #405, opened 2021-07-30
Random Blish crash during typing in textbox

entrhopi:
I noticed random crashes because of a `NullRefernceException` in `public Rectangle AbsoluteBounds`. I think it's because in rare cases there could be a race condition for `_parent` which then would be `null` before the last `return` statement.

Adding `else` seems to help with that issue, preventing change of `_parent` before the `return` is executed.


entrhopi:
Update:
Sorry, didn't notice PR #402 modifying `Control.cs` but I would still suggest putting the rest of the function in an `else`.

PR #406, opened 2021-07-31
🎉 Add fonts

agaertner:
We need to offer more fonts.

These fonts are all free to use for non-commercial projects.

Added 6 readable fonts plus 3 ideograms (Canthan, Ascalonian, Asuran)
Fixed New Krytan

Total: 9 fonts added.

- The 6 free fonts added are a close match to commercial fonts that ArenaNet uses.
- All italic support

```csharp
        public enum FontFace {
            [Description("Guild Wars 2 UI font")]
            Menomonia,
            [Description("Guild Wars 2 Title font")]
            GWTwoFont,
            [Description("Substitute to Eason Pro Inline Caps")]
            Elizabeth,
            [Description("Substitute to Eason Pro Inline Caps")]
            StoweOpenFace,
            [Description("Substitute to Eason Pro")]
            StoweTitling,
            [Description("Substitute to Eason Pro")]
            PTSerif,
            [Description("Substitute to Cronos")]
            Lato,
            [Description("Substitute to Cronos")]
            Cagliostro,
            [Description("New Krytan Typeface")]
            NewKrytan,
            [Description("Ascalonian Typeface")]
            Ascalonian,
            [Description("Canthan Ideogram")]
            Canthan,
            [Description("Asuran Script")]
            Asuran
        }
```

**Matching fonts**
Elizabeth and Stowe Open Face which match the font in this [picture](https://oyster.ignimgs.com/mediawiki/apis.ign.com/guild-wars-2/7/7e/Elementalist.jpg).
PTSerif, Lato and Cagliostro which match the Guild Wars 2 main font. (See [forum thread](https://en-forum.guildwars2.com/topic/3702-free-for-use-fonts-similar-to-recommended-by-guild-wars-2-asset-kit/))

**Other fonts**
Stowe Titling (Same as Stowe Open Face but non-hull letters.)
GWTwo Font (known fan-made title font)

**Typefaces/Ideograms from the game**
Canthan Ideogram
Ascalonian
New Krytan (See https://www.properdave.com/gw2/)
Asuran

- Removed the old bad new krytan fnt conversion from the Content pipeline. It wasn't hooked up with the FontFace enum anyway.

~~These fonts are pre-compiled and included in the project as "Content" with "Copy If New" set since this works way better and faster than to try and get fonts into the fnt format for MonoGame to process them with BMFont (Remember; I tried converting New Krytan before and it was not only tidious and complex but I only got loosely results.). It's also the method for ttfs described in MonoGame docs.~~ Not anymore since this loads them as SpriteFonts which is incompatible with DrawStrgOnCtrl. I've converted every font to a BitmapFont and included them in the Content.mgcb.

I added a small extension method to make use of an "Description" tag on enums because some font names included a dot and I thought I had to name the file names according to the reference to the installed fonts (for compilation) which I realized is not the case. I was able to rename the compiled versions. I kept the method because I think it might come in handy later (displaying friendly names for example).

I had to edit a little bit of code under the ContentService to fallback to Regular if the specified style is not available for a certain font (only the ideograms). We are able to offer every size but not every font has every style.

I would appreciate it very much if this set could be the exception in stuff that core doesn't directly use but offers to modules.


agaertner:
Updated post to show italic support.

dlamkins:
#423 we will track the plans/progress for the new font handling system.  Once implemented, modules will be able to use any combination of fonts that they wish to use instead of us integrating them into core.

PR #412, opened 2021-08-07
Form and idle fixes

### LINKED ISSUE #76: Blish HUD shows as a black square until GW2 is focused for the first time.
**QuitarHero at 7/28/2019 10:21:43 PM +00:00**
> Minor bug: When GW2 is not the active window and Blish HUD is reloaded, this black square will show up and stay there permanentally until GW2 becomes the active window, where it goes away without issue.
> Very minor, but thought I should mention.
https://cdn.discordapp.com/attachments/534492173362528287/605163004501426235/blish_hud_buggy.PNG

REF: https://discordapp.com/channels/531175899588984842/534492173362528287/605163005197549579

### LINKED ISSUE #304: Blish HUD fails to hide itself on the initial launch if game is not already running
**Metallis at 9/16/2020 8:36:36 PM +00:00**
> if blish is running before gw2 client opens (at least this once), tips and tricks show on character select page. also golem icon, markers icon, and search icon all show too

REF: https://discordapp.com/channels/531175899588984842/534492173362528287/755889902209597551

### LINKED ISSUE #411: High CPU / GPU / Memory usage while GW2 is not running
CPU and GPU utilization are far too high while Blish HUD is idle and GW2 is not running.

dlamkins:
- Exits now have a lock to prevent them from running twice.
  - Exit calls were happening quick enough that they could sometimes trigger twice even with a bool to keep record of it.
- Removed flash when the application launches.
  - Monogame source appears to use a subclass'd form which forces the form to become visible again despite us hiding it at launch.  This causes the window to show and flash despite us attempting to hide it.  To avoid this, we just move it offscreen until we're ready.
- Fix resource utilization when Guild Wars 2 is not running.
  - Instead of attempting to unload everything when the game closes (hard to trust modules), we just quickly restart Blish HUD.
  
Closes #411
Closes #304
Closes #76

### DIFF
--- a/Blish HUD/BlishHud.cs
+++ b/Blish HUD/BlishHud.cs
@@ -1,8 +1,9 @@
 ﻿using Microsoft.Xna.Framework;
 using Microsoft.Xna.Framework.Graphics;
 using System;
-using System.Diagnostics;
+using System.Drawing;
 using System.Threading;
+using System.Windows.Forms;
 
 namespace Blish_HUD {
 
@@ -28,7 +29,7 @@ namespace Blish_HUD {
 
         public IntPtr FormHandle { get; private set; }
 
-        public System.Windows.Forms.Form Form { get; private set; }
+        public Form Form { get; private set; }
 
         // TODO: Move this into GraphicsService
         public RasterizerState UiRasterizer { get; private set; }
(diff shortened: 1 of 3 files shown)


greaka (approved):
It is unfortunate that we need to restart bhud

dlamkins:
Yeah, not much we can do - especially if any strings are interned or assemblies are loaded.  No way to reclaim.

PR #414, opened 2021-08-15
Nicer smoothed player position

seark:
I changed the smoothing of the player position to be relative to the camera position as already done by gw2Taco.
This should result in a much smoother player position unless you rotate the camera and move at the same time.

### DIFF
--- a/Blish HUD/GameServices/Debug/DynamicallySmoothedValue.cs
+++ b/Blish HUD/GameServices/Debug/DynamicallySmoothedValue.cs
@@ -16,7 +16,7 @@ namespace Blish_HUD.Debug {
         /// </summary>
         public T Value => _cachedValue ??= GetAverageValue();
 
-        private readonly Func<int> _getBufferLengthFunc;
+        private int _currentSize;
 
         private readonly Func<T, T, T>     _addExpression;
         private readonly Func<T, float, T> _divideExpression;
@@ -26,9 +26,8 @@ namespace Blish_HUD.Debug {
         /// </summary>
         /// <param name="bufferMax">The maximum length of the buffer.</param>
         /// <param name="getBufferLengthFunc">The function used to decide how much of the buffer should be used to calculate the average.</param>
-        public DynamicallySmoothedValue(int bufferMax, Func<int> getBufferLengthFunc = null) : base(bufferMax) {
-            _getBufferLengthFunc = getBufferLengthFunc ?? (() => bufferMax);
-
+        public DynamicallySmoothedValue(int bufferMax) : base(bufferMax) {
+            _currentSize = 0;
             _addExpression    = PrepareAddExpression<T>();
             _divideExpression = PrepareDivideExpression<T>();
         }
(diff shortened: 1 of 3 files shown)


dlamkins:
Once CI gets a build of this spit out, we'll definitely want a few folks to test and weigh in on the differences it provides.

dlamkins:
In my initial testing of this, I am seeing a slight improvement from the current implementation, but I'm seeing far worse tracking when the camera is moved making this perhaps not a good overall replacement.  Are you experiencing these same results?

**No Smoothing:**
https://gfycat.com/equalhiddencrustacean

**Current Implementation:**
https://gfycat.com/admirableexcitableasiansmallclawedotter

**This PR's Implementation:**
https://gfycat.com/keynegligiblehawaiianmonkseal

seark:
The current implementation is pretty good with high fps. If you limit your fps to for example 30, you will see the circle lagging behind the character.

I tried to fix that. Problem with using the camera as reference is if you turn the camera, you get lagging behind the camera movement (which is much worse than lagging behind the character). So I added a flush of the ringbuffer on camera movement. This results in nice results if moving the camera and if moving the character but it stutters if you try to do both at the same time.

Maybe a much simpler solution would be to increase the update rate independent of the framerate. At least for the player position.

dlamkins:
Seems like the ideal setting has the overlay set to update synced with the in-game refresh rate as is the default.  I think locking to 30 or 60 is only good if you don't have in-game 3D elements moving around. 🤔

seark:
At least for me synced with in-game refresh rate is resulting in lagging player position updates. Maybe this is because my framerate drops to about 40 in some regions. If I choose unlimited the current implementation is working pretty well.

I tried another approach using the map position, which results in a different set of problems. So i've got no good alternative to the current implementation for low frame rates without disconnecting frame update and player position update.

dlamkins:
We may wish to either make this dynamic so that it uses a different method depending on the FPS or keep discussing our options.  What issues did you run into with the map method?  That's unfortunate to hear.  😕

AsherGlick:
I tried doing this camera-syncing for [Burrito](https://github.com/AsherGlick/Burrito) too and found most of the same issues. Many attempts later I did come up with a solution, which turned out to be quite simple. A basic rolling average with a 5 element lookback. Here is a chart I used to decide on the solution.
Screenshot from 2021-10-04 03-21-32

Here is the c code I used, it is a conflicting license with Blish HUD but I wrote it myself so dont worry about it.
https://github.com/AsherGlick/Burrito/blob/55fff1edc7b2bd4861cb72c9609d3fe14ba95fb3/burrito_link/gw2mumbleudp.c#L6-L34
https://github.com/AsherGlick/Burrito/blob/55fff1edc7b2bd4861cb72c9609d3fe14ba95fb3/burrito_link/gw2mumbleudp.c#L232-L239

And this is the result, and old dev video but shows the smoothness. The two rings are obvious but there is also a small dot at the feet if you zoom in.

https://user-images.githubusercontent.com/603989/138618167-d2c04b36-7e5d-409c-8c92-d37fbe6760d6.mp4

Hope that helps you develop your solution!

dlamkins:
We've been doing something similar but instead of 5 with 8 points.  I found its success to vary significantly with framerate.  At higher framerates the smoothing works incredibly well, but it quickly fell apart at lower framerates.  🤔  I built out some visualizations to use when I initially reported this, but really like your method of tracking your accuracy vs. the camera offsets.  That would also make our testing a lot clearer, too.

Thanks!

AsherGlick:
I haven't read through the BlishHUD source code, or experimented much on the lower framerate side, but Burrito only adds a new element whenever `uiTick` shared memory variable updates so we don't run into mixed framerate issues between the game and the overlay. The camera object updates every tick in my testing but the player movement does not. Before doing this I found I sometimes double polled the same value for position which caused the rolling average to be off because it contained two of the same values, everything else was atomic so double polling did not cause any issues.

In my testing both a rolling average of 4 and 6 cause a looser fit to the camera positioning. Though if it is framerate related I wonder if there is a correlation that can be determined between the number of points to average and the current framerate.

PR #418, opened 2021-08-30
Window construction improvements

dlamkins:
This PR introduces several major changes:

### WindowBase
This control base class is now deprecated with the expectation it will be removed in the future.  It has been replaced by `WindowBase2`.

### TabbedWindow
This control is now deprecated with the expectation it will be removed in the future.  It has been replaced by `TabbedWindow2`  As it stands, it is only used by Core and has been far closer to a concrete implementation of a tabbedwindow than any even close to customizable for other uses.  It did not expose proper methods of using it for other purposes.

### WindowTab
This class is now deprecated with the expectation it will be removed in the future.  It has been replaced by `Tab` which offers many of the same details but importantly provides a View so that it can be displayed instead of the `ITabOwner` having to manage tabs and their associated views itself.

---
### WindowBase2
WindowBase2 is a base window implementation which supports rendering all parts of a standard Guild Wars 2 window.  This includes:
- Drawing the black sidebar.
- Drawing the titlebar.
- Drawing the window emblem.
- Drawing the exit button.
- Drawing the corner (if resizable: the resize handle instead)
- Dragging the control via the titlebar.

`WindowBase2` is distinct from `WindowBase` in that it provides simplified `ConstructWindow` calls, is less pointlessly "generic", and is a `IViewContainer`.

`WindowBase2` owns the methods to register and unregister `IWindow` controls to allow them to act as windows regardless of their specific implementation details.

### StandardWindow
StandardWindow inherits from `WindowBase2` and acts as a normal Guild Wars 2 window without a sidebar (like the in-game Mail window).  It offers methods to display a view within its content region.

**Documentation for this new control**: [StandardWindow](https://blishhud.com/docs/dev/controls/standardwindow/)

control_StandardWindow_0 4947395

### TabbedWindow2
TabbedWindow2 inherits from `WindowBase2` and acts as a normal Guild Wars 2 window with a tab sidebar.  It offers a `Tabs` collection which can have tabs added or removed.  Selecting a tab will display the associated view in the windows content region.

**Documentation for this new control**: [TabbedWindow2](https://blishhud.com/docs/dev/controls/tabbedwindow2/)

control_TabbedWindow2_0 4642836

### IWindow
Any control can now implement `IWindow` allowing it to be registered with `WindowBase2` to have the ZIndex automatically calculated.  `WindowBase` and `TabbedWindow` have this now implemented to ensure they behave nicely with the new `StandardWindow` and `TabbedWindow2` controls.

---
### IView `Panel` changed to `Container`

IViews were mistakenly set to build their views with the expectation that a `Panel` would be passed after the ViewContainer implementation was split out to make it more generally available to other containers.  This meant that only controls implementing Panel could implement `IViewContainer` properly.  Since `Panel` is meant to just be one such container implementation, this doesn't make sense and needed fixed to ensure that `WindowBase2` could properly implement `IViewContainer`.

### Tooltip changes to inherit `Container`

The above was missed because when this change was made, it was part of a change to allow Tooltips to act as view containers.  Tooltips were mistakenly inheriting `Panel` instead of `Container` which included many things which were completely unnecessary to their implementation.  They have been switched to properly inherit `Container` instead.

Both of these changes are expected to have a small impact.  Few modules use views.  The pathing module uses them extensively, but can be updated in a matching prerelease.  Existing view implementations will continue to run, but may run into unexpected exceptions if they attempt to access the `ViewTarget` after the view has already been built (e.g. the presenter attempts to directly access it) as the ViewTarget will be null since the provided buildPanel is likely unable to cast to Panel.

---
### `Container.GetDescendants()`
We now uses a new implementation to make it a bit faster - especially when iterating the descendants instead of actually needing all of them as it yield returns instead and can thus exit early.

### ITabOwner
This interface now exists with the intention that it is also used by an upcoming TabPanel control.


Blish HUD/Controls/WindowBase.cs
```diff
@@ -6,7 +6,7 @@
 using Microsoft.Xna.Framework.Graphics;
 
 namespace Blish_HUD.Controls {
-    public abstract class WindowBase : Container {
```
greaka, line 9: This is missing a deprecation warning

greaka (approved):
Great change :slightly_smiling_face:

Blish HUD/Controls/WindowBase.cs
dlamkins: Good catch!

Issue #227, opened 2020-05-20
Feature: Pressing Ctrl+Shift+H hides Bhud like GW2 UI.


BlishHUD:
**Tybalt at 5/20/2020 12:53:07 AM +00:00**
> When you press Ctrl + Shift + H BlishHUD should also hide like GW2 UI 😄

REF: https://discordapp.com/channels/531175899588984842/534490472224391169/712467914023895091

Issue #70, opened 2019-07-21
esc key close active Blish HUD window, if focused


BlishHUD:
**Husky at 2/7/2019 9:38:58 PM +00:00**
> not sure if it's been mentioned, but can the Esc key drop the overlay if it's the focussed window?

REF: https://discordapp.com/channels/531175899588984842/534490472224391169/543183919416213525

dlamkins:
> **Suggestion:**
> > Have Esc close Blish HUD windows. It must be a toggle in settings.
> > Reason: GW2 has a toggleable option to close windows via Esc, and in the vein of replicating GW2 functionality, Blish HUD should have it too.
> 
> Esc prioritization should always be given to Blish HUD before GW2.

QuitarHero - https://discordapp.com/channels/531175899588984842/534490472224391169/612368661835022336

dlamkins:
Closed in #443!

PR #422, opened 2021-08-31
Feature for #70 and #227

entrhopi:
This is going to be a 'fix' for the feature requests #70 and #227

Input and suggestions are welcome.


dlamkins:
With the recent #443 merge, we can retarget this for just #227.  :)

dlamkins:
The setting for ESC close does still seem like a good idea, though.  👍

entrhopi:
ESC Key issue #70 has been adressed by PR #443. I'll expand this to incorporate the toggle in the settings.

dlamkins:
Testing build...

dlamkins:
Okay, seems to be nearly there.  I did find that, while the textboxes, etc. don't accept input when the UI is hidden, their hook still blocks all keyboard in put, so in the rare scenario that a user has a textbox selected and they press CTRL + SHIFT + H, their character will be stuck and unable to move.

Perhaps we can move the early exit (when we've detected the UI hidden) to earlier in the hook so that we don't ever block while the UI is hidden.  We can still let input through for triggering the keybindings themselves, but can shortcircuit the response so that we don't block.

entrhopi:
Good catch! Thank you, I'll address that immediately.

entrhopi:
So I decided to change it so that the textbox will loose just focus upon hiding the interface.
It was the easiest solution and seemed logical. I did not implement a restore focus option because effectivly the user did just decide to 'do something else' in the game which would result in focus loss anyways.

The other option would be to replicate the games behavior and block the hotkey entirely when an input is focused.

Please discuss.

entrhopi:
Still hase some issues sometimes, 🤔 

I think I'll wait for your input on what direction I should take and then continue implementing ...

dlamkins:
I think the idea of it losing focus with no re-auto focus is certainly fine.  I'll do some testing tonight on those edge cases and see what I can find.  👍

entrhopi:
Great, I'll push my latest revision later today for testing.

dlamkins:
What if we put the early exit for the keyboard hook somewhere around here?
https://github.com/blish-hud/Blish-HUD/blob/1c7670f92fc9e28eb0727925ed041928e1386df2/Blish%20HUD/GameServices/Input/Keyboard/KeyboardHandler.cs#L147

We'd skip all of the window ESC logic and the textfield input logic (the delegate is handled further down) while still enqueuing keyboard input into the handling buffer for keybindings.

It's not perfect because we'll still handle other keybindings (besides toggling the UI) but maybe that's actually not such a big deal and should still be the expected behavior.  For example, a mount module with special hotkeys for mounts you likely would still want to activate, even with the UI is hidden.

Does that cover the remaining edge case or where there other issues you were running into?  🙂

dlamkins:
Okay, this seems to work great.  Were there any other edge case you noted?  If not, I think this is ready to merge.  🙂

entrhopi:
> Okay, this seems to work great. Were there any other edge case you noted? If not, I think this is ready to merge. 🙂

I haven't found any other edge cases.
Added a TODO comment to move some of the logic from `TextInputBase` to somwhere else in the future.

dlamkins (approved):
Thanks for all of this!  Fantastic!

dlamkins:
Closes #227!

PR #426, opened 2021-09-01
Make Blish portable by creating a Settings directory in path

entrhopi:
Creating a 'Settings' directory in Blish will shift all settings to be defaulted to there instead of 'My Documents'.

This helps with easy portable deployment without using `--setting` parameter.

`--settings` is still honored if set as parameter.

### DIFF
--- a/Blish HUD/_Utils/DirectoryUtil.cs
+++ b/Blish HUD/_Utils/DirectoryUtil.cs
@@ -39,9 +39,12 @@ namespace Blish_HUD {
             // Check if current working directory contains "Settings" folder
             // in that case override MyDocuments location as default for portability
             // --settings cli argument has still priority
-            if (Directory.Exists(Path.Combine(Directory.GetCurrentDirectory(), @"Settings"))) {
+            string BlishExeFilePath = System.Reflection.Assembly.GetExecutingAssembly().Location;
+
+
+            if (Directory.Exists(Path.Combine(Path.GetDirectoryName(BlishExeFilePath), @"Settings"))) {
                 BasePath = ApplicationSettings.Instance.UserSettingsPath
-                        ?? Path.Combine(Directory.GetCurrentDirectory(), @"Settings");
+                        ?? Path.Combine(Path.GetDirectoryName(BlishExeFilePath), @"Settings");
             }
             else {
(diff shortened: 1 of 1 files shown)


entrhopi:
Should we call the directory 'Settings' or 'Portable'?

Blish HUD/_Utils/DirectoryUtil.cs
```diff
@@ -36,10 +36,22 @@ public static class DirectoryUtil {
...
         static DirectoryUtil() {
             // Prepare user documents directory
-            BasePath = ApplicationSettings.Instance.UserSettingsPath
-                    ?? Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.MyDocuments,
-                                                               Environment.SpecialFolderOption.DoNotVerify), 
-                                    ADDON_DIR);
+            // Check if current working directory contains "Settings" folder
+            // in that case override MyDocuments location as default for portability
+            // --settings cli argument has still priority
+            string BlishExeFilePath = System.Reflection.Assembly.GetExecutingAssembly().Location;
```
dlamkins: No need to pull it this way.  You can pull it by the current directory under Directory since we've set it here:

https://github.com/blish-hud/Blish-HUD/blob/fe56e84006b5aa3deae89c85faa4452fb1c727fc/Blish%20HUD/Program.cs#L43

Blish HUD/_Utils/DirectoryUtil.cs
```diff
@@ -36,10 +36,22 @@ public static class DirectoryUtil {
...
+
+
+            if (Directory.Exists(Path.Combine(Path.GetDirectoryName(BlishExeFilePath), @"Settings"))) {
+                BasePath = ApplicationSettings.Instance.UserSettingsPath
+                        ?? Path.Combine(Path.GetDirectoryName(BlishExeFilePath), @"Settings");
+            }
+            else {
```
dlamkins: Lets stick the else on the same line as the `}` above it.

dlamkins (changes requested):
Great idea!

dlamkins:
> Should we call the directory 'Settings' or 'Portable'?

I thing `Settings` is the most appropriate.  Changes look good.  Thank you!

dlamkins (approved):

PR #435, opened 2021-09-07
Ensure we don't render on desktop if game is minimized

### LINKED ISSUE #428: When GW2 is minimized, Blish HUD is still displayed
Blish HUD still displays over the desktop if GW2 is minimized.  Blish HUD should hide when GW2 is minimized.

dlamkins:
Instead of relying on focus order, we just detect if the game is minimized and set our parent form visibility to match as appropriate. 

Fixes #428


greaka (approved):
Let's hope that displays never reach the 30k resolution mark :grin:

PR #441, opened 2021-09-11
Allow using a custom Tooltip for CornerIcon

vacero:
Hi! This is my first PR here! I am tinkering with creating a module for Blish HUD and I wanted to use a custom `Tooltip` for a `CornerIcon`. I saw that the code for `CornerIcon` is always setting `BasicTooltipText` so it is not really allowing that, as `Control` uses this field with more priority.

Does this look like a good approach? Feel free to change whatever or let me know if I need to make any changes!

BTW, I am Fogokhost#5194 on Discord 😄


dlamkins (approved):

dlamkins:
I hadn't even considered it.  Good catch!  Looks great.  👍

PR #443, opened 2021-09-12
Feature #70: Close active window on escape.

GregoryComer:
Implementation of feature #70 - close windows and top-level context menus on escape press.

### DIFF
--- a/Blish HUD/Controls/_Types/IWindow.cs
+++ b/Blish HUD/Controls/_Types/IWindow.cs
@@ -5,6 +5,11 @@
     public interface IWindow {
 
         /// <summary>
+        /// Returns true if the window is currently visible.
+        /// </summary>
+        bool Visible { get; }
+
+        /// <summary>
         /// If the window should be forced on top of all other windows.
         /// </summary>
         bool TopMost { get; }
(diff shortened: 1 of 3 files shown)


Blish HUD/Controls/_Types/IWindow.cs
```diff
@@ -24,5 +29,16 @@ public interface IWindow {
         /// </summary>
         bool CanClose { get; }
```
dlamkins, line 30: We might want to incorporate `CanClose` into this somehow so that a window that implements `IWindow` that doesn't intend on being closed by user input (so no X or no closing on ESC) can exist.

Might just be that if the active window is a window that has `CanClose` set to false, we just ignore it and let the ESC pass through to the game.  Thoughts?

dlamkins (changes requested):

Blish HUD/Controls/_Types/IWindow.cs
GregoryComer: I have added a check to conditionally handle the escape key depending on the CanClose property of the active window.

dlamkins (approved):

dlamkins:
Closes #70!

PR #448, opened 2021-09-13
Color picker improvements

vacero:
- Reduced the default size of the ColorBoxes of the ColorPicker to match the in-game size.
- Add texture for hover and active ColorBoxes.
- Fixed current selected color shown as active in the picker.
- Texture selection based on ColorBox size.
- Size of ColorBoxes in a ColorPicker can be specified through a new `ColorBoxSize` property

Naming of stuff can be a bit off, I changed things around a lot 😅 

This is how this looks like now by default (side-by-side comparison for GW2, left and BHUD, right):
image

We can also have biiig color boxes:
image

Or long ones:
image

We can even have a color picker like this one, because why not:
image
This last one glitches out a bit (notice the colored lines between colors), but also I don't think it will be used like this 😺


dlamkins:
Thank you for this! Looks pretty cool!  This one I need to do some testing with so I apologize if it takes a couple of days to get done. 👍

dlamkins (approved):
Sorry for the wait.  This looks great.  Thanks again!

PR #449, opened 2021-09-13
WindowsBase2 Resizing Changes

manlaan:
Added Saving Size to work with Resizing
Fixed resizing to resize title and background
Changed bottom resize box to be within ContentBounds to allow better positioning within Window
Changed bottom resize box size so its not so large
Moved Exit button a little to be more centered (Blish exit box is a bit smaller than in game box, but aligned top and left, making a bit off centered in the toolbar)

Likely will need a bit of cleanup.
I didn't test how these changes would affect tabbedWindow2.


manlaan:
btw, double check what I did with the background, as you may not like it.  It felt a bit better to me on positions and sizes like this, but I understand if this wasn't the intent on how it should work.

dlamkins:
Thank you for this!  This one I need to do some testing with so I apologize if it takes a couple of days to get done.  👍

dlamkins:
Resizing seems to be a good start.  The sidebar will need to be adjusted, though.  Not sure what the behavior should be if the active sidebar height is larger than the window height.  An easy answer might just be that the default implementation limits the vertical minimum size to that the of the sidebar height if it's enabled.

https://gfycat.com/sociableneglectedasiaticlesserfreshwaterclam

TabbedWindow2 has a lot of issues:

control_TabbedWindow2_0 4507242

Exit button repositioning seems odd.  At least compared to the in-game windows.

image

Could you elaborate some more on some of the changes made to the different regions (presumably what's making TabbedWindow2 come out janky) and the exit button changes?

I've not tested a StandardWindow yet.

manlaan:
I think I took it a bit too far with the Exit button, trying to make it evenly spaced on both sides, when in game isn't that way.   OCD issues...   I do think it still does need to be moved right a bit more than Blish has it currently though, or better yet, the title bar/exit button adjusted to be the same size as in game.   Maybe its an issue with UI Scale.  

As for the background, I really dislike that background image, as it just has so much white space in it and makes it impossible to know where the actual boundaries of the image are.   Maybe its the intent that the background image is to have the title bar and tabs placed over it?   Not sure what the intent is to have so much white space on the bottom of that background though.  

I am curious as to what the background would look like using a background color instead.   It does look like there is still a margin/padding issue with placement with the tabs though, so I'll be seeing if I can find where that is applying it this weekend sometime.

dlamkins:
> or better yet, the title bar/exit button adjusted to be the same size as in game.

It does appear to be ever so slightly off in size, so we can work to adjust that.  Our goal would generally be to match the in-game design and only deviate where it won't distract (or where it maintains the spirit of the original design).  Since the windows can be next to other in-game windows, we don't want to distract by having our exit icons moved over and show visibly off from the actual game windows, if that makes sense.

> As for the background, I really dislike that background image, as it just has so much white space in it and makes it impossible to know where the actual boundaries of the image are. Maybe its the intent that the background image is to have the title bar and tabs placed over it? Not sure what the intent is to have so much white space on the bottom of that background though.

That's simply how the game includes them if you export them from the dat.  GW2 window backgrounds have quite a bit of fading bits that extend beyond their normal region.  It's part of the game's art style.  The blank space below and to the right of the texture isn't particularly relevant but the space to the left and above is very important as that includes the faded window portions that aren't technically in bounds.

Regarding the boundaries, that's what the [blue rectangle region](https://blishhud.com/docs/dev/controls/standardwindow/) is meant to represent when you create a new window.  You tell it where the window bounds are.   You can see in the examples for both the [StandardWindow](https://blishhud.com/docs/dev/controls/standardwindow/) and [TabbedWindow2](https://blishhud.com/docs/dev/controls/tabbedwindow2) the portions that are outside of the windowRegion and are intended to show outside of the actual windows bounds.

The windows are technically rectangles with strict boundaries, but visually are not, so this is the best way I can think of to account for it without forcing you to specify a bunch of padding numbers and whatnot.

> I am curious as to what the background would look like using a background color instead. It does look like there is still a margin/padding issue with placement with the tabs though, so I'll be seeing if I can find where that is applying it this weekend sometime.

If you elaborate some on that, I can take a look as well.

Tab placement is meant to be overlapping with the window background so that the faded edge can peek out from behind the selected tab.

manlaan:
ok, I now understand the intent behind it.  It shouldn't take much to make it work.   I think I'll just need to change the background bounds to 
-window width, (like you had it)
-window height,  (like you had it)
window width + maybe window left so the titlebar is centered
window height + ???  not quite sure on this one, specially considering how much white space is on the bottom of the gw2.dat file.

dlamkins:
Might be something where we have to calculate the difference between the windowRegion and the contentRegion, scale down the main region when resized while maintaining the difference between the windowRegion and the contentRegion (or something to that effect).  🤔

manlaan:
I was playing around with a few different ways to handle it, and overall, it just wasn't working right, as you can tell my looking at my latest ways I was messing with modifying it before I decided I needed to take a break from it.  

I pretty much got stuck trying to modify the whole background.   I was thinking about trying to find a way to keep the background outside of the window area so it only adjusts in 1d (left/right only can adjust up/down, and top/bottom only adjust left/right), and be able to adjust the background inside the window area freely.   Not sure how badly this will distort things, but that was the direction I was leaning.  

I did run into a few issues, which I believe you ran into as well.    Dragging the window size larger than the original window size made the resize corner unclickable.   I did add in some additional checks for that so it couldn't be resized larger than the original background size.  

I also ran into resizing being an issue, and found the easiest way to deal with it is to allow the user to use what they want for the background, then apply a resize amount point that adjusts the window by x, y.  Seems to work ok on standard windows.  Haven't played with tabbed windows yet.

dlamkins:
I'll give it a look!

dlamkins:
Hey, do you think you could either revise this PR or submit a new one that just has the SavesSize property and its implementation (i.e. https://github.com/blish-hud/Blish-HUD/pull/449/files#diff-41b9fefd27226c75fce654bf580e33f6dea3a8e70d6baaa60762947520c51efcR286-R296) for now?

I think we'll need to poke at this a bit more, but it would be good for us to get the property available in the API prior to the next release so that when the rest of the resizing support comes, modules don't need to update to use it and it will just start working.

dlamkins:
Majority of the changes here now resolved in #688.

PR #459, opened 2021-09-22
Add extended key support to Keyboard

manlaan:
There really isn't a lot of information about this, strangely.  I believe most people just use SendKeys instead, but I dont think that'd work great in game.   

MS has a note about it at https://docs.microsoft.com/en-us/windows/win32/inputdev/about-keyboard-input#extended-key-flag    It does work without the Extern.Input that has a wScan = ScanCodeShort.EXTENDEDKEY (0xe0), but since MS said it should be there and the other example I found had it, I left it in.  

About the only other "helpful" information I found was at https://www.michaelwda.com/post/scancodes , but he put me in the right direction.


dlamkins (approved):
Looks great!

Issue #452, opened 2021-09-20
Detect mumble name from command line


dlamkins:
Detect mumble name from command line using `ManagementObject` to pull the CommandLine parameters.

Querying WMI is incredibly slow (like approaching full seconds) so we'll want to be mindful of this when implementing.  Ideally we'd give users the options to attach to an instance in the context menu strip or in UI if we manage to attach to an instance without detecting mumble connected.

Friendly0Fire:
Is querying WMI in a worker thread feasible? Feels like that'd deal with any potential hiccup. You'd only want to run it after you've confirmed the newly focused window is a GW2 window, and then you'd associate the PID of that instance with its MumbleLink instance so it only needs to run once.

dlamkins:
We'd most certainly be able to do that. Wanted to make note of it in case
somebody other than myself decided to try and implement it. 👍

dlamkins:
Closed by #452

PR #462, opened 2021-09-22
Initial implementation of https://github.com/blish-hud/Blish-HUD/issues/452

eksime:
Added the ability to get the current command line from `Gw2InstanceIntegration` and used that as the mumble link file name.

The mumble link file name will prefer the one defined in settings then fall back to the command line, then the default, though I'm not sure if the way I'm swapping between `IGw2MumbleClient` instances is best practice(?). They're not disposed immediately, but they should get disposed when the gw2 client is disposed or they get collected - it looks like there's a weakreference-based cache in gw2sharp.

I thought it may be nice to leverage the EntryPoint package to parse the command line for gw2, however it doesn't appear to support long-form args with a single `-` (as in `-mumble`), so this is using a regex to determine the mumble file from the command line.

P/Invoke is used to get the command line info, with a slightly different pinvoke fallback method for Windows 8.0 and 7 machines, as it it much faster than querying WMI.

### DIFF
--- a/Blish HUD/GameServices/GameIntegration/Gw2InstanceIntegration.cs
+++ b/Blish HUD/GameServices/GameIntegration/Gw2InstanceIntegration.cs
@@ -5,6 +5,7 @@ using System.Diagnostics;
 using System.IO;
 using System.Security.AccessControl;
 using System.Windows.Forms;
+using Blish_HUD._Utils;
 using Blish_HUD.GameServices;
 using Blish_HUD.Settings;
 using Gapotchenko.FX.Diagnostics;
@@ -120,6 +121,15 @@ namespace Blish_HUD.GameIntegration {
             private set => PropertyUtil.SetProperty(ref _appDataPath, value);
         }
 
+        private string _commandLine;
+        /// <summary>
+        /// The full command line of the current Guild Wars 2 process.
+        /// </summary>
+        public string CommandLine {
+            get => _commandLine;
+            private set => PropertyUtil.SetProperty(ref _commandLine, value);
+        }
+
         // Settings
         private SettingEntry<string> _gw2ExecutablePath;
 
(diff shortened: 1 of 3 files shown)


dlamkins:
This is looking fantastic!  I'll need to do a bit of testing but I'm already pretty excited to do that.  I'm glad you found an alternative to querying WMI.

@Archomeda could you comment on the method used here for swapping `IGw2MumbleClient` instances?

Archomeda:
@dlamkins Ah, [technically it should be fine](https://archomeda.github.io/Gw2Sharp/master/guides/introduction.html#custom-mumble-link-names). Of course it's always better to dispose of the custom names yourself, but Gw2Sharp should dispose all of them as a catch-all.

You'll be the first to use this feature to my knowledge, and definitely the first in this way. There are a few tests in Gw2Sharp to make sure that it's working as intended, but I haven't tested it myself extensively and I don't trust this combination of WeakReferences, caching, indexer and disposing myself completely. There's the possibility to get infinite recursion with the indexer, and I have no clue whether disposals are handled perfectly, and whether the cache is correctly used. Let me know if any bugs creep up.

dlamkins:
Works for me and pretty excited for this. 👍

Closes #452.

Blish HUD/_Utils/ProcessUtil.cs
```diff
@@ -0,0 +1,158 @@
+﻿using System;
+using System.ComponentModel;
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+using System.Text;
+
+namespace Blish_HUD._Utils {
```
dlamkins: I apologize as we don't have this really written here.  Any of the directories that are prefixed with `_` are intended to not be part of the namespace.  That is to say, the utils would be placed directly into the Blish_HUD namespace.  With this being pretty specific in purpose we can probably mark it internal or move it to a namespace under gameservices next to where it's used as a internal util there.

Blish HUD/GameServices/GameIntegration/Gw2InstanceIntegration.cs
```diff
@@ -5,6 +5,7 @@
 using System.IO;
 using System.Security.AccessControl;
 using System.Windows.Forms;
+using Blish_HUD._Utils;
```
dlamkins: Once the util's namespace (mentioned below on that new class) is revised, we'll want to remember to delete this line.

dlamkins (changes requested):

Blish HUD/_Utils/ProcessUtil.cs
eksime: Updated the PR to make the class internal & removed the _Utils namespace

dlamkins (approved):

dlamkins:
Looks great!

PR #465, opened 2021-09-23
Fix NullReferenceException on gw2 process exit

eksime:
I was reliably hitting a NullReferenceException on exit as GW2Process was null when calling WindowUtil.GetClassNameOfWindow, this change early-out's when the new process is null.


dlamkins:
Good catch!

dlamkins (approved):

Issue #468, opened 2021-09-25
Don't fall back if connection method is explicit


dlamkins:
https://github.com/blish-hud/Blish-HUD/blob/cfbdea1a331bbc903ea5954cfc03c9d0f62d776b/Blish%20HUD/GameServices/GameIntegration/Gw2InstanceIntegration.cs#L201-L203

Currently we properly prioritize our load order, but we fallback if the method fails which allows us to connect to sessions that we weren't intended to connect to.

For example, if a mumble link name is specified, we shouldn't fallback if another instance is found.

dlamkins:
Closed in #469

PR #469, opened 2021-09-25
Fix for #468

eksime:
Fix for https://github.com/blish-hud/Blish-HUD/issues/468
This disables falling back to other methods of process detection when a mumble link file name has been specified - This should prevent attaching to undesired processes.


dlamkins (approved):
Thanks!

PR #476, opened 2021-09-27
Fix for GetRawClient being called inside Update

eksime:
Fix the raw client being updated (and calling Regex.Match) in Update() by only updating the raw client when we attach to a new process.


dlamkins:
Seems like a perfectly reasonable solution!  I'll test it briefly tonight and then should be good to go.  Thank you so much!

dlamkins:
Works great!

dlamkins (approved):

Issue #470, opened 2021-09-26
Enum settings briefly set themselves to 0


dlamkins:
All settings briefly default to their types default value before becoming assigned.  This causes enum settings to fail if there isn't a value for 0.

REF: https://discord.com/channels/531175899588984842/599270434642460753/891707314896269322

dlamkins:
Fixed by #478

PR #478, opened 2021-09-28
Set default value in SettingView{T} ctor

eksime:
Potentially a fix for https://github.com/blish-hud/Blish-HUD/issues/470

This sets the Value property of the SettingView<T> to the specified default of the underlying setting when constructed. Currently the Value property is initialised to the default value for <T> which causes issues with enum dropdowns.


dlamkins (approved):

dlamkins:
Sure enough!  That looks like all that was needed.  Thank you!

Issue #407, opened 2021-07-31
.NET 5 update progress


Archomeda:
The update to .NET 5 has been a long journey so far. Development is mostly done in the feat/net5 branch.

- [x] Update TFM to net5.0
- [x] Update GitHub actions to build the .NET 5 version
- [x] Remove all Fody references from the repository (Fody is no longer needed as it's being replaced with single file publishing)
- [x] Use single file publishing
- [x] ~~Integrate DebugHelper code into the main Blish HUD project (and remove the separate executable build)~~ (done separately in #427)
- [x] Decide what to do with CSCore and if it's removed, update the code to no longer depend on it (CSCore doesn't have a .NET Standard/Core build, and while it technically should still work, it does still depend on .NET Framework, despite Blish HUD itself being fully .NET 5 instead)
- [ ] Check (and fix if necessary) what's causing the version to be included in the settings.json file, and causing the local VS build to throw an exception (and discard the settings) when it's loaded after using the CI build at least once (some settings.json overhaul is being done on the `net5-settings` branch, a.k.a. moving away from Newtonsoft.Json to System.Text.Json)
- [ ] Decide whether to use the portable build, or the self-contained build (and if trimming is an option, albeit still kinda experimental)
- [ ] Testing...
- [ ] Update module template to use .NET 5
- [ ] Maybe find a way to support modules made in .NET Framework 4.8 for backwards compatibility?
- [ ] Probably update to .NET 6

dlamkins:
Do we want to try and backport some of these changes into the framework version in preparation?  For example, embedding the debug hook stuff or including the setting changes?

Archomeda:
I *think* that's possible, unless I did something very .NET 5 dependent that I can't remember at the moment. We can try to backport the DebugHelper, CSCore (once that's done, or before this) and the settings JSON. Do you have similar stuff in mind?

eksime:
I had a look at swapping the CSCore-dependent code to use [NAudio.Wasapi](https://github.com/naudio/NAudio) - It's a fairly painless swap as they expose basically the same APIs. The code is at https://github.com/eksime/Blish-HUD/tree/feat/naudio if it's any use for task 6.

dlamkins:
Dang, that's pretty sweet!  Would definitely be a great way to get this process moving again since it was originally going to mean ripping out quite a bit and that was a bit tiring to think about.  😅

eksime:
[...]

PR #481, opened 2021-09-29
Swap implementation of CSCore to NAudio

eksime:
This should help with item 6 on https://github.com/blish-hud/Blish-HUD/issues/407 - NAudio seems to be more actively updated & supports targeting netstandard2.0. The api exposed by naudio is quite similar so it looks to be pretty much a drop-in replacement.


dlamkins:
Looks great!  Noting here that this is a breaking change that will require some modules to be recompiled, but there aren't many that should need that.

dlamkins (approved):

Issue #308, opened 2020-10-05
Modify Glide library to allow setting private members


dlamkins:
We utilize the small Glide animation library for tweening.  The library, as is, requires that the tweened properties be public members as the animation library uses reflection to get the setter method and compiles a lambda for updating the property every time an update is called.

While it is easy for us to allow the property to be visible the `NonPublic` flag, this does not work when the setter is compiled into the lambda as we're not calling `SetValue` directly on the private property but instead trying to do it in the compiled lambda.

It would be ideal for us to allow the values to be allowed as private to avoid exposing internal members meant for keeping animation states.

Changes would likely need to occur around https://github.com/blish-hud/Blish-HUD/blob/4f3894baae0e83c04c6e7c916057f299d5473229/Blish%20HUD/Library/Glide/MemberAccessor.cs#L47

Ideally, changes would allow the tweened members to be marked `private` and still allow us to retain the optimizations brought by caching the compiled lambdas (as these animations can occur on multiple objects simultaneously and have the potential to run every frame).

dlamkins:
Closed by #482!

PR #482, opened 2021-09-29
Allow setting private properties through Glide

eksime:
Fix for https://github.com/blish-hud/Blish-HUD/issues/308 - just need to pass `true` for [GetSetMethod(bool)](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.propertyinfo.getsetmethod?view=netframework-4.7.2)


dlamkins:
Tested and it works a charm.  Thanks!

dlamkins (approved):

PR #492, opened 2021-10-03
Rework mutex code to handle restarts 

### LINKED ISSUE #488: Blish HUD stops using a mutex if restarted.
To avoid multiple instances of Blish HUD, a mutex is set on launch.

To avoid issues when restarting, Blish HUD does not set this mutex when called from an existing instance to restart.  This was due to race conditions that would cause the new instance to instantly close because the calling instance still had the mutex open.

Instead of skipping the mutex on detected restarts, we should instead just release the mutex prior to calling the new instance.  That way we know it's released and don't have to worry about race conditions.

As it stands, some users have mistakenly managed to open multiple instances of Blish HUD if one instance restarted (thus not creating a mutex).  It's only happened a few times, but it is certainly confusing for those users.

eksime:
Should fix https://github.com/blish-hud/Blish-HUD/issues/488

Reworked the startup mutex code so that restartskipmutex is no longer required by adding a RestartOnExit static bool property to the Program class. When set, the program will restart on exit, after the mutex has been released.

Added handling for AbandonedMutexException & wrapped startup in a try ... finally block to ensure mutex is released on unhandled exceptions.

### DIFF
--- a/Blish HUD/ApplicationSettings.cs
+++ b/Blish HUD/ApplicationSettings.cs
@@ -46,8 +46,6 @@ namespace Blish_HUD {
          * r, ref       - The path to the ref.dat file.
          * s, settings  - The path where Blish HUD will save settings and other files.
          * w, window    - The name of the window to overlay.
-         *
-         * restartskipmutex - Forces Blish HUD to allow multiple instances.  Used internally to prevent race condition issues when Blish HUD is restarted.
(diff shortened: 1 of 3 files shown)


dlamkins (approved):

dlamkins:
Looks great!

Issue #475, opened 2021-09-27
ManagedConnection does not notify when token is renewed


dlamkins:
ManagedConnection does not indicate when it has successfully acquired a subtoken for a module to use.  As a result, any requests that are made right at launch will fail if they require going against an authenticated endpoint.

A status and possibly some kind of event needs added to facilitate this.

PR #494, opened 2021-10-05
Add subtoken event

agaertner:
Issue: #475


dlamkins:
Let's include a bool somewhere to indicate that the token is set, as well (in case we attempt to do something after the token is already made available).

agaertner:
> Let's include a bool somewhere to indicate that the token is set, as well (in case we attempt to do something after the token is already made available).

thought about that, but then I had the idea of creating new HasPermission methods which use the activePermissions that were used for successful creation of a subtoken. ~~Should be enough.~~ Should be better.

agaertner:
Done if no other requests coming up.

dlamkins (approved):

PR #495, opened 2021-10-05
Add support for custom settings renderers

### LINKED ISSUE #472: Support custom setting renderers
Module users may wish to provide their own custom setting render for a given type.  There should be a way to do this either through our compliance system or a more global way of registering additional type constructors in the SettingsView class.

eksime:
Potentially an implementation for https://github.com/blish-hud/Blish-HUD/issues/472 that adds `ISettingsViewFactory` and `ISettingsViewFactorySelector` interfaces and associated classes to replace functionality in the `SettingView` static class, allowing people to define a view for a specific setting, or override the views created by a `SettingsView` from any `SettingCollection` or its children.

I've attempted to make sure the behaviour of this system is, by default, identical to the current system. The  `SettingView` gets marked obsolete, and it's implementation is swapped out to use this system, in case anyone was using the static methods on that class directly.

There's probably many things that could be done differently with this and it'd be great to have feedback / hear people's thoughts on any of this, but particularly:

- I'm not sure if having the ViewFactory class (& descendants) is necessary, as we could potentially just use delegates for this. It does give us an interface to work with in the future, but it makes coding against it quite boilerplate-y though.
- There may be a better way to implement the defaults for a specific `SettingEntry<T>` - I've tried to keep everything regarding defaults static, and to try and re-use objects where possible rather than constantly new'ing them, and follow (roughly) how the BCL handles things like `Comparer<T>.Default`
- I'm not sure if we need `ISettingsViewFactorySelector` or whether that logic be left to the `SettingEntry<SettingCollection>` view implementation; and if so, how would someone provide an alternate view for the provided `SettingsManager.ModuleSettings` SettingCollection?

Thanks!


dlamkins:
Hey, this is looking great!  Could you give just some examples of how it would look like to define a setting with a custom setting rendering?

It appears you'd build out your view and then create a factory to then pass along in the define call.  Is that all there would be to it?

> I'm not sure if we need ISettingsViewFactorySelector or whether that logic be left to the SettingEntry<SettingCollection> view implementation; and if so, how would someone provide an alternate view for the provided SettingsManager.ModuleSettings SettingCollection?

_Maybe?_  Regarding module settings, modules can also return their own view in more recent versions, which would allow them to skip some of that, but would definitely end up resulting in some redundant stuff that probably should be handled properly in Core.  🤔

I'll begin looking over things more during my lunch break tomorrow.  🙂

eksime:
> Hey, this is looking great! Could you give just some examples of how it would look like to define a setting with a custom setting rendering?
> It appears you'd build out your view and then create a factory to then pass along in the define call. Is that all there would be to it?

Yeah, that's how it works atm - There's 3 ways to override a setting:

1. Pass a delegate to `SettingCollection.DefineSetting<T>` (where `SomeView : IView`):
```
settings.DefineSetting("SomeBoolSetting", false, () => "Some Bool", () => "An example setting", (s, w) => new SomeView(s, w));
```
2. Pass an `ISettingViewFactory<T>` implementation to `SettingCollection.DefineSetting<T>`:
```
public class ExampleViewFactory : ISettingViewFactory<bool> {
    public IView CreateView(SettingEntry<bool> setting, int definedWidth) {
        return new SomeView(s, w);
    }

    public IView CreateView(SettingEntry setting, int definedWidth) {
        return this.CreateView((SettingEntry<bool>)setting, definedWidth);
    }
}
//...
ISettingViewFactory<bool> boolViewFactory = new ExampleViewFactory();
settings.DefineSetting("SomeBoolSetting", false, () => "Some Bool", () => "An example setting", boolViewFactory);
```
3. Provide a new implementation of `ISettingViewFactorySelector` to the `SettingCollection` instance:
```
public class ExampleViewFactorySelector : ISettingViewFactorySelector {
    private static readonly ISettingViewFactory boolfactory = new ExampleViewFactory();

    public ISettingViewFactory<T> GetFactoryForType<T>() {
        return GetFactoryForType(typeof(T)) as ISettingViewFactory<T>;
    }

    public ISettingViewFactory GetFactoryForType(Type t) {
        if (t == typeof(bool)) {
            return boolfactory;
        }

        return null;
    }
}
//...
settings.ViewFactorySelector = new ExampleViewFactorySelector();
settings.DefineSetting("SomeBoolSetting", false, () => "Some Bool", () => "An example setting");
```
In the last example, no factory is provided to the `settings.DefineSetting(...` call, the SettingsView class uses the ISettingViewFactorySelector to retrieve an appropriate ViewFactory for the bool setting. Any child SettingCollections/SettingsViews will also use the same IViewFactorySelector.

> _Maybe?_ Regarding module settings, modules can also return their own view in more recent versions, which would allow them to skip some of that, but would definitely end up resulting in some redundant stuff that probably should be handled properly in Core. 🤔

I was thinking about this a bit as I was writing this - I don't think ISettingViewFactorySelector is really needed at all - we only really need `ISettingViewFactory`/`ISettingViewFactory<T>`. A "selector"-type factory could just implement multiple `ISettingViewFactory<T>` interfaces, with the selecting part handled by `ISettingVewFactory.CreateView(SettingEntry setting, int definedWidth)` that could examine SettingEntry's `SettingType` property and switch as appropriate.

edit: 
Ah, looking into this more, *not* having the concept of a "selector" makes working with enum settings a lot more difficult, unless we don't have a `ISettingViewFactory<T>` and just have the non-generic type.

dlamkins:
Ahh yeah, enum settings have more than once added some complications to our setting handling. 🤔

dlamkins:
Okay, after some testing, only thing that appears to be missing is support for the default keybinding setting view.

image

As far as I can tell, most everything else looks really great and I haven't found any issues with what you've made backwards compatible, which is great.  👍

eksime:
Sorry, I've not had much time in the last few weeks to work on this - the keybinding setting view *was* working at some point, so I'll take a look at that! I was also going to look at the implementation some more to see if there's any possible way to make this cleaner, it's still a bit messy imo

dlamkins:
Understood and no rush!  I really appreciate the work done here.  👍

PR #496, opened 2021-10-06
Fix subtoken requests happening before privileged connection has key

agaertner:
The call of ``RenewSubtoken`` in the constructor of the Gw2ApiManager will cause a subtoken request on each module load before the main key is assigned to the PrivilegedConnection.
That call was probably put there for when you enable a module in an already ongoing session with a privileged connection that already owns a main key.

This resolves the issue for faulting subtoken requests happening at the launch of the program where Gw2ApiManagers are created before the privileged connection is assigned an api key.

(The validity of the api key is already checked elsewhere so a simple non-null-non-empty key check suffices.)

### DIFF
--- a/Blish HUD/GameServices/Gw2WebApi/ManagedConnection.cs
+++ b/Blish HUD/GameServices/Gw2WebApi/ManagedConnection.cs
@@ -53,5 +53,8 @@ namespace Blish_HUD.Gw2WebApi {
             return true;
         }
 
+        public bool HasApiKey() {
+            return !string.IsNullOrEmpty(_internalConnection.AccessToken);
+        }
     }
 }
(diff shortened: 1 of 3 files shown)


dlamkins (changes requested):
Instead of adding `HasApiKey`, lets just add `RequestPrivilegedSubtoken` to the `ManagedConnection` class.  Not having a token at all should be identical to just not having any permissions.

agaertner:
Moved the method. 

> Not having a token at all should be identical to just not having any permissions.

it works exactly like this. If the connection does not have a key to create a subtoken with an empty string is returned which is handled in the ApiManager to just return and use AnonymousConnection by default.

agaertner:
Reverted back to first commit.

As I understand you we only want or need to avoid the first ``InvalidAccessException`` to avoid a stacktrace in the logs that is just saying "hey, the privileged connection has no key yet"

Any other time after BlishHUD is launched and RequestPrivilegedSubtoken is called we do want to handle that exception and log it appropriately.

agaertner:
~~Currently has a bug inwhich _activePermission and _expiresAt are reset due to a new Gw2ApiManager being created with a call of GetModuleInstance when new permissions are selected.~~

Works as expected. Any HasPermission call before subtoken is requested obviously will return false.

PR #500, opened 2021-10-08
Improve Gw2WebApi service relationship with modules

dlamkins:
- Makes Gw2WebApi calls async friendly.
- Improves logging output for Gw2WebApi service.
- Fixes a bug that could accidentally escalate the privileges on the shared anonymous connection.
- Stopped creating so many ManagedConnections and instead just keep using the existing one.
- Subtokens are requested before modules are told to load so the API _should_ be ready when the module loads (assuming a token could be acquired).
- When a token is removed, update all module subtokens to reflect this if they were using the removed token.
- Stopped requesting subtokens when we know the privileged connection doesn't have a key to use yet.

@agaertner Do you feel this covers everything left that was needed?  If so, we can likely close #496.

### DIFF
--- a/Blish HUD/GameServices/Gw2WebApi/ManagedConnection.cs
+++ b/Blish HUD/GameServices/Gw2WebApi/ManagedConnection.cs
@@ -42,7 +42,7 @@ namespace Blish_HUD.Gw2WebApi {
         private void UserLocaleOnSettingChanged(object sender, ValueChangedEventArgs<Locale> e) {
             _internalConnection.Locale = e.NewValue;
 
-            Logger.Debug($"{nameof(ManagedConnection)} updated locale to {e.NewValue} (was {e.PrevousValue}).");
+            Logger.Debug($"{nameof(ManagedConnection)} updated locale to {e.NewValue} (was {e.PreviousValue}).");
         }
 
         public bool SetApiKey(string apiKey) {
--- a/Blish HUD/GameServices/Gw2WebApi/ManagedConnection.cs
+++ b/Blish HUD/GameServices/Gw2WebApi/ManagedConnection.cs
@@ -50,6 +50,10 @@ namespace Blish_HUD.Gw2WebApi {
 
             _internalConnection.AccessToken = apiKey;
 
+            Logger.Debug(apiKey == string.Empty
+                             ? $"{_internalConnection.UserAgent} cleared API token."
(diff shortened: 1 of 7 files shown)


agaertner:
You could have just replied to me..
 Yeah I guess you can close it and merge this now. 
I don't see anything faulty except maybe if you wanted to keep subtokens persistent someday you may want to check against the expiration claim.

dlamkins:
> You could have just replied to me.. Yeah I guess you can close it and merge this now. I don't see anything faulty except maybe if you wanted to keep subtokens persistent someday you may want to check against the expiration claim.

Sorry, I was updating the api service after I noticed some old bugs (unregister not refreshing the keys for modules and us mistakenly giving the anonymous connection a token) and while making things async I noticed where I wanted those changes to go.  Sorry for not messaging you and then forgoing your PR. 🙁

That's a good point about the expiration claim.  I will add a TODO to review that.

agaertner:
Note that subtokens get invalidated if the parent key is deleted. In this case ``_activePermissions`` 
 still has entries and ``HasPermissions`` will return true regardless.
I guess this case falls into failed request task results that have to be caught by the module developer.

dlamkins:
That's a pretty great catch.  Let me PR a fix to that so that we don't leave that on the modules to deal with.  Should be a very rare scenario anyways, though (hopefully).

agaertner:
Should be fine if deletion of key in the UI will result in a subtoken renewal and call of RenewSubtoken.

Not sure, but if a change in parent key triggers a call of the GetModuleInstance method and it always creates a new Gw2ApiManager the constructor creates a new empty ``_activePermissions`` hashset already.

I was also talking about deleting the parent key from the account webpage, but I guess we cannot handle that and module devs will have to check for task fault.

dlamkins:
> Should be fine if deletion of key in the UI will result in a subtoken renewal and call of RenewSubtoken.

Yep, it does now.  🙂 

> Not sure, but if a change in parent key triggers a call of the GetModuleInstance method and it always creates a new

Nah, it sets the API key on the managed connections that they have instances of.  Won't create new ones.  👍

> I was also talking about deleting the parent key from the account webpage, but I guess we cannot handle that and module devs will have to check for task fault.

I will test what happens.  I think that the token will fail to be used because it won't be associated with the user (we query the users character names on launch, so if the token doesn't work then modules will never get anything) and modules won't have to worry about it.  They just won't have permissions listed.

PR #502, opened 2021-10-08
Refactor/Move intern and extern keyboard & mouse input

agaertner:
This was a long time due.

I merged the enums from Extern/ and reduced the number of files.


dlamkins:
Some good changes, probably needs some further discussion.  Input service items probably should be dedicated to just the input.  This likely would be under integration.

Namespace changes would be a breaking change, though, so we'll need to hold this for a bit since there are a lot of modules I think using this at the moment.

agaertner:
One thing to consider is merging the static classes ``Mouse.cs`` and ``Keyboard.cs`` **into** ``MouseHandler.cs`` and ``KeyboardHandler.cs`` respectively which avoids confusion when module developers access ``GameService.Input.Mouse`` or ``GameService.Input.Keyboard``

I just wanted to get your feedback before I merge them in.

dlamkins:
I think sending key or mouse input should not be kept in the InputService namespace.  The input handling is complicated and sensitive and these functions are completely unrelated in terms of what they actually do.  🤔

agaertner:
> I think sending key or mouse input should not be kept in the InputService namespace. The input handling is complicated and sensitive and these functions are completely unrelated in terms of what they actually do. 🤔

I had that thought, but where else would they look better? I feel like Input fits best and I made sure to merge the files into just the Keyboard class and Mouse class.      

 It would make sense for module devs to access them via ``GameService.Input.Mouse`` if I also were to merge the static class into the MouseHandler.

If you have an idea where they are better off I am open to ideas.

agaertner:
One idea would be to rename them into a ``KeyboardUtil`` and ``MouseUtil`` under ``_Util``? They are both static classes anyway.
But then there is still the PInvoke stuff that has to be somewhere. Maybe I can merge everything into just those two util classes then. The only problem would be VirtualKeyShort and ScanCodeShort. I guess those could go in the root of Common.

dlamkins:
FYI, I'm only holding off on merging for now as it's a big breaking change, so I'm saving it for a later version where we can merge in a couple of these sorts of changes.

I'll probably look at doing dev & version branches again in the future to help avoid this issue for future versions.

dlamkins:
Closing this for now until we're ready to do a release with plenty of breaking changes in it.  Probably won't happen until we introduce the SDK anyways so closing for now with the intention that we can re-open it further down the road.

PR #507, opened 2021-10-14
Embed PDB in the exe.

dlamkins:
Embed the PDB into the main exe to reduce our footprint in the destination folder (and ensure that when we provide alt exes for users that the PDB is included all just with the one exe).


dlamkins:
Completely wrong branch somehow.  🤦‍♂️

PR #509, opened 2021-10-14
Color picker - Remove Forced Tint

manlaan:
Replaced forced tint on background of ColorPicker to allow the dev to select Tint, BackgroundColor, or BackgroundTexture, or none (if placing the ColorPicker on a panel, for instance).  

Set the default value ShowTint = true, to closely reproduce the previous setting.  (Panel default of 0.4 instead of previous value 0.5).  This might want to be removed all together to not confuse the dev though.

### DIFF
--- a/Blish HUD/Controls/ColorPicker.cs
+++ b/Blish HUD/Controls/ColorPicker.cs
@@ -61,6 +61,7 @@ namespace Blish_HUD.Controls {
         }
 
         public ColorPicker() : base() {
+            this.ShowTint = true;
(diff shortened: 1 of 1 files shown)


dlamkins:
To avoid the UI change we'll want to keep it as true as you have done.  Best would be to update the documentation to indicate the option of removing the background tint.  I suspect 0.4 will be close enough to the 0.5 it was before to be fine.

dlamkins (approved):

PR #510, opened 2021-10-15
Delete modules from UI

dlamkins:
This PR primarily just extracts some of the module pkg management from the pkgrepo presenter into the repo handler and splits out the delete module functionality so that it can be called from the UI.

### DIFF
--- a/Blish HUD/Controls/MenuItem.cs
+++ b/Blish HUD/Controls/MenuItem.cs
@@ -191,6 +191,8 @@ namespace Blish_HUD.Controls {
         public void Select() {
             if (this.Selected) return;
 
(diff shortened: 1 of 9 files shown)


kazerniel:
So how can I uninstall modules? (either from the UI or in another way)

dlamkins:
@kazerniel you can delete modules by clicking the cog wheel on the module's page and selecting the menu item to delete the module. 

You can also check the manual module install instructions on our website. You can just delete the bhm files from the modules folder.

kazerniel:
Oh my bad, I was using an older version of Blish HUD that didn't have such an option 😅 No wonder I couldn't find it anywhere 😅

PR #512, opened 2021-10-15
add "HasSubtoken"

agaertner:
While coding I noticed you were right. This is still good to have.  This way I just need to check it and not make a delegate for the event and another property.


dlamkins:
For what it's worth, after the recent changes, modules will load only after their subtoken has been acquired, so you probably shouldn't have to listen on the event like you mentioned.

This is probably still good to have, though.

dlamkins (approved):

PR #523, opened 2021-10-20
Fix manually specified mumblelink

### LINKED ISSUE #522: Regression: mumblelink name not used
After a recent update, it appears that launch arg mumblelink names are not properly used to connect to the game.

dlamkins:
- Updated command line search pattern to make double quotes optional since it doesn't appear to be included always.
- Ensure that we refresh the raw client anytime we detect a new process to help ensure we can auto detect, still avoid falling back when manually specified, and fallback when not manually specified.

Fixes #522


eksime (approved):
Looks good! - The CurrentMumbleMapName property approach makes the AboutView much neater too

dlamkins:
Thanks for checking!  🙂

Issue #536, opened 2021-10-29
Only load the latest version of the module


dlamkins:
If multiple copies of the same module are found in the modules directory, we should only load the one with the highest version number.  Currently we will load and potentially even run all copies.

dlamkins:
Closed by #537

PR #537, opened 2021-11-03
Only load one instance of each module, even if multiple are available.

dlamkins:
Also catches malformed manifests and avoids moving forward with registering the module.


dlamkins:
Closes #536

Issue #504, opened 2021-10-11
Hide in taskbar when game is closed


dlamkins:
If show in taskbar is enabled and the overlay is also set to minimize to the tray area when the game is closed, it will still show on the taskbar.

Despite show in taskbar being enabled, we should hide ourselves from there when the game is not running.

entrhopi:
This is resolved and can be closed.

PR #540, opened 2021-11-09
Sync taskbar icon with game running

entrhopi:
Proposal for #504
It has quite a significant delay since `Gw2InstanceIntegration` is not very fast hooking onto the game


entrhopi:
Minimizing of the game is not handled yet.

dlamkins:
This will be very great.  😄

Oh, nice.  Is there any issue calling `SetShowInTaskbar` to the same value?  That is to say, can this be made more stateless by removing that `_inTaskbar` var?

Something else to consider is that we now also restart Blish HUD when the game is closed, so likely at most we'd only ever need to toggle the taskbar item on, once, when the game is first detected since the moment the game is no longer running, we just restart Blish HUD and can default to it not being visible again until the game is detected.  🤔

entrhopi:
> Oh, nice. Is there any issue calling `SetShowInTaskbar` to the same value? That is to say, can this be made more stateless by removing that `_inTaskbar` var?
> 
> Something else to consider is that we now also restart Blish HUD when the game is closed, so likely at most we'd only ever need to toggle the taskbar item on, once, when the game is first detected since the moment the game is no longer running, we just restart Blish HUD and can default to it not being visible again until the game is detected. 🤔

Ah nice, I was wondering about that. Thanks for the input!
In that case we should in fact be able to skip `_inTaskbar`, I'll refactor this and run some tests.

dlamkins:
Great!

dlamkins (approved):

Issue #442, opened 2021-09-11
Setting order is not maintained


dlamkins:
Currently the setting order is based on their order in the settings.json file instead of the order they are defined which can result in strange ordering.

The order of settings should be the order that they are defined.

PR #543, opened 2021-11-13
Setting order is now maintained

dlamkins:
Settings are now more thread-safe and when enumerated will be listed in the order they were defined and then any settings that have not been defined again will be listed afterwards in the order they are listed in the settings.json.


dlamkins:
Closes #442

Issue #229, opened 2020-05-21
Option to manually set UI Size and ignore in-game UI size


BlishHUD:
**Naillyk at 5/21/2020 9:50:38 AM +00:00**
> > To clarify, are you looking for a way to set the UI size of Blish HUD manually?  
> <@!164832069401182210> 
> 
> You couldn't be more spot on than that! 😄

REF: https://discordapp.com/channels/531175899588984842/531175900100952087/712965569674215485

entrhopi:
This can be closed.

dlamkins:
Closed by #547

PR #547, opened 2021-11-21
Allow manual setting of UI size

entrhopi:
Update to address #229

### DIFF
--- a/Blish HUD/GameServices/Graphics/ManualUISize.cs
+++ b/Blish HUD/GameServices/Graphics/ManualUISize.cs
@@ -1,0 +1,20 @@
+﻿using System.ComponentModel;
+
+namespace Blish_HUD.Graphics {
+    public enum ManualUISize : int {
+        [Description("Automatic - Sync With Game")]
+        SyncWithGame = 0,
+
+        [Description("Small")]
+        Small = 1,
+
+        [Description("Normal")]
+        Normal = 2,
+
+        [Description("Large")]
+        Large = 3,
+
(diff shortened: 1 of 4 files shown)


Blish HUD/GameServices/Graphics/ManualUISize.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System.ComponentModel;
+
+namespace Blish_HUD.Graphics {
+    public enum ManualUISize : int {
+        [Description("Automatic - Sync With Game")]
```
dlamkins, line 5: For the sake of consistency, let's just have this read `Sync With Game`.

dlamkins (changes requested):
Let's get the one name change and the rest looks good!

Blish HUD/GameServices/Graphics/ManualUISize.cs
entrhopi: I have copied this from `DPIMethod` should we change it there too then?

entrhopi: Upon further inspection I'd suggest changing the name of the enum to `UISizeMethod` too?

dlamkins: I probably am just misremembering things.  I'll check it tonight.  It's probably all fine to leave as is if you copied it over.  :D

dlamkins (approved):

PR #548, opened 2021-11-23
feat/dynamic-hud

entrhopi:
https://discord.com/channels/531175899588984842/536970543736291346/912567991017816125

### DIFF
--- a/Blish HUD/GameServices/Overlay/DynamicHUDMethod.cs
+++ b/Blish HUD/GameServices/Overlay/DynamicHUDMethod.cs
@@ -1,0 +1,17 @@
+﻿using System.ComponentModel;
+
+namespace Blish_HUD.Graphics {
+    public enum DynamicHUDMethod : int {
+        [Description("Always Show (Default)")]
+        AlwaysShow = 0,
+
+        [Description("Show Only out of Combat")]
+        ShowPeaceful = 1,
+
+        [Description("Show Only in Combat")]
+        ShowInCombat = 2,
+
+        [Description("Never Show")]
+        NeverShow = 3
+    }
+}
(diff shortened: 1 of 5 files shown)


Blish HUD/GameServices/Overlay/DynamicHUDMethod.cs
```diff
@@ -0,0 +1,17 @@
...
+namespace Blish_HUD.Graphics {
+    public enum DynamicHUDMethod : int {
+        [Description("Always Show (Default)")]
+        AlwaysShow = 0,
+
+        [Description("Show Only out of Combat")]
+        ShowPeaceful = 1,
+
+        [Description("Show Only in Combat")]
+        ShowInCombat = 2,
```
dlamkins, line 12: Could we perhaps do `SetExcluded` for the current setting and this enum value?  I like the idea of having this as an option for other items, but having just gotten myself stuck with it set like this and having to go back into combat to change the setting back was kind of a weird experience.

I'm hoping it's safe to assume that there aren't too many people out there that would want this for the corner icons (but again, good to keep as we add support for other UI elements).

Blish HUD/GameServices/Overlay/DynamicHUDMethod.cs
```diff
@@ -0,0 +1,17 @@
...
+
+        [Description("Never Show")]
+        NeverShow = 3
```
dlamkins, line 15: Same goes with `NeverShow` for the corner icons.  It's a quick way to get soft locked without an easy way to fix the setting.

dlamkins:
I noticed that it doesn't appear to update the actual state of the corner icons until a combat event occurs again making it appear to do nothing until the next time you're in combat and it "pops" into the correct state and is correct from that point forward.

dlamkins (changes requested):

dlamkins (approved):

PR #549, opened 2021-11-24
Update active control on keyboard event not only on mouse event

### LINKED ISSUE #23: Control.BasicTooltipText will not disappear
Control.BasicTooltipText will not disappear when BlishHUD window is being closed while the tooltip is open.

Example:
A button shows BasicTooltipText on mouse hover.
On click this button closes the main BlishHUD window.
The BasicToolTipText will not disappear.

### LINKED ISSUE #567: Textbox remains selected when a window is closed with ESC.
We have two options for the behavior here.
- Both the textbox loses focus and the window is closed (this seems a bit more complicated).
- Just the textbox loses focus.

For anything that doesn't close with ESC, the latter would be the observed behavior, so for consistency and for simplicity, it's probably best that we just ensure that the windows don't close if a text field is selected (which should be easy enough).

entrhopi:
Quick and dirty draft to fix #23

`ActiveControl` is now settable and will be set `null` if a parent in the ancestry is hidden.


dlamkins:
I'll have to do some performance profiling here since it's in a potential hot path, but once done and hopefully good will hopefully get merged.  🙂

entrhopi:
I haven't encountered any issues with the solution so far.
But my machine is not really low end, so I think some profiling would be nice.

dlamkins (approved):
Okay, performance seems to be just fine on this and I like how you implemented `GetAncestors()`.

I think the ActiveControl stuff as a whole could likely be cleaned up as I think you've mentioned to me elsewhere, but I think that will likely be a far more involved process.

Was there anything you were wanting to do more before we merge (since it's still marked as a draft)?

entrhopi:
Yes!
I was recently experimenting with hooking only to defined hotkeys so we do not check every single keystroke.
And I wanted to implement disabling the active control when one of its ancestors is detected to be invisible.

Since this is a long standing (non intruding) issue I don't see a problem with tackling `ActiveControl` cleanup here too.

dlamkins:
Did you want to still put those changes in this PR or perhaps those can go into another?  🙂

entrhopi:
No, I guess I'll open another PR for the cleanup stuff.
But I noticed that my changes do not fix #567. Looking at this right now.

entrhopi:
This fixes `ESC` and other mechanics that hide ancestor controls. It is unfortuanetly still a keypress too late most of the times.

Issue #511, opened 2021-10-15
Add option to open Blish HUD settings folder (or logs directly).


dlamkins:
Add the option to open either the logs folder or the settings dir from within the Blish HUD UI.  Places it could go:
- In the right-click menu when right-clicking on the Blish HUD icon in the top left.
- In the right-click menu when right-clicking on the Blish HUD icon in the tray area (meaning it can be opened even when the game is not running).
- In the about view.

manlaan:
I vote for the "About View" and the system tray icon, as both of those are pretty common places for people to go for further diag help.   The in game icon is a bit too used and would get in the way for something most wont need.

dlamkins:
> The in game icon is a bit too used and would get in the way for something most wont need.

For what it's worth, the in-game icon currently is only used for Exit and Restarting so certainly not used for too much.  🤔

But gotcha.  Thanks for the feedback!

entrhopi:
This can be closed.

dlamkins:
Sure can!  Closed by #553.

PR #553, opened 2021-12-11
Added 'open settings' item to tray menu

entrhopi:
Proposal for #511


dlamkins:
Love it!  I think this adequately resolves and closes #551.

PR #555, opened 2021-12-13
Let users enable debug logging from the UI without overlay

### LINKED ISSUE #451: Enable debug logging without --debug flag
Enable more verbose logging options with an option in the UI that does not require showing the debug info in the corner.

entrhopi:
Proposal for #451


Blish HUD/Strings/GameServices/OverlayService.resx
```diff
@@ -223,4 +223,10 @@ It might not be installed.</value>
   <data name="Setting_DynamicHUDMenuBar_DisplayName" xml:space="preserve">
     <value>Dynamic HUD Menu Bar</value>
   </data>
+  <data name="Setting_DebugLogging_Description" xml:space="preserve">
+    <value>If enabled Blish HUD will create more verbose logfiles</value>
+  </data>
+  <data name="Setting_DebugLogging_DisplayName" xml:space="preserve">
+    <value>Enable debug logging</value>
```
dlamkins: Let's revise this to `Enable Debug Logging` as the rest of our settings use title caps.

dlamkins (changes requested):

dlamkins:
It feels like we might need something more for this, but I'm not exactly sure what.  Like perhaps making it a dropdown with the option to show the FPS counter stuff in the corner or not.

I think there are generally three scenarios we're trying to cater to:
1. Users that want to enable debug to see FPS or the timings in the top-right corner.
2. Users that we want to see debug details in their logs.
3. Users that want to use the debug input hook method but that don't want to see the corner FPS/timings and don't need debug logs.

dlamkins:
What's the expected behavior when a user passes the `--debug` flag?  Seems likely we'd want to mark the settings all as disabled when this happens?

entrhopi:
Currently it ignores and overrides all user settings. We can, in that case, either disable the inputs or add a note to description.
What would you prefer?

dlamkins:
> Currently it ignores and overrides all user settings. We can, in that case, either disable the inputs or add a note to description. What would you prefer?

Check out what we did for the FPS launch arg.

https://github.com/blish-hud/Blish-HUD/blob/fdea3473ff2791f2fdf9da9b6ac1929848563d7e/Blish%20HUD/GameServices/GraphicsService.cs#L292-L298

We mark the setting as disabled and then change the tooltip/setting description to one that indicates that the setting is disabled because it was manually overridden by the launch arg.

---

Was the changes to game service order updated before the file based method was decided on?  That is to say, is it still necessary to have the settings service load prior to the debug service?

entrhopi:
Yes. Otherwise it is not easily possible to get the user settings at launch and Blish will crash immediately. (unless we're writing a separate parser for the settings file just for the debug service)

dlamkins:
Okay, that makes sense.  I forgot that both the file and standard settings were being used to facilitate tracking those settings.  That makes sense.  👍

Blish HUD/Strings/GameServices/DebugService.resx
```diff
@@ -0,0 +1,144 @@
...
+    <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <data name="DebugSettingsSection" xml:space="preserve">
+    <value>Debug Settings</value>
+  </data>
+  <data name="Setting_AdditionalDebugDisplay_Description" xml:space="preserve">
+    <value>If enabled additional debug information will be displayed</value>
+  </data>
+  <data name="Setting_AdditionalDebugDisplay_DisplayName" xml:space="preserve">
+    <value>Show Additional Debug Settings</value>
```
dlamkins: Let's change this to "Show Additional Debug Data" or something since the current name makes it sound like more debug _settings_ will become visible when this is enabled when in reality there are no more settings to show.  👍

dlamkins (changes requested):

dlamkins:
This looks great!  Looks like just a single conflict, but once above change is completed and conflict resolved we can merge for the next prerelease!  🙂

entrhopi:
Here you go! xD

dlamkins (approved):

Issue #484, opened 2021-09-30
Trackbar jumps to wrong position when clicked.


dlamkins:
https://i.gyazo.com/c185f613865f1ca38e36b02cbefc7044.gif
https://i.gyazo.com/bfde64ab95e0c0f4844d533fc246cba7.gif

The trackbar jumps when clicked on.  Appears to be made worse when the starting offset is non-zero.  Replicate with the setting:
![image](https://user-images.githubusercontent.com/1950594/135539317-6b65c6f5-61f1-4793-b3eb-d21cee45e010.png)

REF: https://discord.com/channels/531175899588984842/599270434642460753/892811075966349353

entrhopi:
I tried to mitigate this by subtracting 29 from `_dragOffset` in https://github.com/blish-hud/Blish-HUD/blob/dev/Blish%20HUD/Controls/TrackBar.cs#L96 like so:
`var relMousePos = this.RelativeMousePosition - new Point(_dragOffset - 29, 0);`

This seems to lessen the effect but the closer you get to the left or right of the bar there will be a tiny drift to either side.

I will investigate this further.

entrhopi:
Found part of the issue:
If a MinValue other than 0 is specified, then this would be the `wrong` slider offset as seen in the examples above.

Working on a fix now.

dlamkins:
Closed by #556!

PR #556, opened 2021-12-14
Fixed TrackBar value generation/tracking #484

entrhopi:
Fix for #484

### DIFF
--- a/Blish HUD/Controls/TrackBar.cs
+++ b/Blish HUD/Controls/TrackBar.cs
@@ -70,7 +70,6 @@ namespace Blish_HUD.Controls {
         }
 
         private bool _dragging   = false;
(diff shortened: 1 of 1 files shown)


dlamkins:
Okay, this already feels a ton better than it did before.

Only issue I notice remaining is that without the drag offset being traced, the nub always snaps with the offset shown below:

image

The drag offset, if I remember correctly was to help remember how off we were offset on the nub to remember that as we dragged it.  Regardless, the snapping appears to be very little so this is quite nice.

dlamkins:
I'll check it out tonight!

dlamkins (approved):
Buttery smooth!  This is great and I love the new 10% increments - they feel great.  👍

Issue #550, opened 2021-11-26
Log count isn't properly enforced 


dlamkins:
https://github.com/blish-hud/Blish-HUD/blob/dev/Blish HUD/GameServices/DebugService.cs#L56

The number of logs can exceed this count. Due to our date time based log files, this feature requires further configuration to work as expected.

dlamkins:
Closed by #557!

PR #557, opened 2021-12-14
Fix for logfile cleanup

entrhopi:
Fix for #550

nLog does not support fancy filenames, so we have to enroll our own cleanup routine.

### DIFF
--- a/Blish HUD/GameServices/DebugService.cs
+++ b/Blish HUD/GameServices/DebugService.cs
@@ -49,10 +49,6 @@ namespace Blish_HUD {
                 FileNameKind      = FilePathKind.Absolute,
                 ArchiveFileKind   = FilePathKind.Absolute,
                 FileName          = Path.Combine(logPath, "blishhud.${cached:${date:format=yyyyMMdd-HHmmss}}.log"),
-                ArchiveFileName   = Path.Combine(logPath, "blishhud.${cached:${date:format=yyyyMMdd-HHmmss}}.{#}.log"),
-                ArchiveDateFormat = "yyyyMMdd-HHmmss",
-                ArchiveAboveSize  = MAX_LOG_SIZE,
-                ArchiveNumbering  = ArchiveNumberingMode.Sequence,
                 MaxArchiveFiles   = MAX_LOG_SESSIONS,
                 EnableFileDelete  = true,
                 CreateDirs        = true,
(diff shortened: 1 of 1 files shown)


snakefoot:
If you remove these settings, then NLog cleanup will work just fine when using NLog 4.5 (or newer):

```c#
               ArchiveFileName   = Path.Combine(logPath, "blishhud.${cached:${date:format=yyyyMMdd-HHmmss}}.{#}.log"),
               ArchiveDateFormat = "yyyyMMdd-HHmmss",
               ArchiveNumbering = ArchiveNumberingMode.Sequence,
```

See also: https://github.com/NLog/NLog/wiki/File-target#archive-old-log-files

P.S. The reason why your current attempt is not working is because you are mixing [static- with dynamic](https://github.com/NLog/NLog/wiki/File-target#dynamic-vs-static-archive-logic) archive-logic.

entrhopi:
Nice! Thank you for the nudge in the right direction @snakefoot !

Blish HUD/GameServices/DebugService.cs
```diff
@@ -49,10 +49,6 @@ public class DebugService : GameService {
                 FileNameKind      = FilePathKind.Absolute,
                 ArchiveFileKind   = FilePathKind.Absolute,
                 FileName          = Path.Combine(logPath, "blishhud.${cached:${date:format=yyyyMMdd-HHmmss}}.log"),
-                ArchiveFileName   = Path.Combine(logPath, "blishhud.${cached:${date:format=yyyyMMdd-HHmmss}}.{#}.log"),
-                ArchiveDateFormat = "yyyyMMdd-HHmmss",
-                ArchiveAboveSize  = MAX_LOG_SIZE,
```
snakefoot, line 54: You probably want to keep `ArchiveAboveSize = MAX_LOG_SIZE,`

entrhopi: Hmmm ... but then we should use not only `Date` but `DateAndSequence` in the filename?
Aka what would happen if we surpass the file size?

entrhopi: Would NLog simply create a new file with the current time? That way it would be very unlikely that we surpass 1MB in less than a second.

snakefoot: NLog will by default archive using sequence-no (unless something else has been specified):

```
blishhud.20010101-010101.log <- Active
blishhud.20010101-010101.0.log
blishhud.20010101-010101.1.log
blishhud.20010101-010101.2.log
```

Notice the use of `${cached}` means the filename stays the same for the entire application-lifetime (Similar to `${processinfo:StartTime:format=yyyyMMdd-HHmmss:cached=true}`)

entrhopi: Nice!
Thank you very much for your help! Much appreciated!

dlamkins:
Looks great and thanks for the surprise assist @snakefoot!

dlamkins (approved):

PR #565, opened 2022-01-04
Fix #563

### LINKED ISSUE #563: Large memory allocations when Blish HUD regains focus.
When Blish HUD gains focus, large amounts of memory are allocated for a short moment.  It appears that on systems where GC is struggling (or some other undetermined cause) that this memory usage quickly stacks up without releasing.

eksime:
Hopefully fixes #563 - or is at least a band-aid for it. The allocation was occurring inside the `GraphicsDeviceManager.ApplyChanges()` call in: https://github.com/blish-hud/Blish-HUD/blob/87d19ea57cf63f98c9bd0f03db698c3727dcf4f5/Blish%20HUD/GameServices/GraphicsService.cs#L198-L214

`ApplyChanges` was called each time the resolution property was set, which was each time the window gained focus - this change should prevent it from being called when the Resolution is unchanged on re-focus.


dlamkins (approved):
Thanks again!

Issue #569, opened 2022-01-07
Blish HUD main icons stops working


Timon76:
I am one of them that suffers from the strange jumping mouse cursor jumping in GW2 (this is a known bug for years). So i wrote a small autohotkey script that battles this, cant play without it.

; #IfWinActive, ahk_class ArenaNet_Dx_Window_Class ; if dx 9
#IfWinActive, ahk_class ArenaNet_Gr_Window_Class ; if dx 11

$RButton::
	MouseGetPos, CoordXRec, CoordYRec
	sendevent {RButton down}
	keywait RButton
	sendevent {RButton up}
	MouseMove, %CoordXRec%, %CoordYRec%
return

what it does is that it records the mouse cursor position when i click me right mouse key the return the cursor to this potion when i let it go.

To the problem. When i start Blish Hud it works fine but as sone as i click my right mouse key to pan the camera, the icons stops functioning.

dlamkins:
There is a known issue with certain AutoHotKey scripts breaking our mouse hook.  Until we figure out what's causing this, you can attempt to run Blish HUD with the `--debug` flag at launch which slightly changes the way the mouse hook works and should allow it to continue to work while you have AHK enabled and running for Guild Wars 2.

dlamkins:
Thank you so much for providing an example script.  I've finally been able to replicate the issue and I believe we have it solved now.  Could you please try this build and let me know if this resolves the issue for you?

[Blish HUD.0.11.3-PullRequest0574.13.zip](https://github.com/blish-hud/Blish-HUD/files/7859156/Blish.HUD.0.11.3-PullRequest0574.13.zip)
.

PR #574, opened 2022-01-13
Hopefully fixes all of the input issues encountered occassionally

dlamkins:
Looks like the struct we were using for the mouse input was wrong and there's some flag that causes the Extras field to overflow the poor int which was crashing the mouse hook.  We weren't catching this because the mouse hook was successfully being created - it just wasn't crashing until after one of these inputs with the extra data occurred and as it was on a different thread, we weren't seeing it throw.

I'm not actually sure what the extra data is on the input, but we've corrected the data type as a long which appears to have resolved the issue.


dlamkins:
Closes #569

PR #591, opened 2022-01-27
Allow custom textbox background color

agaertner:
Small tweak to hide the background of a multiline textbox in order to use a custom one or to have it react dynamically on focus change.
The simple textbox uses a texture and I wasn't sure if I should redo it to be drawn by a single pixel like in the multiline textbox. I could remove the texture and allow color customization here as well.

### DIFF
--- a/Blish HUD/Controls/MultilineTextBox.cs
+++ b/Blish HUD/Controls/MultilineTextBox.cs
@@ -8,6 +8,16 @@ namespace Blish_HUD.Controls {
         private const int TEXT_TOPPADDING  = 7;
         private const int TEXT_LEFTPADDING = 10;
 
+        private Color _textBackgroundColor = Color.Black;
+        public Color TextBackgroundColor {
+            get => _textBackgroundColor;
+            set {
+                if (_textBackgroundColor.Equals(value)) return;
+                _textBackgroundColor = value;
+                OnPropertyChanged(nameof(TextBackgroundColor));
+            }
+        }
+
         public MultilineTextBox() {
             _multiline = true;
             _maxLength = 524288;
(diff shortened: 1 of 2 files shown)


dlamkins:
Seems most intuitive / useful if we just have the option to hide the background.  Controls already have a background color, but if a custom background was needed it would probably be best to just throw the multiline textbox without a background into a container which draws what is needed.

agaertner:
Like so?

Blish HUD/Controls/MultilineTextBox.cs
```diff
@@ -8,6 +8,16 @@ public class MultilineTextBox : TextInputBase {
         private const int TEXT_TOPPADDING  = 7;
         private const int TEXT_LEFTPADDING = 10;
 
+        private bool _hideBackground;
+        public bool HideBackground {
+            get => _hideBackground;
+            set {
```
dlamkins: Use `SetProperty` instead which automates all of this.

Blish HUD/Controls/MultilineTextBox.cs
```diff
@@ -193,24 +203,26 @@ public class MultilineTextBox : TextInputBase {
         protected override void UpdateScrolling() { /* NOOP */ }
 
         protected override void Paint(SpriteBatch spriteBatch, Rectangle bounds) {
-            // Background tint
-            spriteBatch.DrawOnCtrl(this, ContentService.Textures.Pixel, new Rectangle(1, 1, bounds.Width - 2, bounds.Height - 2), Color.Black * 0.5f);
+            if (!HideBackground) {
```
dlamkins: For consistency, let's do `this.HideBackground`.

Blish HUD/Controls/Textbox.cs
```diff
@@ -26,6 +26,16 @@ public class TextBox : TextInputBase {
         /// </summary>
         public event EventHandler<EventArgs> EnterPressed;
 
+        private bool _hideBackground;
+        public bool HideBackground {
+            get => _hideBackground;
+            set {
```
dlamkins: Use `SetProperty` instead which automates all of this.

Blish HUD/Controls/Textbox.cs
```diff
@@ -129,14 +139,20 @@ public class TextBox : TextInputBase {
...
         protected override void Paint(SpriteBatch spriteBatch, Rectangle bounds) {
-            spriteBatch.DrawOnCtrl(this,
-                                   _textureTextbox,
-                                   new Rectangle(Point.Zero, _size - new Point(5, 0)),
-                                   new Rectangle(0, 0, Math.Min(_textureTextbox.Width - 5, _size.X - 5), _textureTextbox.Height));
-
-            spriteBatch.DrawOnCtrl(this, _textureTextbox,
-                                   new Rectangle(_size.X - 5, 0, 5, _size.Y),
-                                   new Rectangle(_textureTextbox.Width - 5, 0, 5, _textureTextbox.Height));
+            if (!HideBackground) {
```
dlamkins: Same here, let's use `this.HideBackground`.

dlamkins (changes requested):

dlamkins:
Yeah, that's exactly what I meant.  🙂 Made a couple of notes for quick revisions that should only take a moment.  👍

agaertner:
> Yeah, that's exactly what I meant. 🙂 Made a couple of notes for quick revisions that should only take a moment. 👍

All done 👍

dlamkins (approved):

dlamkins:
Perfection.  Thanks!

PR #593, opened 2022-01-28
HorizontalAlignment.Right & .Center for TextBox

agaertner:
Custom HorizontalAlignment for TextBox.

Currently just for the single line TextBox. 
Eventually cool to have for MultiLineTextBox as well, but a bit trickier to implement (recalculation of highlightregion etc).

Preview: https://i.imgur.com/DUQt0lB.gif

### DIFF
--- a/Blish HUD/Controls/TextInputBase.cs
+++ b/Blish HUD/Controls/TextInputBase.cs
@@ -80,6 +80,16 @@ namespace Blish_HUD.Controls {
             InputFocusChanged?.Invoke(this, e);
         }
 
+        protected HorizontalAlignment _horizontalAlignment = HorizontalAlignment.Left;
+
+        /// <summary>
+        /// Gets or sets the horizontal alignment of the input.
+        /// </summary>
+        public HorizontalAlignment HorizontalAlignment {
+            get => _horizontalAlignment;
+            set => SetProperty(ref _horizontalAlignment, value);
+        }
+
         protected string _text = string.Empty;
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


dlamkins:
If it's gonna be exposed publicly on the base class but not implemented in the multiline one yet, I don't think we can leave it only implemented on the standard textbox.

I think I'd prefer the property and its implementation remain in the standard textbox control - i.e. not in the base input control class.  I don't see an immediate use case for horizontal alignment for the multi-line box (outside of RTL, which this really isn't a solution for and kind of ambitious for Blish HUD).

If it becomes difficult to implement that way, you could perhaps include the horizontal alignment as a parameter in the call to draw text and have it default left side or something.

Let me know your thoughts.

dlamkins:
An aside, but wanted to mention it looks really good in that book UI demo you showed.

agaertner:
> If it's gonna be exposed publicly on the base class but not implemented in the multiline one yet, I don't think we can leave it only implemented on the standard textbox.
> 
> I think I'd prefer the property and its implementation remain in the standard textbox control - i.e. not in the base input control class. I don't see an immediate use case for horizontal alignment for the multi-line box (outside of RTL, which this really isn't a solution for and kind of ambitious for Blish HUD).
> 
> If it becomes difficult to implement that way, you could perhaps include the horizontal alignment as a parameter in the call to draw text and have it default left side or something.
> 
> Let me know your thoughts.

I had that thought as well. It's all in the TextBox class now.

dlamkins:
Looks great!

dlamkins (approved):

Issue #533, opened 2021-10-23
Alt-tab/Switching windows causes full-screen black flash


tuftelien:
With BHUD Frame Limiter set to Sync With Game.

WIth other BHUD Frame Limiter settings menus still flash.
BHUD Vsync doesn't seem to change the behavior.

GW2 settings: Windowed Fullscreen, frame limiter unlimited, FPS 60-80

dlamkins:
Hello!

This is a known issue related to how we handle frame synchronization to help keep CPU utilization low when the game framerate is low.

We'll  keep this issue to track alternative methods to how we handle frame syncing accuracy and our balance with CPU utilization.

In the meantime, If it is a common issue for you, I'd recommend setting Blish HUD to 90 FPS on the limiter with vsync enabled and in fact is our new default setting in the latest pre-releases. 👍

dlamkins:
This should no longer be happening.

PR #594, opened 2022-01-31
Fix window blinking on launch

jojorne:
Must be the same color as the device clear color but without alpha transparency.

### DIFF
--- a/Blish HUD/BlishHud.cs
+++ b/Blish HUD/BlishHud.cs
@@ -60,9 +60,8 @@ namespace Blish_HUD {
             Form       = Control.FromHandle(FormHandle).FindForm();
 
             // Avoid the flash the window shows when the application launches
-            // Must be the same color as the device clear color but without alpha transparency
-            Form.BackColor       = System.Drawing.Color.Black;
-            Form.TransparencyKey = System.Drawing.Color.Black;
+            Form.BackColor = System.Drawing.Color.Black;
+            Form.Location  = new System.Drawing.Point(-Form.Width * 2, -Form.Height * 2);
 
             this.Window.IsBorderless = true;
             this.Window.AllowAltF4   = false;
--- a/Blish HUD/BlishHud.cs
+++ b/Blish HUD/BlishHud.cs
@@ -60,8 +60,9 @@ namespace Blish_HUD {
             Form       = Control.FromHandle(FormHandle).FindForm();
 
             // Avoid the flash the window shows when the application launches
-            Form.Size     = new Size(1, 1);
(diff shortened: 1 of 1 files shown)


jojorne:
Complete code to launch a transparent window:
```csharp
_form = Control.FromHandle(this.Window.Handle).FindForm();
_form.FormBorderStyle = FormBorderStyle.None;
_form.BackColor = Color.Black;
_form.TransparencyKey = Color.Black;
_form.AllowTransparency = true;
```

entrhopi:
As I understand, this will just override the white flash with a black one. From my testing I can't get it to work in transparent mode using the `TransparencyKey` you suggested. Thou more pleasant for the eye aren't there any 'better' solutions for this?

I'm not very deep in this topic but maybe this could be improved further to fix #533 too?

Thanks for looking into this topic!
The white flash GW2 makes is annoying and if we could get this to work we would have a nice edge over Anet. xD

jojorne:
Because Blish HUD uses `WS_EX_LAYERED` to control transparency, it overrides the MonoGame transparency control. Setting `TransparencyKey` does nothing. The idea here is to set `BackColor` to `Black` so `SetLayeredWindowAttributes` finds the color reference but there's still a black rectangle until `SetLayeredWindowAttributes` executes, because of that we set the form's `Location` off the screen. When the form is fully initialized and back in place, it is transparent. After that, MonoGame takes over and starts rendering transparently.

dlamkins:
I'll take a look and try out the build.

We might be able to close #533 after the next release - I've removed the "Sync with Game" frame limiter option since it was causing more problems than it was solving.

entrhopi:
Nice, thanks for the explanation.
One question ... what happens if the user has another display at the (top-left) location?

jojorne:
I haven't tried it, but perhaps changing the order in which things execute might fix things. Because the latest version of MonoGame supports form style, no borders, etc. First we open a transparent form then we set the `WS_EX_LAYERED` and start the engine? I don't know if we can set that after the form is created tho. Also, MonoGame does not support toggle between show and hide on taskbar after form creation.

dlamkins:
> One question ... what happens if the user has another display at the (top-left) location?

This is actually what we're already doing and it's set to a negative position.  It isn't relative, so no matter what it will be off screen.

It's actually nearly what Windows does when you press the minimize button.  It moves it to like -50000, -50000.

dlamkins (approved):

dlamkins:
Good detective work narrowing down the white flash.  Thanks again.  👍

PR #597, opened 2022-02-04
User defined interact key

### LINKED ISSUE #582: Allow setting the keybind for interaction
Since users can rebind it in the game, it should be editable without having to compile. Might only be used in the pathing module at the moment; that's where I noticed it.

https://github.com/blish-hud/Blish-HUD/blob/279c620296327bea48a2583046da1085cfb06c73/Blish%20HUD/Common/Gw2/KeyBindings.cs#L15

Referred to from here: https://github.com/blish-hud/Pathing/blob/f8bb925c33b7d13dcd6c77ad469a7a3d616b41dd/State/SharedPackState.cs#L50

entrhopi:
#582

New settings entry so users can set interact to the same keybind they use ingame.

### DIFF
--- a/Blish HUD/GameServices/OverlayService.cs
+++ b/Blish HUD/GameServices/OverlayService.cs
@@ -46,6 +46,7 @@ namespace Blish_HUD {
         public SettingEntry<Locale> UserLocale    { get; private set; }
         public SettingEntry<bool>   StayInTray    { get; private set; }
         public SettingEntry<bool>   ShowInTaskbar { get; private set; }
+        public SettingEntry<KeyBinding> InteractKey { get; private set; }
         public SettingEntry<KeyBinding> ToggleBlishWindow { get; private set; }
         public SettingEntry<bool>   CloseWindowOnEscape { get; private set; }
         public SettingEntry<KeyBinding> HideAllInterface { get; private set; }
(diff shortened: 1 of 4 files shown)


Blish HUD/GameServices/OverlayService.cs
```diff
@@ -46,6 +46,7 @@ public class OverlayService : GameService {
         public SettingEntry<Locale> UserLocale    { get; private set; }
         public SettingEntry<bool>   StayInTray    { get; private set; }
         public SettingEntry<bool>   ShowInTaskbar { get; private set; }
+        public SettingEntry<KeyBinding> InteractKey { get; private set; }
```
dlamkins: I think eventually we'll want to rework where all of the settings for the keybinds so to avoid a breaking change down the line, could we make this an internal property for now?  😃

dlamkins:
Once the above is addressed, we should be ready to merge.  🙂 

EDIT: would be good if I actually clicked the submission on the review.  😅

dlamkins (changes requested):

dlamkins (approved):

jojorne:
Can we use the mouse for binding?

dlamkins:
Only if it produces a key press event.  Similarly to the game itself, we don't support mouse button bindings.

jojorne:
Oh? Okay. It's just that my in-game action key is Mouse3.

dlamkins:
Ah, I see, the game does allow mouse buttons for binding.  You are correct.

In any case, though, the answer is that this is not currently supported in Blish HUD.  It would require some bifurcation of the input logic.

Most mice support rebinding the mouse buttons to standard key presses, so at this time that would be our only supported way you could use that button for interact.

PR #601, opened 2022-02-08
Fix tween update

jojorne:
This fixes the animation on the update icon and any animation that uses the tween method.

### DIFF
--- a/Blish HUD/Library/Glide/Tween.cs
+++ b/Blish HUD/Library/Glide/Tween.cs
@@ -112,17 +112,17 @@ namespace Glide
 				
 				if (time == 0 && timesRepeated == 0 && begin != null)
 					begin();
-
-				time += elapsed;
-				float setTimeTo = time;
-				float t = time / Duration;
+                
 				bool doComplete = false;
-				
+                
+                float t;
+                time += elapsed;
 				if (time >= Duration)
 				{
+                    t = 1f;
 					if (repeatCount != 0)
 					{
-						setTimeTo = 0;
+						time = 0;
 						Delay = repeatDelay;
 						timesRepeated++;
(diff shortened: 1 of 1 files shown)


jojorne:
Uh... I think there's something wrong with GitHub's indentation...

jojorne:
Also, I noticed that the animation is being affected by game time. 60 FPS is twice as fast as 30 FPS.

Blish HUD/Library/Glide/Tween.cs
```diff
@@ -55,348 +55,349 @@ public enum RotationUnit
...
-						
-						if (repeatCount > 0) {
+                float t;
+                time += elapsed;
+                if (time >= Duration)
+                {
+                    t = 1;
+                    if (repeatCount != 0)
+                    {
+                        time = 0;
```
jojorne: When reseting time, the right way should be:
```csharp
time -= Duration;
```
By doing it this way, you don't lose time precision.

entrhopi:
This is working nicely now. I can't see any issues. Even the rubberbanding effect when framerate is set to unlimited seems to be gone.

jojorne:
Fixed a rare case where opening/closing a window would cause it to blink.

dlamkins:
FYI, I know that the original files were poorly formatted when we pulled and committed them from their original repo, but the formatting changes included in this PR are making it quite hard to tell what specifically was changed for the review.

jojorne:
image
Does this help? Because Visual Studio kept changing the format to follow `.editorconfig` then I manually kept changing it back every time. I was just losing time with the formatting that in the and I gave up... lol

dlamkins (approved):

PR #604, opened 2022-02-11
Window Improvements

jojorne:
### Window Improvements
- No longer steal focus from the game on launch.
- No longer flashes on launch.
- Can be captured by software for video recording and live streaming even when not in taskbar.

### DIFF
--- a/Blish HUD/_Utils/WindowUtil.cs
+++ b/Blish HUD/_Utils/WindowUtil.cs
@@ -40,6 +40,11 @@ namespace Blish_HUD {
         private const uint SWP_NOMOVE         = 0x0002;
         private const uint SWP_NOACTIVATE     = 0x0010;
 
+        private const uint RDW_INVALIDATE  = 0x001;
+        private const uint RDW_ERASE       = 0x004;
+        private const uint RDW_ALLCHILDREN = 0x080;
+        private const uint RDW_FRAME       = 0x400;
+
         private const int MINIMIZED_POS = -32000;
 
         [DllImport("user32.dll", SetLastError = true)]
@@ -80,6 +85,9 @@ namespace Blish_HUD {
 
         [DllImport("user32.dll", EntryPoint = "SetWindowLongPtr")]
         private static extern uint SetWindowLongPtr64(IntPtr hWnd, int nIndex, UIntPtr dwNewLong);
+
+        [DllImport("user32.dll")]
+        private static extern bool RedrawWindow(IntPtr hWnd, IntPtr lprcUpdate, IntPtr hrgnUpdate, uint flags); 
 
         [DllImport("user32.dll")]
         private static extern bool SetWindowPos(IntPtr hWnd, IntPtr hWndInsertAfter, int X, int Y, int cx, int cy, uint uFlags);
@@ -137,6 +145,21 @@ namespace Blish_HUD {
 
         internal static void SetTransparentLayered(IntPtr winHandle) {
             SetWindowLong(winHandle, GWL_EXSTYLE, GetWindowLong(winHandle, GWL_EXSTYLE) | WS_EX_TRANSPARENT | WS_EX_LAYERED);
+        }
+
+        internal static void InvalidateWindow(IntPtr winHandle) {
+            RedrawWindow(winHandle, IntPtr.Zero, IntPtr.Zero, RDW_INVALIDATE | RDW_ERASE | RDW_ALLCHILDREN | RDW_FRAME);
+        }
+
+        internal static void ExtendFrameAcrossEntireWindow(IntPtr winHandle) {
+            var marg = new Margins {
+                    cxLeftWidth    = -1,
+                    cyTopHeight    = -1,
+                    cxRightWidth   = -1,
+                    cyBottomHeight = -1
+            };
+
+            DwmExtendFrameIntoClientArea(winHandle, ref marg);
         }
 
         internal static void SetNoActivate(IntPtr winHandle, bool noActivate) {
@@ -204,6 +227,7 @@ namespace Blish_HUD {
 
                 SetTransparentLayered(winHandle);
                 SetLayeredWindowAttributes(winHandle, 0, 255, LWA_ALPHA);
+                InvalidateWindow(winHandle);
             }
 
             return (OverlayUpdateResponse.WithFocus, screenPoint.X == MINIMIZED_POS, 0);
--- a/Blish HUD/_Utils/WindowUtil.cs
+++ b/Blish HUD/_Utils/WindowUtil.cs
@@ -19,6 +19,7 @@ namespace Blish_HUD {
         private const uint WS_EX_CONTROLPARENT = 0x00010000;
         private const uint WS_EX_APPWINDOW     = 0x00040000;
         private const uint WS_EX_LAYERED       = 0x00080000;
+        private const uint WS_EX_NOACTIVATE    = 0x08000000;
 
         private const int GWL_STYLE   = -16;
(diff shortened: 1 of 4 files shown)


entrhopi:
I do not experience any more flashing. This is great!

However ... I'm not sure if we want to hide all off the interface when there is no focus on the GW2 window. People who switch between processes constantly might find this very confusing.

jojorne:
What do you mean? I mean, when GW2 loses focus, the HUD becomes unresponsive, useless. I could even hide it from task bar so it does not become selectable. Maybe, perhaps even adding a setting called `Hide when GW2 loses focus`? idk... Although apps like Obs need it in taskbar so the HUD can be captured. I still don't know how it would be the experience for those users.

Blish HUD/_Utils/WindowUtil.cs
```diff
@@ -182,14 +181,16 @@ public enum OverlayUpdateResponse {
 
             // If Guild Wars 2 is not the focused application, set Blish HUD to
             // be above Guild Wars 2, but below the next application's window in z-order
-            if (activeWindowHandle != gw2WindowHandle && Form.ActiveForm == null) {
```
dlamkins, line 185: No, this is necessary for modules that use external forms.  It's one of the few times that we can truly take focus from the game when not shown in taskbar.

dlamkins:
> What do you mean? I mean, when GW2 loses focus, the HUD becomes unresponsive, useless. 

Just as the game itself doesn't stop rendering it is intended that we continue to render.  
- Important for multiboxers where the UI would be popping in and out as they switched between windows which is jarring.
- It would make screenshots with certain utilities more difficult.
- Removes the immersion that the UI is part of the game.

> I could even hide it from task bar so it does not become selectable.

There is already an option to toggle if it shows on the taskbar or not.  By default it is not intended to be shown on the taskbar.

> Maybe, perhaps even adding a setting called `Hide when GW2 loses focus`?

That is an option that could be considered.

Blish HUD/_Utils/WindowUtil.cs
```diff
@@ -223,6 +224,10 @@ public enum OverlayUpdateResponse {
                 DwmExtendFrameIntoClientArea(winHandle, ref marg);
             }
 
+            if (!wasOnTop) {
+                SetLayeredWindowAttributes(winHandle, 0, 255, LWA_ALPHA);
```
jojorne: If not hiding the window, I might need a better place for this then, a place where it executes only once when the window is ready to be seen.

Blish HUD/_Utils/WindowUtil.cs
jojorne: Do you know if there's any module like that? I want to check how it behaves.

dlamkins:
> Improve the fix for the HUD flashing on launch.

Could you elaborate on where else you were seeing flashing?  I can't seem to replicate any flashing after that last fix you added in #594.

jojorne:
> > Improve the fix for the HUD flashing on launch.
> 
> Could you elaborate on where else you were seeing flashing? I can't seem to replicate any flashing after that last fix you added in #594.

I consider that as a hack. This one starts with an invisible window but needs to make it opaque again when the HUD is ready to be seen. And before the first draw call.

But thinking about it if any module makes any changes to the form or window style, MonoGame resets the form style, this includes the `Topmost`, making the window opaque again. I might need to look into that...

Anyways, this try to avoid those overrides. What I mean is that the settings on initialization must be in sync with MonoGame.

Blish HUD/_Utils/WindowUtil.cs
dlamkins: Yes, there's the debug module: https://cdn.discordapp.com/attachments/536970543736291346/877228330213048330/Debug.bhm

Also the Pathing module.  Right-click a marker on the minimap while holding SHIFT and you'll get additional entries in the dropdown menu including "Edit Marker".  Clicking that will open the WIP marker editing window.

Blish HUD/_Utils/WindowUtil.cs
```diff
@@ -130,18 +137,8 @@ private enum GW : uint {
             }
         }
 
-        internal static void SetShowInTaskbar(IntPtr winHandle, bool showInTaskbar) {
-            SetWindowParam(winHandle, showInTaskbar);
-        }
-
-        internal static void SetupOverlay(IntPtr winHandle) {
-            SetWindowLong(winHandle, GWL_STYLE, CS_HREDRAW | CS_VREDRAW);
```
jojorne, line 138: I believe this doesn't belong here. See [GWL_STYLE](https://docs.microsoft.com/en-us/windows/win32/winmsg/window-styles).

dlamkins:
Okay, initial things after first pass review:

1. **Show in Taskbar** should show the icon regardless of if the game has focus or not if checked.

image

Hiding it when the game loses focus doesn't make much sense.  Especially since this feature is commonly used by folks who need to select it in recording software as a source.  When it is hidden from the taskbar, it will stop showing in the list of applications for most users.  The only time we want to hide the Blish HUD icon from the taskbar while this setting is enabled is when we're not actively attached to an instance of Guild Wars 2 and are sitting in the notification area waiting for a Guild Wars 2 instance.

2. It seems showing in the taskbar and showing the overlay in-game at all have been conflated for some reason.  I'm not sure why these two are connected.

For example, most users prefer not to have **Show in Taskbar** enabled (so it is off by default) but would expect the overlay itself to display with the focus is changed to a different application.  Hiding the application from the taskbar and keeping the overlay visible when Guild Wars 2 loses focus should not be mutually exclusive features.

A setting labeled **Hide Blish HUD When GW2 Is in the Background** (phrased like the option in the Guild Wars 2 settings for muting the music) may be more appropriate and describe better what perhaps your intentions are here, perhaps?

3. Something weird seems to have happened to focus when WinForms are selected.  It is expected that, regardless of the settings selected, Blish HUD will not hide regardless of what settings are selected in the overlay settings as long as there is an active GW2 instance that we're attached to.

https://gfycat.com/allelderlyharlequinbug

You can see the behavior here.  Two things stand out: 1) The WinForm window is somehow ending up behind what Blish HUD is rendering. 2) Blish HUD is sometimes hiding itself when the WinForm window is selected.

You can see how I can't recreate this behavior in v0.11.4:

https://gfycat.com/majorbossyangwantibo

Blish HUD/_Utils/WindowUtil.cs
```diff
@@ -180,24 +210,39 @@ public enum OverlayUpdateResponse {
...
                 return (OverlayUpdateResponse.WithoutFocus, screenPoint.X == MINIMIZED_POS, 0);
             }
 
+            UpdateProc(winHandle, gw2WindowHandle, wasOnTop, clientRect, screenPoint);
+
             if (!wasOnTop) {
                 Logger.Debug("GW2 is now the active window - reactivating the overlay.");
+
+                SetTransparentLayered(winHandle);
+                SetLayeredWindowAttributes(winHandle, 0, 255, LWA_ALPHA);
```
jojorne, line 229: Transparent window bug:
SetLayeredWindowAttributes(winHandle, 0, 0, LWA_COLORKEY);
SetLayeredWindowAttributes(winHandle, 0, 255, LWA_ALPHA);

jojorne: oh! i see what you did there. i was trying to find a solution for Linux too but without success, they all have the black background...

jojorne: oh! someone made an Overlay class for MonoGame on GitHub... now where is it again...

PR #605, opened 2022-02-12
Only accept clicks when the mouse is visible

jojorne:


Blish HUD/GameServices/Input/Mouse/MouseHandler.cs
```diff
@@ -103,7 +104,11 @@ public class MouseHandler : IInputHandler {
         }
 
         public void Update() {
-            if (!GameService.GameIntegration.Gw2Instance.Gw2IsRunning || !GameService.GameIntegration.Gw2Instance.Gw2HasFocus || GameService.Overlay.InterfaceHidden) {
+            var cursorInfo = PInvoke.GetCursorInfo();
+
+            var isMouseVisible = cursorInfo.flags == CURSORFLAGS.CURSOR_SHOWING;
+
+            if (!GameService.GameIntegration.Gw2Instance.Gw2IsRunning || !GameService.GameIntegration.Gw2Instance.Gw2HasFocus || GameService.Overlay.InterfaceHidden || !isMouseVisible) {
```
dlamkins: As modules may introduce more complex behaviors, it's likely better that we not ignore ALL mouse input if the cursor is hidden.  Better might be to expose if the cursor is visible on the MouseHandler and skip click input being passed along and perhaps avoid sending input to controls.

This allows our input states to still be accurate while avoiding the issues you've experienced with controls receiving input while in action camera.

dlamkins (changes requested):

dlamkins (approved):

PR #607, opened 2022-02-19
Add OpenInDefaultBrowser

agaertner:
Tested and cleaned.

~~There is only one annoyance in which the Bhud overlay doesn't hide when the browser is brought to top.~~
~~Possible solutions:~~
~~1. Send an input message to the browser to acquire focus so that bhud realizes gw2 lost focus.~~
~~2. Minimize bhud main window handle forcefully in hopes it will restore itself when gw2 reacquires focus.~~

EDIT: Focus improvements and fixes in #604 

This can be used to open download links in the Marker Repo etc.

### DIFF
--- a/Blish HUD/Controls/Extern/PInvoke.cs
+++ b/Blish HUD/Controls/Extern/PInvoke.cs
@@ -41,9 +41,6 @@ namespace Blish_HUD.Controls.Extern
         internal delegate void SendAsyncProc(IntPtr hWnd, uint uMsg, uint dwData, int lResult);
 
         [DllImport("user32.dll")]
-        internal static extern bool AttachThreadInput(uint idAttach, uint idAttachTo, bool fAttach);
-
-        [DllImport("user32.dll")]
         [return: MarshalAs(UnmanagedType.Bool)]
         internal static extern bool SetForegroundWindow(IntPtr hWnd);
 
--- a/Blish HUD/_Utils/BrowserUtil.cs
+++ b/Blish HUD/_Utils/BrowserUtil.cs
@@ -10,20 +10,108 @@ using System.Threading.Tasks;
 namespace Blish_HUD._Utils {
     public class BrowserUtil {
         /// <summary>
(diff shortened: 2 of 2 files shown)


dlamkins:
This procedure relies on too many fragile / unintended things.  I don't know if I'd like to support this as a util within core.  It may be best kept in the individual modules if they actually end up relying on them.

agaertner:
You mean querying the registry?

If you tell me which part you are unsure about I can try and find a different solution.

I PR'd it into Core because I found it to be a solution for not having visual feedback when you click on a marker pack to download in the Pathing Module and would likely use it in all of my modules, too.

agaertner:
Done with refactor. I've drastically reduced complexity. You might want to take a look at it again now.
If you are still unsure, would you like me to PR it into Pathing and fix those download links or would you like to have a different visual feedback solution entirely in the future? (Eg. download progress as a control without opening browser at all, ...)

dlamkins:
Oh, I meant the fact that you were circumventing the anti-focus features in Windows.  That's what was fragile to me.

Currently pathing module doesn't offer links to open up and download markers packs.  It's just the timers module that I think links to any downloads currently.  Pathing module will be downloading them behind the scenes in-UI much like how the module repo itself handles downloads.

agaertner:
Oh, I must have forgotten which module it was where I got a dozen hero.zips before I noticed that downloads were happening.

Tbh, I don't think that part isn't too fragile. It's meant to be this way. We are becoming part of the foreground thread for a fraction of a second.

> will be downloading them behind the scenes in-UI much like how the module repo itself handles downloads.

Perfect!

Can I convince you to use it for other non-file links (ko-fi, discord, etc) that should open in browser? Otherwise this PR can be closed and I will be adding this to all my modules.

dlamkins:
We don't currently have any other links that come to mind in core, so I'll close and if it comes up in the future in core we'll re-review and consider it.  👍

PR #626, opened 2022-03-07
Rename the presenter file to align with its inner class name

sonicolasj:
In the code, the Presenter class has 2 generics parameters: `TView` and `TModel`, but the file code is named `Presenter[TView,TPresenter].cs`, using TPresenter `instead` of `TModel`


dlamkins:
Thanks!

dlamkins (approved):

PR #649, opened 2022-03-26
Fix #614 Add Selection dragging support to TextInputBase

### LINKED ISSUE #614: TextInputBase does not support mouse drag to mark text
Current behavior:
Holding the left mouse button and moving to a different location there the button is release will only update the cursor to be at the new position.

Expected behavior:
All text between mouse down and mouse up events should be selected with the cursor at the mouse up position

dylanrenwick:
Fixes issue #614 

Notably this change moves text cursor index movement to mouse pressed, rather than mouse released (click). This is consistent with other text input fields such as VS, GitHub, and GW2's own chat, but requires that we also move setting the control's focus to global mouse pressed, from global mouse released.

### DIFF
--- a/Blish HUD/Controls/TextInputBase.cs
+++ b/Blish HUD/Controls/TextInputBase.cs
@@ -752,8 +752,8 @@ namespace Blish_HUD.Controls {
         }
 
         protected void HandleMouseSelectionDrag(int newIndex) {
-            this.SelectionStart = Math.Min(_cursorIndex, newIndex);
-            this.SelectionEnd = Math.Max(_cursorIndex, newIndex);
+            UserSetCursorIndex(newIndex);
+            this.SelectionEnd = newIndex;
         }
 
         protected override void OnLeftMouseButtonPressed(MouseEventArgs e) {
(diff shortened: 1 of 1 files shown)


dlamkins:
This is looking pretty awesome!  I'm seeing some edge cases here that we'll want to be fixed, though.  I'm seeing the selection area persist after content is deleted if the text is selected right-to-left.  Selecting text left-to-right and then doing something to delete the text did not incur this issue.

You can see this behavior below:
https://gfycat.com/imaginarylividiberiannase

You can ignore the mouse showing offset from the selections.  It's an artifact of the recording and not actually representative of where the mouse was.

dylanrenwick:
Good catch! The latest commit fixes that.

Did you find any other edge cases that need to be addressed?

dlamkins:
Thanks!  I'll try again and check.  👍

dlamkins:
Those issues appear to be resolved now and I haven't spotted any other issues in testing.  Thanks for the contribution!

dlamkins (approved):

PR #650, opened 2022-03-27
Add mouse and keyboard util; Mark old intern/extern stuff as obsolete

agaertner:
This time I did not remove the old stuff just marked them as obsolete so existing modules won't break.
This way we can give it some time to settle until we can confidently remove the old stuff.

Also as people have pointed out the utils now take an int as keyCode instead of an enum (shorts) which is more convinient.

### DIFF
--- a/Blish HUD/Common/Gw2/GuildWarsControls.cs
+++ b/Blish HUD/Common/Gw2/GuildWarsControls.cs
@@ -1,10 +1,5 @@
-﻿using System;
-
-namespace Blish_HUD.Controls.Intern
-{
-    [Obsolete("Controls.Intern.GuildWarsControls is obsolete. Please use Common.Gw2.GuildWarsControls instead.", true)]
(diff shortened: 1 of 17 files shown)


dlamkins:
https://github.com/blish-hud/Blish-HUD/pull/502#issuecomment-1032215453

I think no reason to do this until we switch to SDK.  Only gives folks work to swap their code over which will then have to get swapped again in the future.

If there are individual fixes to things here, though, you can PR those improvements to the existing implementation.

agaertner:
What about not setting ``error: true`` so they can still compile it? It gives us flexbility as SDK is still in planning (I believe)?. Or is SDK like a confirmed thing and still coming this year? I take it back if so. 

That reminds me, I really need some way like pins to stay up-to-date. Like a kanban. The discussions on discord just disappear too quickly. I did not follow the discussion about an SDK too much,

PR #653, opened 2022-03-29
Revise graphics device locking method.

dlamkins:

PR #654, opened 2022-03-29
Add GraphicsDeviceContext helper struct + methods

eksime:
Hope these types of PR are ok - this is pretty low priority, as it's just a developer usability thing, but it may be useful :)

This adds a `GraphicsDeviceContext` helper struct + associated methods for use with the tri-lock code added in #644 that makes idiomatic code like: 

```csharp
using var ctx = Graphics.LendGraphicsDeviceContext()
{
// some code using ctx
}
```

possible, which will hopefully help to ensure that graphics devices are returned, including in the case of errors, without having to remember to use a try/finally block or call the release method.

It's implemented as a `readonly ref struct` so it shouldn't incur extra heap allocations and performance impact should be minimal. The existing `LendGraphicsDevice` methods remain, so that both approaches can be used.

### DIFF
--- a/Blish HUD/GameServices/GraphicsDeviceContext.cs
+++ b/Blish HUD/GameServices/GraphicsDeviceContext.cs
@@ -1,0 +1,32 @@
+﻿using Microsoft.Xna.Framework.Graphics;
+
+namespace Blish_HUD {
+    public readonly ref struct GraphicsDeviceContext {
+
+        private readonly GraphicsService _service;
+
+        /// <summary>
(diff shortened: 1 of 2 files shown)


eksime:
gah, I must've started this right as #653 went in and I didn't notice >.< - feel free to close this if you want, as they both sorta accomplish the same thing

dlamkins:
I think this is a better solution!  Let me review and ensure it resolves the issue I was running into.  👍

dlamkins (approved):

PR #664, opened 2022-04-05
Align settings entries

entrhopi:
Better alignment for settings

grafik

The goal is to align dropdowns, keybinds and sliders within a settings section.


dlamkins:
This will be quite nice!

entrhopi:
Sooo ... after pondering about it for apparently far too long, I've come to the conclusion that it is better not to align all adjusted for the label size, but to use sane defaults and let singular settings 'jump out' if needed.

image
This is the current state.

What I would like to change thou is the font size I think. Comparing the settings to the game UI 'our' text is a bit smaller and not as 'white'.

What do you think?

greaka:
Imo, this looks great as it is. Text scaling is hard, if you can get it right, then go for it in a separate PR :)
For the color, a separate PR is also fine, we will need to see how much it affects different systems and modules.

greaka (approved):

entrhopi:
I'll rebase and update this, now that #1040 is merged.

dlamkins:
Honestly, looks pretty great!  I'm noticing in some modules that the ViewContainer that has the settings rendered in it goes outside of the parent view.  🤔 

image

Not sure if this is caused by the module providing the wrong width for the settings rendering or if these changes are causing it to break out passed the width.  Here is the width of the parent container:

image

Same with this module (though it is reaaaaally close just a couple pixels off):

image

image

---- 

A closer look at both modules shows that it escapes the parent boundary even before this change, but it just goes unnoticed because the bar isn't quite as far over.  🤔 

image

Same here:

image

---

If we can confirm that this is just the modules specifying too wide of an area then I think I'm happy to move forward with this.  🙂

entrhopi:
I'll have a look at this.
At first glance it seems to me that at least @manlaan made their own settings alignment within the module, overwriting the defaults.

I can however add a check, that adjusts the with depending on the parent view. What do you think?

dlamkins (approved):

PR #675, opened 2022-04-26
Fixed various memory leaks

DarkLiKally:
1. removed the adhesive bindings which were causing to keep alive all objects using a scrollbar component and through this all related event listeners, views, presenters and controls
2. fixed some event listeners in various controls added but not removed anymore
3. fixed ViewContainer / WindowBase2  not calling DoUnload on old view when it is replaced with a new view instance.
4. resolved some exceptions occuring due to prevented access to read and write locks


Blish HUD/Controls/TextInputBase.cs
```diff
@@ -64,7 +64,10 @@ public abstract class TextInputBase : Control {
         /// </summary>
         public event EventHandler<ValueEventArgs<bool>> InputFocusChanged;
 
-        protected void OnTextChanged(ValueChangedEventArgs<string> e) => TextChanged?.Invoke(this, e);
+        protected void OnTextChanged(ValueChangedEventArgs<string> e) {
+            OnPropertyChanged(nameof(Text));
```
dlamkins: This PR may not be appropriate for this change, but in any case, this is already called in `SetText` when `SetProperty` is called.  The issue is just that the callername is `SetText` instead of `Text`.  The fix is to update `if (!SetProperty(ref _text, value)) return false;` to include the explicit property name instead of it pulling it with CallerMemberName.

Blish HUD/Controls/ViewContainer.cs
```diff
@@ -42,6 +42,11 @@ public class ViewContainer : Panel, IViewContainer {
 
             ViewState = ViewState.Loading;
 
+            if (this.CurrentView != null) {
+                this.CurrentView.Loaded -= BuildView;
+                this.CurrentView.DoUnload();
+            }
```
dlamkins: DoUnload is already called in `Clear()`.  The event handler can be removed there as well.

Blish HUD/Controls/WindowBase2.cs
```diff
@@ -308,6 +308,11 @@ public abstract class WindowBase2 : Container, IWindow, IViewContainer {
 
             this.ViewState = ViewState.Loading;
 
+            if (this.CurrentView != null) {
```
dlamkins: Same thing here.  This all should be moved to ClearView.  not within ShowView.

Blish HUD/Controls/Tooltip.cs
```diff
@@ -249,6 +249,11 @@ public class Tooltip : Container, IViewContainer {
         protected override void DisposeControl() {
             this.CurrentView?.DoUnload();
 
+            foreach (var control in this._children) {
```
dlamkins: No `this`

agaertner:
Looks good to me. :)

Blish HUD/Controls/_Types/ControlCollection.cs
```diff
@@ -59,7 +59,8 @@ private class ControlEnumerator<TEnum> : IEnumerator<TEnum> {
         }
 
         public IEnumerator<T> GetEnumerator() {
-            _listLock.EnterReadLock();
+            if (!_listLock.IsReadLockHeld)
```
dlamkins, line 62: To clarify, is this for same-thread locking because `ReaderWriterLockSlim` does not support reentrants here?

Blish HUD/GameServices/Modules/ModuleManager.cs
```diff
@@ -38,6 +38,8 @@ public class ModuleManager : IDisposable {
         /// </summary>
         public bool Enabled { get; private set; }
 
+        public bool IsDisabling { get; private set; }
```
dlamkins: These module changes probably shouldn't be in the PR and should probably have an explanation in the PR body.  Not sure the purpose here.

dlamkins (changes requested):

Blish HUD/GameServices/Modules/ModuleManager.cs
agaertner: Oh to add doesn't this have the same worth as Module.Loaded which already exists in the Module base?

DarkLiKally: Purpose:

When a module has some unload code that requires to load another dependency assembly, because the module was not used in the current session and the same dependency was not required in into code, then the assembly resolve function was not able to load the assembly because the enabled flag of the module was already set to false before. 

As discussed in Discord I'll create an issue for this instead of fixing it now.

Blish HUD/Controls/_Types/ControlCollection.cs
DarkLiKally: Yes when the same thread tries to enter the lock a second time, it throws an exception, preventing following code from executing.

Blish HUD/Controls/TextInputBase.cs
DarkLiKally: You'Re right :) Overlooked that one. I'll adjust

Blish HUD/Controls/Tooltip.cs
DarkLiKally: Adjusted this, also on the other spot.

Blish HUD/Controls/WindowBase2.cs
DarkLiKally: Fixed

Blish HUD/Controls/ViewContainer.cs
DarkLiKally: Fixed

Blish HUD/GameServices/Modules/ModuleManager.cs
dlamkins: Thanks, mate!

Blish HUD/Controls/_Types/ControlCollection.cs
dlamkins: Sounds good.  👍

dlamkins:
Thanks, mate!  Waiting for final build to test and then merge assuming all is solid.  😄

dlamkins:
Tests went well.  Thank you!

dlamkins (approved):

PR #676, opened 2022-04-27
FlowPanel & layout suspension updates

eksime:
* Add IDisposable suspend layout scope to SuspendLayout() - this should be backwards compatible with old code, but allows the use of the code below.  It'll hopefully help ensure people remember to call ResumeLayout, even if there's an error, and should be safe to use in any context. This also swaps the backing field from a bool to an int to track the scope depth, and introduces the internal property `IsLayoutSuspended` which checks if the control, or any ancestor, is currently suspended.
  ```
using (control.SuspendLayout())
{
    //...
}
  ```

 * Convert FlowPanel to use .Invalidate() where appropriate - this is related to the discord discussion, the flowPanel should now respect the layout being suspended before trying to arrange its children.

### DIFF
--- a/Blish HUD/Controls/Control.cs
+++ b/Blish HUD/Controls/Control.cs
@@ -12,6 +12,7 @@ using Blish_HUD.Controls.Effects;
 using Blish_HUD.Input;
 using Newtonsoft.Json;
 using MouseEventArgs = Blish_HUD.Input.MouseEventArgs;
+using System.Threading;
 
 namespace Blish_HUD.Controls {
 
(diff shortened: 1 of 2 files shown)


eksime:
I wasn't 100% sure if the parent being suspending should also mean that all descendants should also be suspended - but conceptually it _seems_ to make sense for that to be the case?

dlamkins (approved):

dlamkins:
> I wasn't 100% sure if the parent being suspending should also mean that all descendants should also be suspended - but conceptually it _seems_ to make sense for that to be the case?

I think that seems like a reasonable assumption.  

Thank you again for these great improvements!

PR #685, opened 2022-05-06
Gw2AuthService

agaertner:
Adds a service for gw2auth.com support


dlamkins:
This seems like it would be more appropriate to do as a ServiceModule within the existing API service since we don't need to expose anything extra in the API for modules as far as I am aware.  Luckily isn't much to move it over into a ServiceModule since there isn't much difference.

dlamkins:
A reminder that the implementation should mostly pop in behind the existing one since it's mostly only the source of the key that changes.  The rest of the progression of the tokens can be handled how it is currently (since you can get subtokens of subtokens).

Blish HUD/GameServices/Gw2WebApi/Gw2Auth/CallbackListener.cs
```diff
@@ -0,0 +1,69 @@
...
+        public void Stop() {
+            if (_listener.IsListening) {
+                _listener.Stop();
+            }
+        }
+
+        /// <summary>
+        /// Releases the resources held by the encapsulated HttpListener object.
+        /// </summary>
+        /// <seealso cref="https://docs.microsoft.com/en-us/dotnet/api/system.net.httplistener.system-idisposable-dispose?view=netframework-4.7.2"/>
```
agaertner: Needs to be
```csharp
/// <remarks>
///     <seealso href="https://docs.microsoft.com/en-us/dotnet/api/system.net.httplistener.system-idisposable-dispose?view=netframework-4.7.2"/>
/// </remarks>
```

dlamkins:
I think we've ultimately decided to not implement OAuth support.  I think it's ultimately a cool idea, but not something that users find intuitive/willing to put up with.

PR #687, opened 2022-05-08
Split close on Escape out of CanClose

Denrage:
This PR splits the option to close a window with esc out of the CanClose-Property, so a window can still be closed by the user with the X-Icon but not anymore with escape.
There are use cases in which closing with esc can close a window on coincidence way more often than on purpose and where there is no need to be able to close it with escape.
That is really apparent in the case of the AchievementTrackerModule where myself and multiple other already experienced that the current behavior is a hindrance.

### DIFF
--- a/Blish HUD/Controls/WindowBase2.cs
+++ b/Blish HUD/Controls/WindowBase2.cs
@@ -121,6 +121,12 @@ namespace Blish_HUD.Controls {
             set => SetProperty(ref _canClose, value);
         }
 
+        private bool _canCloseWithEscape = true;
+        public bool CanCloseWithEscape {
+            get => _canCloseWithEscape;
+            set => SetProperty(ref _canCloseWithEscape, value, true);
+        }
+
         private bool _canResize = false;
         /// <summary>
         /// Allows the window to be resized by dragging the bottom right corner.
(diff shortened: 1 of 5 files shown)


Blish HUD/GameServices/Input/Keyboard/KeyboardHandler.cs
```diff
@@ -229,7 +229,7 @@ public class KeyboardHandler : IInputHandler {
                         // If we found an active window, close it
                         var activeWindow = WindowBase2.ActiveWindow;
 
-                        if (activeWindow != null && activeWindow.CanClose) {
+                        if (activeWindow != null && activeWindow.CanCloseWithEscape) {
```
dlamkins: This likely still needs the check for CanClose.  This change would allow windows with this set to false still able to close with ESC.

`if (activeWindow != null && activeWindow.CanClose && activeWindow.CanCloseWithEscape)) {`

Blish HUD/Controls/WindowBase2.cs
```diff
@@ -121,6 +121,12 @@ public abstract class WindowBase2 : Container, IWindow, IViewContainer {
             set => SetProperty(ref _canClose, value);
         }
 
+        private bool _canCloseWithEscape = true;
+        public bool CanCloseWithEscape {
+            get => _canCloseWithEscape;
+            set => SetProperty(ref _canCloseWithEscape, value, true);
```
dlamkins: No need to invalidate the window when this property is updated.

dlamkins (changes requested):

dlamkins (approved):

dlamkins (approved):

PR #690, opened 2022-05-10
Set Focused property after internal logic

tentose:
`Focused` on `TextInputBase` is currently set on click or left mouse pressed. However, this happens before internal logic to place the cursor. This makes it difficult to customize the behaviour of `TextInputBase` on focus such as selecting all text on focus.

This change moves setting of the Focused property to after internal logic has run to place the cursor.

Tested locally. With this change it's now possible to set the selection range directly in the handler for the `InputFocusChanged` event.


dlamkins:
Seems like a reasonable change.  👍 

Thank you!

dlamkins (approved):

PR #694, opened 2022-05-23
Introducing: DatAssetCache

dlamkins:
Request asset textures by ID with the following new APIs:

```cs
AsyncTexture2D AsyncTexture2D.FromAssetId(int assetId);
bool AsyncTexture2D.TryFromAssetId(int assetId, out AsyncTexture2D texture);

AsyncTexture2D GameService.Content.DatAssetCache.GetTextureFromAssetId(int assetId);
bool GameService.Content.DatAssetCache.TryGetTextureFromAssetId(int assetId, out AsyncTexture2D texture);
```

Internal views and control references to raw assets have been replaced with this implementation in _most_ scenarios.

For now, the referenced textures have been kept in as some modules still reference these.  We'll likely want to mark the public method as deprecated and have Blish HUD use its own internal method when accessing the ref.dat.  This way we can wean modules off of using the textures so that they can be removed from the dat (resulting in a much smaller release package).

----

The current implementation is meant to be somewhat robust.  If the metadata file which defines which textures are available is not accessible on the web server (or the request fails), we use the one loaded from the last session.  If there is no metadata file from the last session (or it fails to load), we use the one provided in the ref.dat.

We maintain a basic token bucket for retries.  If a texture fails to load for any reason, we'll retry until the token bucket is exhausted.  If a texture fails and the token bucket (which is shared for all textures) is empty, then a broken texture image is returned instead and we stop retrying.

Textures will release on their own when no longer referenced.  A count of the **asset textures** (it does not account for _all_ textures) still loaded can now be seen in the top right.  The number will rise and fall as textures become visible on screen.


dlamkins:
Closing in favor of fixed branch #696

PR #712, opened 2022-06-29
Add Nvidia settings contingency & update About view

eksime:
This adds a contingency dialog for incompatible nvidia control panel settings - I'm unfortunately unable to test this on an AMD system, but it should just catch a missing file exception and skip the check.

The nvidia setting code is sourced from [Orbmu2k/nvidiaProfileInspector](https://github.com/Orbmu2k/nvidiaProfileInspector) as a submodule - with a netstandard 2.0 project building the service implementations into a library for use in Blish HUD - it may be preferable to fork the profile inspector instead of the submodule (?)

This also updates the about view with a scrollable box to include software licences in & adds a target for adding software licences in the csproj - again, a different solution may be preferred (?)


dlamkins:
Great!  I've gone ahead and wired up https://link.blishhud.com/nvidiasettings and will check the build directly.  Glad to have a spot for the licensing to be included as well.  😄

dlamkins:
Okay, I apologize, I know it's been a while since you put this together.

Could you help answer the following for me?
- Where can I find a list of the available key values (for example, `ESetting.MAXWELL_B_SAMPLE_INTERLEAVE_ID`) (I'm trying to make sure we've got FXAA listed as well since it doesn't seem to trigger the popup for that).
- If this checks global settings as well as program specific ones or not.

Otherwise, it does look quite good.  I like the touch you did in the about menu.  It looks nice.

eksime:
Hey, no worries, I've also been pretty busy!

> * Where can I find a list of the available key values (for example, `ESetting.MAXWELL_B_SAMPLE_INTERLEAVE_ID`) (I'm trying to make sure we've got FXAA listed as well since it doesn't seem to trigger the popup for that).

They're originally from [NvApiDriverSettings.h](https://github.com/Orbmu2k/nvidiaProfileInspector/blob/master/nspector/Native/NVAPI/NvApiDriverSettings.h) but are transformed via t4 template into [NvApiDriverSettings.cs](https://github.com/Orbmu2k/nvidiaProfileInspector/blob/master/nspector/Native/NVAPI/NvApiDriverSettings.cs) I believe the one for FXAA would be `ESetting.FXAA_ENABLE_ID` but I'd need to double check by toggling it in control panel & seeing what setting(s) get applied.

> * If this checks global settings as well as program specific ones or not.

It checks both - it checks Blish HUD's "view" of the settings, so any global options plus any that are overriden on a per-exe basis.

These lines perform the lookup for any profiles, or return an empty string indicating a global profile:
https://github.com/eksime/Blish-HUD/blob/8cfc94721b75608205326692133820b0214c32fc/Blish%20HUD/GameServices/Debug/ContingencyChecks.cs#L89-L90

I could try to separate out the about page changes to a different PR if that'd be preferable? I realise this is quite a messy change with the submodule and everything.

dlamkins:
No worries!  We can keep both. I'll test with the other settings we've seen cause this and move towards merging assume everything else looks good. Thank you!

dlamkins (approved):

PR #752, opened 2022-09-07
Fix/arcdpsservice

Tharylia:
Attempts to fix the spontaneous disconnections from the ArcDPS BlishHUD Bridge.

The prior coding relied on the SocketAsyncEventArgs class, which fires a ConnectionReset event when some actions complete async and not in sync. This state was hard to recover from.


Blish HUD/GameServices/ArcDpsService.cs
```diff
@@ -107,8 +118,37 @@ public class ArcDpsService : GameService {
...
+        private void Stop() {
+            if (this.Loaded) {
+                this._server.Stop();
+            }
+        }
+
+        /// <summary>
+        /// Restarts the socket listener for the arc dps bridge.
+        /// </summary>
+        public void Restart() {
```
dlamkins: Is this expected to be used outside of Core?  Who are the expected callers?

Tharylia: I added that in case it does actually disconnect at some point and a module dev wants to restart it immediately

Blish HUD/GameServices/ArcDpsService.cs
```diff
@@ -27,6 +33,11 @@ public class ArcDpsService : GameService {
         /// </summary>
         public bool RenderPresent { get; private set; }
 
+        /// <summary>
+        ///     Indicates if the socket listener for the arcdps service is running and arcdps send an update in the last second.
```
dlamkins: Where do we do the check via a second?  To me it looks like if we can any data we show it as running.  Is that correct or is there a place where we set it back to false if we've not received any traffic?

Obviously, it makes sense that once we're connected we would expect it to stay connected since it's not meant to be a feature that can turn off during the game session, but I suppose this doesn't give insight into it actually disconnecting (which may be harder to tell vs. just a quiet client).

Tharylia: Its calculated on the setter of HudIsActive which should be updated every frame in game. If that does not happen, RenderPresent is set to false after 1 sec. (In Update Loop)

Tharylia: This for example does cover a few edge cases where we are connected socket wise but arc does not actually send anything because some pattern can't be found after an update which disabled the arc api.
In this case the Service is effectively not running even though we are connected.

dlamkins: Ahh!  Gotcha.  Thank you!

Blish HUD/GameServices/ArcDpsService.cs
dlamkins: I would prefer that we do not expose this - I don't want modules calling this.  If a module could detect failure, then it seems reasonable to me that Core should be able to as well.  👍

Tharylia: I have removed the Restart and Stop (0 references) functions

Blish HUD/GameServices/ArcDps/SocketListener/SocketListener.cs
```diff
@@ -35,142 +46,216 @@ public sealed class SocketListener {
...
+                Logger.Error(ex, "Failed to connect socket:");
+
+                this.Stop();
             }
         }
 
-        private void ProcessReceive(SocketAsyncEventArgs e) {
-            var token = (AsyncUserToken)e.UserToken;
+        private void StartReceive(Socket socket, SocketState state = null, int? retries = null) {
+            retries ??= _retryReceiveCount;
```
greaka: Optional: If you change the _retryReceiveCount to be const, you could inline it directly in the function signature

Blish HUD/GameServices/ArcDpsService.cs
```diff
@@ -11,11 +11,17 @@
 namespace Blish_HUD {
 
     public class ArcDpsService : GameService {
+        private static readonly Logger Logger = Logger.GetLogger<ArcDpsService>();
 
         private static readonly object WatchLock = new object();
         #if DEBUG
         public static long Counter;
-        #endif
+#endif
```
greaka: the if and endif are not matching in indentation

Blish HUD/GameServices/ArcDpsService.cs
greaka: minor typo in sent

Blish HUD/GameServices/ArcDpsService.cs
```diff
@@ -153,8 +179,10 @@ public class ArcDpsService : GameService {
         }
 
         private void SocketErrorHandler(object sender, SocketError socketError) {
-            var listener = (SocketListener)sender;
-            listener.Stop();
+            // Socketlistener stops by itself.
+            Logger.Error("ArcDpsService has socket error: {0}", socketError.ToString());
```
greaka: Is `has socket error` worded correctly? I think `encountered` or something along those lines would fit better here

greaka (commented):
I love this PR! The code is much easier to understand now, only found a few nitpicks
Great work! :slightly_smiling_face:

Blish HUD/GameServices/ArcDps/SocketListener/SocketListener.cs
Tharylia: No idea why I didn't do that but I have changed that!

Blish HUD/GameServices/ArcDpsService.cs
Tharylia: Fixed

Blish HUD/GameServices/ArcDpsService.cs
Tharylia: Fixed

Blish HUD/GameServices/ArcDpsService.cs
Tharylia: Yeah, that sounds wrong. I have changed that to encountered.

dlamkins (approved):

dlamkins:
Looks good to me, now.  @greaka I appreciate you chiming in with the review.  Anything else before we merge?  😄

greaka (approved):
Off we go!

PR #754, opened 2022-09-10
add KeyBinding.BindingChanged event

agaertner:
I am dealing with saving and loading KeyBindings from settings a lot.

**Problems**
- Only the KeyAssigner control itself has an appropriate event for when the assignment changed which forbids one to reuse the SettingsView inside a custom ViewContainer for self managment and instead build the KeyAssigner.
- The SettingEntry<KeyBinding>.SettingChanged event only fires when the object itself changes.

**Solution:**
- Event that makes it possible to continuously work with the KeyBinding stored in the SettingEntry while reusing the SettingsView inside a custom ViewContainer.

### DIFF
--- a/Blish HUD/GameServices/Input/Keyboard/KeyBinding.cs
+++ b/Blish HUD/GameServices/Input/Keyboard/KeyBinding.cs
@@ -10,6 +10,10 @@ namespace Blish_HUD.Input {
     /// Allows for actions to be ran as the result of a provided key combination.
     /// </summary>
     public class KeyBinding {
+        /// <summary>
+        /// Fires when the keys of the <see cref="KeyBinding"/> are changed.
+        /// </summary>
+        public event EventHandler<EventArgs> KeysChanged; 
 
         /// <summary>
         /// Fires when the <see cref="KeyBinding"/> is triggered.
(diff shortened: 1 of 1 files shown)


Blish HUD/GameServices/Input/Keyboard/KeyBinding.cs
```diff
@@ -10,6 +10,10 @@ namespace Blish_HUD.Input {
     /// Allows for actions to be ran as the result of a provided key combination.
     /// </summary>
     public class KeyBinding {
+        /// <summary>
+        /// Fires when the keys of the <see cref="KeyBinding"/> are changed.
+        /// </summary>
+        public event EventHandler<EventArgs> KeysChanged;
```
dlamkins: Let's name this `BindingChanged`

dlamkins (changes requested):

Blish HUD/GameServices/Input/Keyboard/KeyBinding.cs
agaertner: Funny, that was how I named it before I went with KeysChanged. :)

Resolved with https://github.com/blish-hud/Blish-HUD/pull/754/commits/5a925cb6f6512cc7512b928f7092cd60d19ca057

dlamkins (approved):

dlamkins (approved):

PR #766, opened 2022-09-22
Fix ScreenNotification location

Tharylia:
This PR fixes the ScreenNotification's location being displayed in the top left instead of the center.

This happens because the ScreennNotification is created before the SpriteScreen has finished resizing resulting in a location of 0, 0.


dlamkins (approved):
Awesome - thanks!

PR #770, opened 2022-09-25
🐛fix SettingsCollection.RenderInUi not being used in constructor

agaertner:
renderInUi was not used in the simplest SettingsCollection constructor which forced us to use the longest overload everytime.

I added another overload without lazyLoaded and made all overloads reference the longest overload directly.


dlamkins (approved):
Looks good, thank you!

PR #771, opened 2022-09-28
Cleanup and semver check for prereleases.

dlamkins:


dlamkins:
https://discord.com/channels/531175899588984842/599270434642460753/1024609460167127090

It seems like a reasonable tradeoff to keep the current version as release versions should not be targeting prerelease versions of Blish HUD, anyways.

PR #787, opened 2022-10-17
Fix menu item re-aligning

Tharylia:
This PR fixes the issue with menu items not re-aligning after an menu item has been deleted in the middle.
The current state would leave a gap.


dlamkins (approved):

dlamkins:
Thanks!

Issue #631, opened 2022-03-13
Translucent overlay GUI with no way to set opacity


rootkie:
Hi, not sure if this is a bug or there is some hidden setting to change opacity.

Windows settings:
Dark, transparency effects on

![image](https://user-images.githubusercontent.com/20994542/158046322-8a96ed0f-df6e-488e-8a71-1272671725e8.png)

Log from this run:
```text
Blish HUD v0.11.7+Branch.tags-v0.11.7.Sha.0c709c76cb502d310f2964dca92ce770e533dae2
13:18:53.1950 | INFO  | Blish_HUD.Program | Running Build "Branch.tags-v0.11.7.Sha.0c709c76cb502d310f2964dca92ce770e533dae2"
13:18:53.1950 | INFO  | Blish_HUD.Program | Launched from "D:\Personal\Games\Guild Wars 2\Blish.HUD.0.11.6" with args "".
13:18:54.8147 | INFO  | Blish_HUD.Modules.Pkgs.StaticPkgRepoProvider | Getting module list...
13:18:55.5562 | INFO  | Blish_HUD.Contexts.FestivalContext | Active festival(s): ""
13:19:00.1614 | INFO  | Blish_HUD.GameIntegration.WinFormsIntegration | Blish HUD starting Guild Wars 2 with args '"-autologin"'
13:19:57.0075 | INFO  | Blish_HUD.GameIntegration.ClientTypeIntegration | Detected Guild Wars 2 client to be the Standard version.
13:19:57.0535 | INFO  | Blish_HUD.Gw2WebApiService | Could not find registered API key associated with character "Social Darwinist"
13:23:08.2095 | WARN  | Blish_HUD.WindowUtil | GetClientRect failed with error code 1400.
13:23:08.4573 | INFO  | Blish_HUD.GameIntegration.Gw2InstanceIntegration | Guild Wars 2 application has exited!
13:23:08.4573 | INFO  | Blish_HUD.OverlayService | Exiting [0]!
```

dlamkins:
Hello!  This is a bug in the most recent version.  Please try using this custom build and let us know if it resolves the issue for you or not: [Blish HUD 0.11.8-Xanthus-custom-transparency.zip](https://github.com/blish-hud/Blish-HUD/files/8240327/Blish.HUD.0.11.8-Xanthus-custom-transparency.1.1.zip).

rootkie:
[...]

PR #817, opened 2022-12-29
Remove borderless from BlishHud

kalebmcgregor:


dlamkins:
Helps to address #631.

dlamkins (approved):

PR #830, opened 2023-01-09
fix module repo filter checkbox being inverted

Cytraen:
This fixes an issue where toggling filters in the module repo would do nothing on the first toggle, and subsequent toggles would do the opposite of what is expected.


dlamkins (approved):

dlamkins:
Thank you!

PR #885, opened 2023-05-26
thread safety changes for tweener class

### LINKED ISSUE #812: Fatal on tweener (race condition)
lish_HUD.DebugService | Blish HUD encountered a fatal crash!
System.IndexOutOfRangeException: Index was outside the bounds of the array.
at System.Collections.Generic.Dictionary`2.Insert(TKey key, TValue value, Boolean add)
at Glide.Tween.TweenerImpl.AddAndRemove() in D:\a\Blish-HUD\Blish-HUD\Blish HUD\Library\Glide\Tweener.cs:line 225
at Glide.Tween.TweenerImpl.Tween[T](T target, Object values, Single duration, Single delay, Boolean overwrite) in D:\a\Blish-HUD\Blish-HUD\Bli

eksime:
[Discord Discussion](https://discord.com/channels/531175899588984842/536970543736291346/1111425360282800190)

Breaking change: no

This is an attempt to fix #812 - mostly by converting the main Dictionary object used by the tweener class to a ConcurrentDictionary and adding locks to inner list access.

It also swaps out the implementations of toadd/toremove to ConcurrentQueue instead of list,  as it's a more natural fit for the usecase.

There's also some minor cleanup of unused variables and formatting.

### DIFF
--- a/Blish HUD/Library/Glide/MemberAccessor.cs
+++ b/Blish HUD/Library/Glide/MemberAccessor.cs
@@ -1,30 +1,26 @@
-using System;
-using System.Linq.Expressions;
-using System.Reflection;
+namespace Glide {
+    using System;
+    using System.Linq.Expressions;
+    using System.Reflection;
 
-namespace Glide
-{
-    internal class MemberAccessor
-    {
-        public object Target { get; private set; }
+    internal class MemberAccessor {
         public string MemberName { get; private set; }
         public Type MemberType { get; private set; }
 
-        public object Value
-        {
-            get { return getMethod(this.Target); }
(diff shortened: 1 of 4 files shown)


eksime:
To expand on this - the tweener code is quite messy, and relies on a lot of reflection and boxing operations. In the future, if it's something you would be open to, I'd quite like to tackle reimplementing the core of it to make better use of generics for non-boxing operations on tween values & attaching the tweens to their targets using ConditionalWeakTables so that they don't prevent objects getting collected.

dlamkins (approved):

dlamkins (approved):

PR #888, opened 2023-05-28
Added overloads to specify own logger names

Tharylia:
This PR aims to enable module developer to specify their own logger names.

My example for the need of this implementation:

The module Event Table can have an "unlimited" amount of event areas on the screen. Each area logs specific data.
Currently there is no way to know from which area this log statement comes.
The only alternative currently would be to prefix the name of the area on each logger call. This is prone the be forgotten and not the best method.

The new method allows me to specify my own name and results in logs like this:

`14:31:12.9626 | DEBUG | Estreya.BlishHUD.EventTable.Controls.EventArea - Main | Added event The Path to Ascension with occurence 28.05.2023 13:30:00` ( Note the name "Main" after the type )

The logger can be created like this:
`Logger.GetLogger<EventArea>($"{typeof(EventArea).FullName} - {this.Configuration.Name}")`

The new implementation does not break existing loggers in any way.

### DIFF
--- a/Blish HUD/GameServices/Debug/Logger.cs
+++ b/Blish HUD/GameServices/Debug/Logger.cs
@@ -7,23 +7,33 @@ namespace Blish_HUD {
 
         #region Static
 
-        private static readonly ConcurrentDictionary<Type, Logger> _loggers = new ConcurrentDictionary<Type, Logger>();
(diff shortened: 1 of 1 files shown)


dlamkins:
I'm not sure I see any major value in this change.  You can wrap the logger and prefix your messages with the area if this is necessary, but this changes the structure of the log messages and seems like a very specific use case.

Tharylia:
Alright, I will do that. 👍

PR #907, opened 2023-08-20
style: add menu click sounds from the game's menu.

agaertner:
Menu click sounds for the MenuItems.
1:1 like of the original menu of the game.

Not playing for checkable menu items.

The sounds will end up in many modules otherwise. Menus like the following would get a huge boost in feel.

image

## Is this a breaking change?
No


dlamkins:
I like this.  I see the files are added to both ref and content, though.  I'm looking to slowly start reducing what we have in content and move it to the ref.  Would it be possible to have them only in the ref?

agaertner:
> I like this. I see the files are added to both ref and content, though. I'm looking to slowly start reducing what we have in content and move it to the ref. Would it be possible to have them only in the ref?

Sure. I will check.

dlamkins (approved):

PR #923, opened 2023-10-10
Load modules sorted by dependencies

Tharylia:
This PR implements module loading sorted by dependencies. 
This is needed for contexts to work properly if the context class is contained inside the module.

## Discussion Reference

https://discord.com/channels/531175899588984842/599270434642460753/1160999293361061979

## Is this a breaking change?

No

### DIFF
--- a/Blish HUD/GameServices/ModuleService.cs
+++ b/Blish HUD/GameServices/ModuleService.cs
@@ -320,16 +320,9 @@ namespace Blish_HUD {
         }
 
         private void Gw2Instance_Gw2Started(object sender, EventArgs e) {
-            var resolvedDependencyOrder = SortByDependency(_modules, module => {
-                var dependencyModules = _modules.Where(m => {
-                    // Get all modules which have a dependency on the current module
-                    return module.Manifest?.Dependencies?.Any(dm => !dm.IsBlishHud && dm.Namespace == m.Manifest.Namespace) ?? false;
-                });
+            var resolvedDependencyOrder = SortByDependencies(_modules).ToList();
 
-                return dependencyModules;
-            }).ToList();
-
-            Logger.Debug($"Resolved dependency order: {string.Join(", ", resolvedDependencyOrder.Select(x => x.Manifest.Namespace).ToArray())}");
+            Logger.Debug($"Resolved load dependency order: {string.Join(", ", resolvedDependencyOrder.Select(x => x.Manifest.Namespace).ToArray())}");
 
             foreach (var module in resolvedDependencyOrder) {
                 if (module.State.Enabled) {
@@ -338,12 +331,19 @@ namespace Blish_HUD {
             }
         }
 
-        private static IEnumerable<ModuleManager> SortByDependency(IEnumerable<ModuleManager> source, Func<ModuleManager, IEnumerable<ModuleManager>> dependencies) {
+        private static IEnumerable<ModuleManager> SortByDependencies(IEnumerable<ModuleManager> modules) {
             var sorted = new List<ModuleManager>();
             var visited = new HashSet<ModuleManager>();
 
-            foreach (var item in source) {
-                VisitDependency(item, visited, sorted, dependencies);
+            foreach (var module in modules) {
+                VisitDependency(module, visited, sorted, m => {
+                    var dependencyModules = modules.Where(m => {
+                        // Get all modules which have a dependency on the current module
+                        return m.Manifest?.Dependencies?.Any(md => !md.IsBlishHud && md.Namespace == m.Manifest.Namespace) ?? false;
+                    });
+
+                    return dependencyModules;
+                });
             }
 
             return sorted;
@@ -432,7 +432,12 @@ namespace Blish_HUD {
         protected override void Unload() {
             GameService.GameIntegration.Gw2Instance.Gw2Started -= Gw2Instance_Gw2Started;
 
-            foreach (var module in _modules) {
+            // We need to unload modules in reverse order as modules could need to execute context methods on unload.
+            var resolvedDependencyOrder = SortByDependencies(_modules).Reverse().ToList();
+
+            Logger.Debug($"Resolved unload dependency order: {string.Join(", ", resolvedDependencyOrder.Select(x => x.Manifest.Namespace).ToArray())}");
+
+            foreach (var module in resolvedDependencyOrder) {
                 if (module.Enabled) {
                     try {
                         Logger.Info("Unloading module {module}.", module.Manifest.GetDetailedName());
--- a/Blish HUD/GameServices/ModuleService.cs
+++ b/Blish HUD/GameServices/ModuleService.cs
@@ -300,8 +300,10 @@ namespace Blish_HUD {
                     Logger.Warn("Failed to load module from path {modulePath}.", ApplicationSettings.Instance.DebugModulePath);
                 }
 
-                debugModule.State.Enabled = true;
-                _modules.Add(debugModule); // Don't start directly and let it load normally with dependency order
+                if (debugModule != null) {
+                    debugModule.State.Enabled = true;
+                    _modules.Add(debugModule); // Don't start directly and let it load normally with dependency order
+                }
             }
 
             HandleRefLoading();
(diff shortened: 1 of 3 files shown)


Blish HUD/GameServices/ModuleService.cs
```diff
@@ -300,7 +300,10 @@ public class ModuleService : GameService {
                     Logger.Warn("Failed to load module from path {modulePath}.", ApplicationSettings.Instance.DebugModulePath);
                 }
 
-                debugModule?.TryEnable();
+                if (debugModule != null) {
+                    debugModule.State.Enabled = true;
```
dlamkins, line 304: This impacts the state even after we're no longer using `--module`, which is a change in behavior we'll just want to keep in mind.

Before: when no longer using `--module`, a module would adhere to the state it was previously in.
After: when no longer using `--module`, the module will be enabled until disabled again.

Blish HUD/GameServices/ModuleService.cs
```diff
@@ -317,20 +320,61 @@ public class ModuleService : GameService {
...
+            if (resolvedDependencyOrder != null) {
+                Logger.Debug($"Resolved dependency order: {string.Join(", ", resolvedDependencyOrder.Select(x => x.Manifest.Namespace).ToArray())}");
+
+                foreach (var module in resolvedDependencyOrder) {
+                    if (module.State.Enabled) {
+                        module.TryEnable();
+                    }
+                }
+            }else {
+                Logger.Error($"Could not resolve dependencies of any module.");
```
dlamkins: If this is truly an unexpected error, we should provide more information about the error.  Otherwise, the entry in Sentry will likely be useless.

Blish HUD/GameServices/ModuleService.cs
```diff
@@ -317,20 +320,61 @@ public class ModuleService : GameService {
...
+            }
+        }
+
+        private static IEnumerable<ModuleManager> SortByDependency(IEnumerable<ModuleManager> source, Func<ModuleManager, IEnumerable<ModuleManager>> dependencies, bool throwOnCycle = false, bool logOnCycle = true) {
+            var sorted = new List<ModuleManager>();
+            var visited = new HashSet<ModuleManager>();
+
+            foreach (var item in source)
+                VisitDependency(item, visited, sorted, dependencies, throwOnCycle, logOnCycle);
```
dlamkins: Include braces around the foreach contents.

Blish HUD/GameServices/ModuleService.cs
```diff
@@ -317,20 +320,61 @@ public class ModuleService : GameService {
...
+
+            return sorted;
+        }
+
+        private static void VisitDependency(ModuleManager item, HashSet<ModuleManager> visited, List<ModuleManager> sorted, Func<ModuleManager, IEnumerable<ModuleManager>> dependencies, bool throwOnCycle, bool logOnCycle) {
+            if (!visited.Contains(item)) {
+                visited.Add(item);
+
+                foreach (var dep in dependencies(item))
```
dlamkins: Include braces around the foreach contents.

Blish HUD/GameServices/ModuleService.cs
```diff
@@ -317,20 +320,61 @@ public class ModuleService : GameService {
...
+                    VisitDependency(dep, visited, sorted, dependencies, throwOnCycle, logOnCycle);
+
+                sorted.Add(item);
+            } else {
+                if (!sorted.Contains(item)) {
+                    if (logOnCycle) Logger.Warn($"Cyclic dependency found: {item.Manifest.Namespace}");
```
dlamkins: Use `.GetDetailedName()` when logging if we already have the manifest.

Blish HUD/GameServices/ModuleService.cs
```diff
@@ -317,20 +320,61 @@ public class ModuleService : GameService {
...
+                    if (throwOnCycle) throw new Exception($"Cyclic dependency found: {item.Manifest.Namespace}");
```
dlamkins: If this throws, what happens?

Blish HUD/GameServices/Modules/ModuleManager.cs
```diff
@@ -38,9 +39,11 @@ public class ModuleManager : IDisposable {
         /// </summary>
         public bool Enabled { get; private set; }
 
+        private bool _dependenciesAvailable => Manifest.Dependencies.TrueForAll(d => d.GetDependencyDetails().CheckResult == ModuleDependencyCheckResult.Available);
```
dlamkins: This walks like a function.  Talks like a function.  It should just be written as a function.

Blish HUD/GameServices/Modules/ModuleManager.cs
```diff
@@ -69,6 +72,13 @@ public class ModuleManager : IDisposable {
              || GameService.Module.ModuleIsExplicitlyIncompatible(this)) // Module is on the explicit "incompatibile" list.
                 return false;
 
+            if (!this.DependenciesMet) {
+                Logger.Warn($"Module {this.Manifest.Namespace} can not be loaded as not all dependencies are available. Missing: {string.Join(", ", this.GetMissingDependencies().Select(md => $"{md.Namespace} ({md.GetDependencyDetails().CheckResult})"))}");
```
dlamkins: Use `.GetDetailedName()` when logging if we already have the manifest.

Blish HUD/GameServices/Modules/ModuleManager.cs
```diff
@@ -69,6 +72,13 @@ public class ModuleManager : IDisposable {
...
+                return false;
+            } else if (!_dependenciesAvailable && this.State.IgnoreDependencies) {
+                Logger.Warn($"Module {this.Manifest.Namespace} has not all dependencies available but is set to ignore. Missing: {string.Join(", ", this.GetMissingDependencies().Select(md => $"{md.Namespace} ({md.GetDependencyDetails().CheckResult})"))}");
```
dlamkins: Use `.GetDetailedName()` when logging if we already have the manifest.

Blish HUD/Properties/launchSettings.json
```diff
@@ -5,7 +5,7 @@
     },
     "PowerShell HUD": {
       "commandName": "Project",
-      "commandLineArgs": "-p powershell -w ConsoleWindowClass"
+      "commandLineArgs": " --process WindowsTerminal --window \"Windows PowerShell\" "
```
dlamkins, line 8: This appears to be completely unrelated.

dlamkins:
I did a first-pass lookover.

Other things that need answering:

1. What is the expected behavior when a module is allowed to start prior to GW2?  How does this affect the load order?
2. What happens when a user manually disables a dependency?
3. What UI feedback is provided to the user to explain a module not enabled because of a dependency issue?  Is it just the dependency section?

Regarding your use case, it's still unclear to me why the dependent module has a different implementation of an assembly being loaded, making the load order important.  Can you provide a minimal code example that exhibits this issue so that I can see it?
1. Load order doesn't explicitly ensure that a module's assemblies are loaded prior to another in the event that the module doesn't attempt to use an assembly first.

Blish HUD/GameServices/ModuleService.cs
```diff
@@ -317,20 +320,61 @@ public class ModuleService : GameService {
...
+            var resolvedDependencyOrder = SortByDependency(_modules, module => {
+                var dependencyModules = _modules.Where(m => {
+                    // Get all modules which have a dependency on the current module
+                    return module.Manifest?.Dependencies?.Any(dm => !dm.IsBlishHud && dm.Namespace == m.Manifest.Namespace) ?? false;
+                });
+
+                return dependencyModules;
+            }).ToList();
+
+            if (resolvedDependencyOrder != null) {
```
dlamkins: In what scenario could resolvedDependencyOrder be null?

Blish HUD/GameServices/ModuleService.cs
Tharylia: I think I can remove that if again. I only added that because SonarCloud complained I should add it. After I added it, it complained that it is useless..

Blish HUD/GameServices/ModuleService.cs
Tharylia: I was undecided what to do with that. I will probably remove it as its not really needed.
I already tested it without the throw. It just does not load both modules

Blish HUD/Properties/launchSettings.json
Tharylia: That's true.. not sure how that got in here. How do I remove that again?

Blish HUD/GameServices/ModuleService.cs
Tharylia: I have no idea. SonarCloud wanted it there

Tharylia:
> I did a first-pass lookover.
> 
> Other things that need answering:
> 
> 1. What is the expected behavior when a module is allowed to start prior to GW2?  How does this affect the load order?
> 2. What happens when a user manually disables a dependency?
> 3. What UI feedback is provided to the user to explain a module not enabled because of a dependency issue?  Is it just the dependency section?

2. Nothing, the same as prior to this PR. The dependent module needs to check for this case. In case of contexts: The state of Expired.
3. No change. It is still the same as prior to the PR. Its only visible in the dependency section. I considered ScreenNotifications but I'm not sure if we want to potentially spam users.

Blish HUD/GameServices/ModuleService.cs
Tharylia: Removed.

Tharylia:
[BlishHUD Module Load Order Test.zip](https://github.com/blish-hud/Blish-HUD/files/12859809/BlishHUD.Module.Load.Order.Test.zip)


ModuleB registers a context and ModuleA is dependent on B as it wants to use it.

Steps:
- Build both modules and place in modules folder
- Start blish and start ModuleB and then ModuleA
- See log for `INFO  | AnotherTest.ModuleA | Found context of module B.` entry
- Restart blish and wait for module A to load completely ( takes 30 seconds )
- See log for `INFO  | AnotherTest.ModuleA | Could not find context of module B after 30 seconds.`
- Close blish
- Prefix the module file "LoadOrderTest.bhm" with "00_" to make it first in list and simulate a different load order
- Start blish
- See log for `INFO  | AnotherTest.ModuleA | Found context of module B.`

Blish HUD/GameServices/ModuleService.cs
dlamkins: SonarCloud is truly hilarious sometimes.  :D

PR #947, opened 2024-03-11
Arcdps Integration Rewrite

Denrage:
## Discussion Reference

https://discord.com/channels/531175899588984842/536970543736291346/1179936940477394945

## Is this a breaking change?
No

This PR updates the ArcDPS-Service of Blish-Hud. It is in preparations for the rewrite of the arcdps bridge update.
The socket is completely rewritten from scratch with performance and readability in mind. It is also way simpler which should make the code less prone to errors.
There is a new V2 Service as discussed that has a simpler API to interact with the bridge. All calls to third parties are also run in parallel and away from the socket thread, so it wont block if someone has a long running handlet.
The V1 Service now wraps the V2 service and should behave exactly the same as before (tested with CustomCombatText).
The V2 Service checks for the open port of the V2 arcdps bridge dll and if it cant make a connection to it, it will create a V1 client that is still fully supported.

The goal was that everything existing that is using the ArcDPS-Service doesn't realize that everything changed under the hood and everyone with an old arcdps bridge dll won't notice any problems either that didn't exist bf the rewrite


dlamkins (approved):
Code looks good.  Seems to successfully remain backwards compatible after doing some testing with different modules.

Thanks!

PR #951, opened 2024-04-02
Target Framework & Packages Update

M3C3I:
## Key Points
All packages now use the latest version.
Target Framework has now been set to net8.0-windows.
SemVer was changed to SemanticVersioning due to that being the actual package that was included originally
ex.Call.Response.ReasonPhrase updated to ex.Call.Response.ResponseMessage since ReasonPhrase was replaced in newer versions

## Discussion Reference
https://discord.com/channels/531175899588984842/536970543736291346/123123123123123
(Awaiting approval to discuss in the channel)

## Is this a breaking change?
No


dlamkins:
Thanks for reaching out.  As discussed, we won't be merging this PR at this time, but we appreciate it.  This change would cause several breaking changes for our modules.

PR #955, opened 2024-04-10
feat: add VolumeChanged event

agaertner:
continuously call GetVolume to let modules react easier to either a game volume or Blish HUD volume setting change.

## Is this a breaking change?
No

### DIFF
--- a/Blish HUD/GameServices/GameIntegration/AudioIntegration.cs
+++ b/Blish HUD/GameServices/GameIntegration/AudioIntegration.cs
@@ -13,6 +13,8 @@ namespace Blish_HUD.GameIntegration {
     public sealed class AudioIntegration : ServiceModule<GameIntegrationService> {
         private static readonly Logger Logger = Logger.GetLogger<AudioIntegration>();
 
+        public event EventHandler<ValueEventArgs<float>> VolumeChanged;
+
         public enum Devices {
             [Description("GW2 Output Device")]
             Gw2OutputDevice,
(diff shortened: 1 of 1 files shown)


Blish HUD/GameServices/GameIntegration/AudioIntegration.cs
```diff
@@ -42,14 +44,21 @@ public enum Devices {
...
         /// </summary>
-        public float Volume => _volume ??= GetVolume();
+        public float Volume {
+            get => _volume ??= GetVolume();
+            set {
+                if (_volume != value) {
+                    _volume = value;
+                    VolumeChanged?.Invoke(this, new ValueEventArgs<float>(value));
+                }
+            }
```
dlamkins: This is currently marked public.  if this is regularly being updated by core, what is the use case to set it outside of core?  Should this be private or internal?

agaertner: First, backward compatibility. 
But also I'd like to have access to Volume outside the event for calculations.
I use this Volume as a master volume for when I setup volume slider settings. I calculate a normalized volume based off of it.

The event is just so I don't need to continuously check the volume in modules (comfortability not having todo that in every module's Update) and can react to a change in Volume to adjust the volume of sounds still running (eg. music).

dlamkins: Oh, I meant the setter specifically, which this did not have before.

agaertner: Oh true. The setter should be private. An oversight. Will update soon!

dlamkins: Thanks!

dlamkins (approved):

PR #963, opened 2024-04-27
Feature/runtime fonts

agaertner:
## Discussion Reference
https://discord.com/channels/531175899588984842/536970543736291346/1233539611368751185

## Is this a breaking change?
No.

I outcommented the ranges that crash for users that do not have a specific .NET Runtime installed and run BlishHUD only with the windows shipped one.

I let @teebee-git  (Discord: tommyball) test each range one by one.

As a final test I also had him install .NET Runtime 8 and test a version with all the ranges active which worked.

There are now two options: Either we recommend to people to install the latest .NET Runtime somewhere visible so we are able to use the whole charset of gw2 or stick with the ranges that work even without it.


KirillSerogodsky:
Need to be able to pass own CharacterRange, for example for Cyrillic and Asian fonts

dlamkins:
In fact, it might be best if the caller always has to specify the ranges. 

That can also help reduce the memory needed - especially if a module only needs to cover a small range (e.g. only needing numbers for a stopwatch module).

agaertner:
> In fact, it might be best if the caller always has to specify the ranges.
> 
> That can also help reduce the memory needed - especially if a module only needs to cover a small range (e.g. only needing numbers for a stopwatch module).

predefined like that?

dlamkins:
Yes, looks good. 🙂

dlamkins (approved):

greaka (approved):

PR #970, opened 2024-05-29
Fixed ArcDPS PlayerAdded and PlayerRemoved events 

JoeMylinn:
Fixes several issues with the ArcDPS Rewrite, which caused PlayerAdded and PlayerRemoved events to not work properly anymore.

### DIFF
--- a/Blish HUD/GameServices/ArcDps/V2/CommonFields.cs
+++ b/Blish HUD/GameServices/ArcDps/V2/CommonFields.cs
@@ -18,9 +18,9 @@ namespace Blish_HUD.GameServices.ArcDps.V2 {
         ///     Contains every player in the current group or squad.
         ///     Key: Character Name, Value: Account Name
         /// </summary>
-        public IReadOnlyDictionary<string, Player> PlayersInSquad => _playersInSquad;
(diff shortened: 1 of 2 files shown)


Blish HUD/GameServices/ArcDps/V2/CommonFields.cs
```diff
@@ -18,9 +18,9 @@ public class CommonFields {
         ///     Contains every player in the current group or squad.
         ///     Key: Character Name, Value: Account Name
         /// </summary>
-        public IReadOnlyDictionary<string, Player> PlayersInSquad => _playersInSquad;
+        public IReadOnlyDictionary<ulong, Player> PlayersInSquad => _playersInSquad;
```
dlamkins, line 21: Can I get context on the type change?  The `summary` above still suggests that the key is a Character Name.

dlamkins (approved):

Blish HUD/GameServices/ArcDps/V2/CommonFields.cs
JoeMylinn: The type change was necessary due to ArcDPS not sending the character name anymore when an agent gets removed, only the Agent ID. I updated the summary to reflect that change.

PR #985, opened 2024-09-25
Sort module menu items

Tharylia:
This PR aims to add sorting to the module menu items.
In the current implementation it will sort the modules by the enabled state (desc) and then by name (asc).
This will result in disabled modules being at the bottom of the list and enabled at the top. Both sorted by name (asc).

## Discussion Reference

https://discord.com/channels/531175899588984842/534490472224391169/1219116477999743056

## Is this a breaking change?

No

### DIFF
--- a/Blish HUD/GameServices/ModuleService.cs
+++ b/Blish HUD/GameServices/ModuleService.cs
@@ -326,14 +326,23 @@ namespace Blish_HUD {
 
         private          MenuItem                            _rootModuleSettingsMenuItem;
         private readonly Dictionary<MenuItem, ModuleManager> _moduleMenus = new Dictionary<MenuItem, ModuleManager>();
+
+        private IOrderedEnumerable<KeyValuePair<MenuItem, ModuleManager>> GetSortedMenuItems() {
+            return _moduleMenus.OrderBy(c => c.Value.Manifest.Name);
+        }
         
         private void RegisterModuleMenuInSettings(ModuleManager moduleManager) {
             var moduleMi = new ModuleMenuItem(moduleManager) {
-                BasicTooltipText = moduleManager.Manifest.Description,
-                Parent           = _rootModuleSettingsMenuItem
+                BasicTooltipText = moduleManager.Manifest.Description
             };
 
             _moduleMenus.Add(moduleMi, moduleManager);
+
+            _rootModuleSettingsMenuItem.ClearChildren();
+
+            foreach (var menuItemPair in this.GetSortedMenuItems()) {
+                menuItemPair.Key.Parent = _rootModuleSettingsMenuItem;
+            }
         }
 
         private void UnregisterModuleMenuInSettings(ModuleManager moduleManager) {
@@ -342,7 +351,7 @@ namespace Blish_HUD {
                     _moduleMenus.Remove(moduleMenuPair.Key);
 
                     MenuItem toSelect = moduleMenuPair.Key.Selected
-                        ? _moduleMenus.FirstOrDefault().Key ?? _rootModuleSettingsMenuItem
+                        ? this.GetSortedMenuItems().FirstOrDefault().Key ?? _rootModuleSettingsMenuItem
                         : null;
 
                     moduleMenuPair.Key.Parent = null;
--- a/Blish HUD/GameServices/ModuleService.cs
+++ b/Blish HUD/GameServices/ModuleService.cs
@@ -328,7 +328,15 @@ namespace Blish_HUD {
         private readonly Dictionary<MenuItem, ModuleManager> _moduleMenus = new Dictionary<MenuItem, ModuleManager>();
 
         private IOrderedEnumerable<KeyValuePair<MenuItem, ModuleManager>> GetSortedMenuItems() {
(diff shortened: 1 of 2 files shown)


Blish HUD/GameServices/ModuleService.cs
```diff
@@ -326,14 +326,27 @@ private void Gw2Instance_Gw2Started(object sender, EventArgs e) {
...
+            return _moduleMenus.OrderByDescending(c => c.Value.Enabled).ThenBy(c => c.Value.Manifest.Name);
+        }
+
+        internal void SortMenuItems() {
+            _rootModuleSettingsMenuItem.ClearChildren();
+
+            foreach (var menuItemPair in this.GetSortedMenuItems()) {
+                menuItemPair.Key.Parent = _rootModuleSettingsMenuItem;
+            }
+        }
```
Flyga-M, line 340: We might want to think about implementing a way that `Container` children can be reordered, so this removing and re-adding becomes unnecessary. That is out of the scope of this PR though.

Blish HUD/GameServices/ModuleService.cs
```diff
@@ -342,7 +355,7 @@ private void UnregisterModuleMenuInSettings(ModuleManager moduleManager) {
                     _moduleMenus.Remove(moduleMenuPair.Key);
 
                     MenuItem toSelect = moduleMenuPair.Key.Selected
-                        ? _moduleMenus.FirstOrDefault().Key ?? _rootModuleSettingsMenuItem
+                        ? this.GetSortedMenuItems().FirstOrDefault().Key ?? _rootModuleSettingsMenuItem
                         : null;
```
Flyga-M, line 359: Do I assume correctly, that the idea here is to select the topmost module, when a module is uninstalled from its module page?

If we want to keep this behavior it might just be enough to use the `_rootModuleSettingsMenuItem.Children`, because it is already ordered. It could look something like this:
```csharp
MenuItem toSelect = moduleMenuPair.Key.Selected
    ? _rootModuleSettingsMenuItem.Children.FirstOrDefault() as MenuItem ?? _rootModuleSettingsMenuItem
    : null;
```
Of course setting the `moduleMenuPair.Key.Parent = null` would have to be moved in front of this.

After testing a bit, I'm not sure this behavior makes so much sense in the first place. In my opinion it would be less disruptive if the next closest (prev. or next) module would be selected instead. But this would mean that we have to keep the `IOrderedEnumerable` in memory or iterate over it and I'm not sure that effort would be worth the result.

I'd appreciate your input on this!

Blish HUD/GameServices/Modules/ModuleManager.cs
```diff
@@ -113,6 +113,8 @@ public bool TryEnable() {
             this.State.Enabled = this.Enabled;
             GameService.Settings.Save();
 
+            GameService.Module.SortMenuItems();
```
Flyga-M, line 116: Although this gets the job done just fine, I feel like the `ModuleManager` of a specific module shouldn't be responsible for sorting the menu items of the `ModuleService`. (The same applies to `ModuleManager.Disable()` method of course.)

It would add a bit of overhead, but another solution might be to subscribe to the `ModuleManager.ModuleEnabled` and `ModuleManager.ModuleDisabled` events inside of `ModuleService.RegisterModuleMenuInSettings` and unsubscribe in the corresponding `ModuleService.UnregisterModuleMenuInSettings`.

This way the `ModuleManager.SortMenuItems()` method could be kept private.

I'm not sure this is worth the extra lines of code, any additional thoughts are appreciated!

Flyga-M (approved):
It'd be great to finally have this in Blish HUD. No more searching through a huge (seemingly) unsorted list of modules. Thank you for moving this forward!

The PR is nice and succinct. I've left a few minor questions and suggestions, but approved the changes.

Blish HUD/GameServices/ModuleService.cs
Tharylia: I didn't want to make that big of changes to one of the most used classes in core. Thats why I settled on a solution that gets the job done with the smallest chance of breaking any modules.

Blish HUD/GameServices/ModuleService.cs
Tharylia: Yes, IIRC this will just select the topmost icon if the existing selection is deleted. Any changes how to handle that more elegant are welcome.

Blish HUD/GameServices/Modules/ModuleManager.cs
Tharylia: I feel the same way about that. But it already does similar stuff like the Save() above. I kept it in line with what I saw in core in other places.
Changes to make the method private and handling that via events is welcome if it can be ensured that (un-)subscribing to the events is handled correctly.

Flyga-M: I see what you mean.
It's probably out of scope of this PR to move the `Settings.Save()` as well. I'll try to have a look at it and see if I can come up with a neat solution.

In the meanwhile I think this PR can safely be merged.

dlamkins (approved):

greaka (approved):

PR #997, opened 2025-01-10
Fix incorrect calc width in WrapTextSegment

KirillSerogodsky:
MeasureString incorrectly calculates space width.

`GameService.Content.DefaultFont16.MeasureString(" ").Width`  = 3
there's no problem with 
`GameService.Content.DefaultFont16.MeasureString("_").Width` = 21

## Discussion Reference
No

## Is this a breaking change?

No


KirillSerogodsky:
before
image
image
after
image
image

It's not perfect, but it works better

dlamkins:
Could you please share some more comparison screenshots of other wrapped text in Blish HUD (and perhaps in a module or two)?  I want to make sure this doesn't look weird with anything existing.

Blish HUD/_Utils/DrawUtil.cs
```diff
@@ -47,7 +47,7 @@ private static string WrapTextSegment(BitmapFont spriteFont, string text, float
             string[] words      = text.Split(' ');
             var      sb         = new StringBuilder();
             float    lineWidth  = 0f;
-            float    spaceWidth = spriteFont.MeasureString(" ").Width;
+            float    spaceWidth = spriteFont.MeasureString("_").Width;
```
sliekens, line 50: Perhaps a more conservative value like 4 or 5 can be used because now I see the text wraps too early and it leaves a lot of empty space at the end of a line.

```diff
-float spaceWidth = spriteFont.MeasureString("_").Width;
+float spaceWidth = 4;
```

KirillSerogodsky:
I found a better solution

DefaultFont12
image
DefaultFont14
image
DefaultFont16
image
DefaultFont18
image
DefaultFont32
image

KirillSerogodsky:
Module list before
image
image

Mystic Crafting before
image
image
image

Module info before
image

Pathing markers before
image

KirillSerogodsky:
Module list after
image
image

Mystic Crafting after
image
image
image

Module info after
image

Pathing markers after
image

KirillSerogodsky:
Legendary crafting before
image
after
image

KirillSerogodsky:
Farming Tracker before
image
after
image

Flyga-M:
While this seems to fix the issue well enough, I think it does make sense to implement a custom `MeasureString(..)` method (in the future?) that uses `XAdvance` instead of `Width` to determine glyph sizes.

`MeasureString(..)` is just used in so many places, that a change like that would affect many places in Core.

PR #999, opened 2025-01-18
Add the ability to stop mouse event propagation

sliekens:
This PR adds the ability to stop click events (or any other mouse event) from propagating.

```diff
private void OnClick(object sender, MouseEventArgs e) {
    Console.WriteLine("Click was handled here");
+    e.StopPropagation();
}
```

This is useful when a click should be handled by a node and then stopped without propagating all the way down to leaf nodes.

An example where this is useful is in scrollable accordions. Previously, clicking the header to collapse the panel would also propagate the click to controls behind the header.

The fix for clicks on Panel headers is included in the PR.

**Before**

https://github.com/user-attachments/assets/cc088e66-29db-4536-8dba-22e8a66c2324

**After**

https://github.com/user-attachments/assets/2d7c5916-4337-4f00-9f6f-5bd6e2f7b3ac

## Discussion Reference
https://discord.com/channels/531175899588984842/536970543736291346/1330148945749475358

## Is this a breaking change?

I don't expect this change to break any modules, although I marked 2 methods as Obsolete because they did not support this change. I added new overloads to support this change. The old methods will continue to work, just without the ability to stop propagation, i.e. calling `StopPropagation()` will do nothing.

### DIFF
--- a/Blish HUD/Controls/Container.cs
+++ b/Blish HUD/Controls/Container.cs
@@ -210,12 +210,16 @@ namespace Blish_HUD.Controls {
             }
         }
 
-        public override Control TriggerMouseInput(MouseEventType mouseEventType, MouseState ms) {
+        public override Control TriggerMouseInput(MouseEventArgs args, MouseState ms) {
             Control thisResult  = null;
             Control childResult = null;
 
             if (CapturesInput() != CaptureType.None) {
-                thisResult = base.TriggerMouseInput(mouseEventType, ms);
+                thisResult = base.TriggerMouseInput(args, ms);
+            }
+
+            if (args.PropagationStopped) {
+                return thisResult;
             }
 
             List<Control>               children        = _children.ToList();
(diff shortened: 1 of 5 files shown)


sliekens:
Last minute change: also stop propagating other mouse inputs like `MouseMove` from Panel headers


https://github.com/user-attachments/assets/5e50b204-2ef7-4065-880c-10607e505b3c

dlamkins:
The change seems reasonable.  I am surprised that the clicks are able to propagate _through_ the panel header, though.  🤔  The controls are within its content region (which was moved down by the panel header, I assume).  Even without this, it kinda seems like a bug that this would occur (I wouldn't expect to have to actively detect and prevent this scenario).

sliekens:
@dlamkins it's been a while but as I recall, when the panel is scrollable, the content can slide up behind the header, where it still receives mouse events.

https://github.com/user-attachments/assets/f9cc99d7-7064-4bf0-91b1-3b158c485b5e

PR #1001, opened 2025-02-05
Added ArcdpsListener-Removal & NpcChatter from UE

### LINKED ISSUE #625: Integrate with ArcDPS Unofficial Extras
Following [arcdps-bhud's issue](https://github.com/blish-hud/arcdps-bhud/issues/2), it would be great to handle the new data the bridge sends, and exposing them in the ArcDPS service

Denrage:
## Discussion Reference
Added the possibility to remove a listener with disposing it
Split ChatMessageType into SquadMessage & NpcMessage with the new Information that UnofficialExtras will give us.(Made possible by @KirillSerogodsky) 
## Is this a breaking change?
No, old interface is marked as obsolet and cant be removed as a listener
 The UE interface was not able to be used bc the new bhud-bridge is not released, therefor it has no function and no one should have it in use

closes #625


Blish HUD/GameServices/ArcDps/V2/Models/Unofficial Extras/NpcMessageInfo.cs
```diff
@@ -0,0 +1,11 @@
+using System;
+
+namespace Blish_HUD.GameServices.ArcDps.V2.Models.UnofficialExtras {
+    public struct NpcMessageInfo {
+        public string CharacterName { get; set; }
+
+        public string Message { get; set; }
+        
+        public ulong TimeStamp { get; set; }
```
Flyga-M, line 13:
```suggestion
        /// <summary>
        /// Time since epoch in nanoseconds.
        /// This can be used to sort messages, when they are out of order.
        /// </summary>
        public ulong TimeStamp { get; set; }
```
I feel like it's helpful to document the `TimeStamp`, because the different time formats used all over ArcDPS and arcdps_unofficial_extras can quickly become confusing.

Blish HUD/GameServices/ArcDps/V2/CommonFields.cs
```diff
@@ -41,7 +41,7 @@ public void Activate() {
             if (_enabled) return;
 
             _enabled = true;
-            GameService.ArcDpsV2.RegisterMessageType<CombatCallback>(MessageType.CombatEventArea, CombatHandler);
+            GameService.ArcDpsV2.RegisterMessageType<CombatCallback>(new ArcDpsMessageListener<CombatCallback>(MessageType.CombatEventArea, CombatHandler));
```
Flyga-M, line 44: Is there a way to make use of the new listener removal feature here? Or is this not needed, because the `ArcDpsServiceV2` (and therefor the `CommonFields` instance) usually (or always?) lives as long as Blish HUD itself?

If it's useful, it could be implemented by adding a `Deactivate()` method and calling it in `ArcDpsServiceV2.Unload()`.

This question applies in the same way to the `ArcDpsService.SubscribeToCombatEventId(..)` and `ArcDpsService.Initialize()` as well as `ArcDpsServiceV2.RegisterMessageType(..)` (for the obsolete versions) and `ArcDpsServiceV2.Start(..)` methods.

Blish HUD/GameServices/ArcDps/V2/IArcDpsMessageListener.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
```
Flyga-M:
```suggestion

```

Blish HUD/GameServices/ArcDps/V2/IArcDpsMessageListener.cs
```diff
@@ -0,0 +1,17 @@
...
+using System.Threading.Tasks;
+
+namespace Blish_HUD.GameServices.ArcDps.V2 {
+    public interface IArcDpsMessageListener<T> : IDisposable
+        where T : struct {
+        event Action<IArcDpsMessageListener<T>> Disposing;
+
+        MessageType MessageType { get; }
+
+        Task HandleAsync(T message, CancellationToken ct);
```
Flyga-M, line 12: Nitpick: It wasn't immediately clear to me what the `HandleAsync(..)` method does just by its name, although its not too hard to infer in this context.

I think a name like `HandleMessageAsync(..)` or even `SubmitMessageAsync(..)` would help to make it more obvious. But this is probably just personal preference.

Blish HUD/GameServices/ArcDps/V2/Processors/MessageProcessor{T}.cs
```diff
@@ -19,14 +19,20 @@ public override void Process(byte[] message, CancellationToken ct) {
...
 
-        public void RegisterListener(Func<T, CancellationToken, Task> listener) {
+        public void RegisterListener(IArcDpsMessageListener<T> listener) {
             listeners.Add(listener);
+            listener.Disposing += RemoveListener;
+        }
+
+        private void RemoveListener(IArcDpsMessageListener<T> listener) {
+            listener.Disposing -= RemoveListener;
+            listeners.Remove(listener);
```
Flyga-M, line 35: I'm wondering if `listeners` needs to be a threadsafe collection.

For most cases it's probably unlikely that a listener is removed (or registered for that matter) at the same time that a message is processed, but it seems to me, that they technically still (can) occur on different threads.

I'm assuming this never caused issues in the past (registering while a message is processed), because it's just so unlikely. Unless I'm misunderstanding the situation.

Blish HUD/GameServices/ArcDpsServiceV2.cs
```diff
@@ -77,20 +70,27 @@ private set {
...
+        public void RegisterMessageType<T>(IArcDpsMessageListener<T> listener)
             where T : struct {
-            RegisterMessageType<T>((int)type, listener);
-        }
-
-        public void RegisterMessageType<T>(int type, Func<T, CancellationToken, Task> listener)
-            where T : struct {
-            Action action = () => _arcDpsClient.RegisterMessageTypeListener(type, listener);
+            Action action = () => _arcDpsClient.RegisterMessageTypeListener(listener);
             _registerListeners.Add(action);
```
Flyga-M, line 76: I think keeping the `Action` for the whole lifetime of the `ArcDpsServiceV2` will cause issues, because the `Action` can be invoked again if the Mumble process id changes (details below). It might even get worse, if the module that registered the listener is unloaded at some point.

This could probably be handled by a similar approach to the `MessageProcessor<T>`:
Subscribing to the `IArcDpsMessageListener<T>.Disposing` event and removing the `Action` from `_registerListeners`.

Since these actions are anonymous, no easy solution comes to mind. Either `_registerListeners` would need to be changed to `List<IArcDpsMessageListener>` while also introducing a non-generic `IArcDpsMessageListener` interface, or even using something like `Dictionary<IArcDpsMessageListener, Action>` if storing potentially different kinds of actions should be kept as a (currently unused) feature.

The same remarks I made about Thread-Safety in `MessageProcessor<T>` apply here if this change was made I would assume.

---
### The problematic situation in detail
1. A listener is added
   - so the `Action` to register the listener is kept in `_registerListeners`
2. The listener is disposed
   - the listener is unregistered and removed from the corresponding `MessageProcessor`
   - the callback that the listener carries is assumed to be unregistered
4. The Mumble process id changes
   - `ArcDpsServiceV2.Start(..)` is called
   - the - in the meantime disposed - listener is re-registered with the new `ArcDpsClient`
5. A new message (of the relevant `MessageType`) is sent
   - the corresponding `MessageProcessor` will forward the message to the listener

Flyga-M (commented):
Thank you all (@Denrage @KirillSerogodsky @greaka) for the work on both features! Judging from the discussions on Discord a lot of effort and coordination went into this PR (and everything connected to it).

It's great that you make the new arcdps_unofficial_extras features available for Blish HUD modules and also ensure that the listeners can be removed again.

I haven't worked a lot with the `ArcDpsService` in the past, so I'm happy to be corrected if some of my assumptions are incorrect.

I left most of my questions or suggestions directly with the relevant lines of code. The only topic that didn't really fit was the discussion about whether some changes should be considered breaking or not. So I'll leave my thoughts here.

## Is this a breaking change?
There are some changes in this PR that would typically be considered breaking:
- renaming `struct` `ChatMessageInfo` to `SquadMessageInfo`
- namespace changes
  - `ChannelType.cs`
  - `Language.cs`
  - `SquadMessageInfo.cs` (prev. `ChatMessageInfo.cs`)
  - `UserInfo.cs`
  - `UserRole.cs`
- method name change (and return type change) of public extension method `BincodeBinaryReaderExtensions.ParseSquadMessageInfo`

If I understand you correctly, you consider them not-really-breaking, because the currently released `arcdps-bhud` does not support the `arcdps_unofficial_extras` in any capacity. So it's highly unlikely that anybody would have been using the changes introduced in the last Blish HUD release in the first place, right?

Is there any way to make sure this is the case? Do we know which modules make use of the ArcDPS integration?

I'm sure this only (potentially) affects a very small amount of all modules, but it might still be worth to have a look at.

If any dev (now or in the future) would target Blish HUD 1.2.0 specifically and make use of the arcdps_unofficial_extras integration, their module would still break in any other version (e.g. 1.3.0), right?
I guess this is a very specific edge case - relevant to keep in mind none the less.

If we keep these technically breaking changes and only increment the minor version for the next release, we should document this behavior somewhere very visible.

The other alternative would be to leave the names and namespaces as is, though I do think it would add confusion to diverge from the naming used by arcdps_unofficial_extras.

I'd appreciate your and others thoughts on that!

---

Just for completeness, here are some changes that I first thought might be breaking, but on closer inspection are not:
- signature change of public method `MessageProcessor<T>.RegisterListener`
   - not breaking, because `MessageProcessor<T>` is an `internal` `class`
- renaming of `UnofficialExtrasMessageInfoProcessor` to `UnofficialExtrasSquadMessageInfoProcessor`
   - not breaking, because it is an `internal` `class`
- naming change in public enum `MessageType`
   - not breaking, because the old `ChatMessage` is identical to the new `SquadMessage`
- signature change in `interface` method `IArcDpsClient.RegisterMessageTypeListener<T>(..)` <- same thing applies to the implementation `ArcDpsClient`
   - not breaking, because the `interface` is `internal`

Flyga-M:
Ah I also forgot to ask: Has anybody made a test module so it's easier to test the new ArcDPS integration?

I'm not currently working on any module that makes use of the feature, so I would have to make one just for this purpose. Figured I'd ask, before doing redundant work.

Flyga-M:
Just for documentation purposes, because I think this hasn't been linked yet: This PR would close issue #625

Blish HUD/GameServices/ArcDps/V2/IArcDpsMessageListener.cs
Denrage: `HandleMessageAsync` feels redundant, because it is a `MessageListener` and the method takes `T message` as the first argument. `Submit` feels weird because that is normally an action a user takes explicitly.
So i would leave that as it is

Blish HUD/GameServices/ArcDpsServiceV2.cs
Denrage: Good catch. The main reason for this approach was that you can register listeners before the service was available. Didn't think about the mumble process id changing. Question is here does the mumble process id change without blish being restarted bc gw2 was closed?
Afaik the mumble process id change will be called once for gw2 being started and never again.
I think that is also the main reason why i didn't do any cleanup here

Denrage:
> Ah I also forgot to ask: Has anybody made a test module so it's easier to test the new ArcDPS integration?
> 
> I'm not currently working on any module that makes use of the feature, so I would have to make one just for this purpose. Figured I'd ask, before doing redundant work.

Nope, i'm currently testing it by abusing the aboutview adding code as i need it to test it 👀

Blish HUD/GameServices/ArcDps/V2/Processors/MessageProcessor{T}.cs
Denrage: Listeners should probably be threadsafe, not because of the issue you describe (having a one message delay before the listener gets removed is fine imo) but because we go through the list while we could possibly remove smth from the list, which changes the collection while we foreach through it

Blish HUD/GameServices/ArcDps/V2/CommonFields.cs
Denrage: I would not change anything in the V1 stuff because we would need to think about breaking changes there otherwise. The V1 should behave exactly the same as before and you were not able to remove anything before too, so we shouldn't start doing that.

The obsolet versions of the RegisterMessage is the same as with the V1. The behaviour before was "you can't remove stuff" and if you start removing it for them, the behaviour changes. If it is relevant, idk. But maybe someone insane actually abuses this behaviour, we just can't know.

Using the removal feature in `CommonFields` seems reasonable. Even if it is not necessary, we should clean up. I didn't do it before, bc the V1 version also didn't do any cleanup and i didn't thought about it, the same with the `ArcDpsServiceV2.Start` method.

dlamkins (approved):

Blish HUD/GameServices/ArcDpsServiceV2.cs
greaka: IIRC Blish should restart after every GW2 close, yes.

greaka (approved):

PR #1004, opened 2025-03-23
Add fallback to process attach

eksime:
This allows Blish to fall back through processes if the primary options aren't available - this should prevent us showing contingency messages for when there's two gw2 processes running and one is admin, or running as another user, or similar.
We still show the contingency message in the case that the last process we considered was unable to be opened, and there are no more options. Fallback doesn't occur when using the explicit named mumble link.

It's possible we may attached to a non-admin gw2 instance, while the mumble data is being written by an admin instance & get mismatched mumble data, but I'm not sure it's possible to identify that situation cleanly.

## Discussion Reference
https://canary.discord.com/channels/531175899588984842/534490472224391169/1352591368387563655

## Is this a breaking change?
No


eksime:
This needs further work as it seems to prevent attach in normal circumstances occasionally.

PR #1040, opened 2026-01-04
Add number input controls

sliekens:
This PR adds `NumberInput` and `FloatInput` controls, matching the in-game number input controls.

(I contemplated making a single `NumberInput<T>` but this very quickly became too complicated.)

Additionally, I added number inputs alongside the existing trackbar for number setting views.

### Supported features

- Programatically read or write values
- Customizable min/max values
- Customizable number formatting
- Click or press "spinner" buttons to increase/decrease the value
- Customizable increment intervals for spinner buttons
- Use mouse wheel scroll to increase/decrease the value
- Click the number to start typing a value
    - Press Return/Esc to stop typing
- Looks cool

### NumberInput control in this PR

<table>
  <tr>
    <th>normal</th>
    <th>active</th>
    <th>disabled</th>
  </tr>
  <tr>
    <td>image</td>
    <td>image</td>
    <td>image</td>
  </tr>
</table>

### Inspired by in-game control

<table>
  <tr>
    <th>normal</th>
    <th>active</th>
  </tr>
  <tr>
    <td>image</td>
    <td>image</td>
  </tr>
</table>

### Number setting views example

image

## Discussion Reference

Not so much a discussion but having a number input is a common ask.

https://discord.com/channels/531175899588984842/599270434642460753/1208669259677569034
https://discord.com/channels/531175899588984842/534492012263636992/1329031835434487819

## Is this a breaking change?

No


dlamkins:
If the textures are unmodified from what is in the dat, can this be updated to pull them from the `DatAssetCache` instead?  We've already migrated a number of the older textures to rely on it instead of adding new textures to the ref.

Blish HUD/Controls/FloatInput.cs
```diff
@@ -0,0 +1,472 @@
...
+
+        private float _maxValue = float.MaxValue;
+
+        private float _incrementAmount = 0.1f;
+
+        public FloatInput() {
+            Width = 150;
+            Height = SpinnerButtonHeight * 2;
+            TextChanged += OnTextChanged;
+            InputFocusChanged += OnInputFocusChanged;
```
dlamkins: Any reason to add listeners to itself instead of overriding the handler like you've done with OnMouseMoved and OnMouseMoved (the latter is more common with other Blish HUD controls)?

dlamkins:
I've tested in some modules.  

1. I'm seeing that the text within the control seems to be misaligned to the top right.  Can that be fixed?
2. The number of decimals shown seems to vary from property.  I think the control will need to have a property specifying the number of decimal places with a sensible default of 1 or 2.

CleanShot 2026-01-04 at 17 47 22@2x

Blish HUD/Controls/NumberInput.cs
```diff
@@ -0,0 +1,451 @@
...
+
+        private int _minValue = int.MinValue;
+
+        private int _maxValue = int.MaxValue;
+
+        public NumberInput() {
+            Width = 150;
+            Height = SpinnerButtonHeight * 2;
+            TextChanged += OnTextChanged;
+            InputFocusChanged += OnInputFocusChanged;
```
dlamkins: Same as the other control.  Any reason to add listeners to itself instead of overriding the handler like you've done with OnMouseMoved and OnMouseMoved (the latter is more common with other Blish HUD controls)?

sliekens: No deliberate reason, I will change it

sliekens:
> If the textures are unmodified from what is in the dat, can this be updated to pull them from the DatAssetCache instead?

Not a problem, just gotta find those asset IDs again 🫡

> I'm seeing that the text within the control seems to be misaligned to the top right.  Can that be fixed?

Ah yea I see what you mean, the textbox sprite is too tall, I'll see what can be done

> The number of decimals shown seems to vary from property. I think the control will need to have a property specifying the number of decimal places with a sensible default of 1 or 2.

I am not sure which way to proceed with this:

1. Round the number when it is assigned to the control (loss of precision)
2. Display a rounded number but keep the value with full precision

Blish HUD/Controls/NumberInput.cs
sliekens: As it turns out, originally I programmed it with event handlers in my module because those methods aren't virtual, but now I have the ability to change that.

Blish HUD/Controls/FloatInput.cs
sliekens: Updated PR

sliekens:
Progress report:

- The self-subscriptions have been replaced with method overrides (the base methods had to be made `virtual` to facilitate this)
- The `ref` textures have been replaced with usages of `AsyncTexture2D.FromAssetId`
- You can now assign a [format string](https://learn.microsoft.com/en-us/dotnet/standard/base-types/standard-numeric-format-strings) to `NumberInput.FormatString` and `FloatInput.FormatString`
   - Default: `G` ([general number](https://learn.microsoft.com/en-us/dotnet/standard/base-types/standard-numeric-format-strings#general-format-specifier-g))
   - `FloatSettingView` sets this to `F2` ([Fixed-point format](https://learn.microsoft.com/en-us/dotnet/standard/base-types/standard-numeric-format-strings#fixed-point-format-specifier-f))
- Fixed sizing and alignment problems that I could find

Alternatives considered:

I decided against adding a `Precision` for float inputs because it doesn't solve inconsistent formatting (among other reasons).

dlamkins:
Looks great!  Thank you for those changes and improvements!  I suspect that the numbers being visible might encourage some module developers to modify their ranges to make them easier for end-users to work with (as I will probably do with Pathing to avoid such large units that get cut off).

CleanShot 2026-01-06 at 15 36 32@2x

Right now, floats above 100.00 are cropped.  I suppose the control width could be increased a bit to help give some extra space, but that does cut more into what is available of the bar itself.

I did notice one last strange thing.  The alignment seems to break some after the number exceeds the size of the box.  Once you select into the box again, the alignment is corrected.

https://github.com/user-attachments/assets/7710d967-b485-4c5b-99f5-c977a59eadd6

sliekens:
@dlamkins thanks for the visuals, I think I found a reasonable solution for both issues

1. Always right-align numbers
2. Grow textbox based on `MeasureString`
   - I cheated a bit by allowing text overflow (`ClipsBounds = false`) to avoid complex layout updates
   
(The downside of `ClipsBounds  = false` is that the overflowing text is not clickable.)   
   

https://github.com/user-attachments/assets/966e1172-c395-4242-a719-3739c5449746

sliekens:
I just noticed that the text is not selected on focus as intended... 🤔

/edit: fixed, not sure how that was working inside my module before, that should never have worked.

sliekens:
I'm fixing one more thing, the value should not update as you type, only when the control loses focus.

sliekens:
Okay all done, I am happy with the current state of this PR.

https://github.com/user-attachments/assets/573c8417-a2a1-4026-8bb7-7ba1271f452c

dlamkins:
Okay, this is getting very close.  My testing today found 3 oddities:

1. The spinners seem to allow non-number values to be entered into them.  I saw that if a number was included, only the number was kept (which seems reasonable), but if only text was provided, the text remains.
2. Selecting the spinner value highlights it (which I like!), but you are then unable to select anything specific within the number (for example, if I want to only change one digit it is a bit of a fight).  Can it be changed so only the first click highlights all of the text and subsequent clicks allow you to move the cursor as expected?  Once the control becomes unfocused again, clicking into it should again highlight the full number. 
3. The ClipBounds trick does seem to do a good job of making the full number visible, but it took me a bit to figure out why I couldn't click into the box until I realized that only the portions of the box on the right-side were actually part of the control and the rest was outside of its bounds. 

https://github.com/user-attachments/assets/db5eb387-a7b8-42ac-88cf-71f35d3f8322

dlamkins:
There also appears to be a breaking change with the changes to the SpriteBatchExtension.  I encountered this crash when trying another module:

> 03:01:35.7332 -05:00 | FATAL | Blish_HUD.DebugService | Blish HUD encountered a fatal crash!
System.MissingMethodException: Method not found: 'Void Blish_HUD.SpriteBatchExtensions.DrawStringOnCtrl(Microsoft.Xna.Framework.Graphics.SpriteBatch, Blish_HUD.Controls.Control, System.String, MonoGame.Extended.BitmapFonts.BitmapFont, Microsoft.Xna.Framework.Rectangle, Microsoft.Xna.Framework.Color, Boolean, Boolean, Int32, Blish_HUD.Controls.HorizontalAlignment, Blish_HUD.Controls.VerticalAlignment)'.
   at MysticCrafting.Module.Menu.FilterMenuItem.PaintBeforeChildren(SpriteBatch spriteBatch, Rectangle bounds)
   at Blish_HUD.Controls.Container.Paint(SpriteBatch spriteBatch, Rectangle bounds) in D:\a\Blish-HUD\Blish-HUD\Blish HUD\Controls\Container.cs:line 291
   at Blish_HUD.Controls.Control.Draw(SpriteBatch spriteBatch, Rectangle drawBounds, Rectangle scissor) in D:\a\Blish-HUD\Blish-HUD\Blish HUD\Controls\Control.cs:line 920

dlamkins:
Clears Tracker shows the bar going off the side.  🤔  Not sure where the fault lies for that, though.

CleanShot 2026-01-07 at 19 52 45@2x

vs. 

CleanShot 2026-01-07 at 19 54 21@2x

sliekens:
> The spinners seem to allow non-number values to be entered into them.  I saw that if a number was included, only the number was kept (which seems reasonable), but if only text was provided, the text remains.

I have some real trauma from trying to prevent non-numeric input (it would still cause other side-effects like advancing the cursor) so I decided to not even attempt this again. Instead, my solution captures the current value when the control gains focus, and the original value is restored if the new text can't be parsed as a number.

I did remove the ability to enter numbers mixed with text, except for what is allowed by `TryParse`. That means you can now enter numbers like `1e3` (10³) and it is parsed as `1000` instead of `13`, I'll let you decide whether that's a bug or a feature.

> Selecting the spinner value highlights it (which I like!), but you are then unable to select anything specific within the number

Solved by selecting all text when the control gains focus instead of on every click.

> The ClipBounds trick does seem to do a good job of making the full number visible, but it took me a bit to figure out why I couldn't click into the box until I realized that only the portions of the box on the right-side were actually part of the control and the rest was outside of its bounds.

I don't have a good solution for this. I could attempt an unreasonable workaround like recalculating the layout of the entire SettingView, but it would be an inconvenience to review. I recommend making a change elsewhere such that overflowing content can receive mouse events. 

> There also appears to be a breaking change with the changes to the SpriteBatchExtension.

Huge blunder, good thing you caught it early 😅.

> Clears Tracker shows the bar going off the side

I adjusted the size of the trackbar to keep the total width of the SettingView the same as before.

dlamkins:
> I did remove the ability to enter numbers mixed with text, except for what is allowed by TryParse. That means you can now enter numbers like 1e3 (10³) and it is parsed as 1000 instead of 13, I'll let you decide whether that's a bug or a feature.

Yep!  That's perfect.  👌 

> Solved by selecting all text when the control gains focus instead of on every click.

This also seems to work well now.  Thank you!

> I don't have a good solution for this. I could attempt an unreasonable workaround like recalculating the layout of the entire SettingView, but it would be an inconvenience to review. I recommend making a change elsewhere such that overflowing content can receive mouse events.

No, I believe changing input handling in such a way would be far more unreasonable.  I think not clipping bounds in this instance is just not the right way to go about this.

> I adjusted the size of the trackbar to keep the total width of the SettingView the same as before.

Everywhere I'm checking it looks good so far.  Thanks!

sliekens:
> No, I believe changing input handling in such a way would be far more unreasonable. I think not clipping bounds in this instance is just not the right way to go about this.

Okay, the initial idea was to grow the textbox area to fit the value, but I changed it to a fixed width to avoid this category of layout problems.

As a result big numbers are now clipped, you have to focus the control and scroll with arrow keys to see the rest. I am unsure what to do here, what do you think?

https://github.com/user-attachments/assets/9532e09c-a224-4189-850a-e6c4ea0fd36f

---

By the way I extracted duplicate code between number/float inputs into a common `NumberInputBase` class.

Blish HUD/GameServices/Settings/UI/Views/FloatSettingView.cs
```diff
@@ -4,42 +4,82 @@
...
 
         protected override void BuildSetting(Container buildPanel) {
             base.BuildSetting(buildPanel);
 
             _valueTrackBar.SmallStep = true;
+
+            _floatInput = new FloatInput() {
+                Width           = INPUT_WIDTH,
+                Left            = TRACKBAR_LEFT,
+                ClipsBounds     = false, // bit of a hack to allow overflowing text for large numbers
```
dlamkins: We can probably remove this line.

Blish HUD/GameServices/Settings/UI/Views/IntSettingView.cs
```diff
@@ -1,18 +1,45 @@
...
+
         public IntSettingView(SettingEntry<int> setting, int definedWidth = -1) : base(setting, definedWidth) { /* NOOP */ }
+
+        protected override void BuildSetting(Container buildPanel) {
+            base.BuildSetting(buildPanel);
+
+            _numberInput = new NumberInput() {
+                Width       = INPUT_WIDTH,
+                Left        = TRACKBAR_LEFT,
+                ClipsBounds = false, // bit of a hack to allow overflowing text for large numbers
```
dlamkins: Same for this one - we can remove to prevent confusion from the comment.

Blish HUD/GameServices/Settings/_Compliance/FloatRangeRangeComplianceRequisite.cs
```diff
@@ -1,12 +1,28 @@
...
+
+namespace Blish_HUD.Settings {
     public readonly struct FloatRangeRangeComplianceRequisite : INumericRangeComplianceRequisite<float> {
         
         public float MinValue { get; }
         public float MaxValue { get; }
+        public float IncrementAmount { get; }
 
-        public FloatRangeRangeComplianceRequisite(float minValue, float maxValue) {
+        public FloatRangeRangeComplianceRequisite(float minValue, float maxValue, float incrementAmount) {
```
dlamkins, line 10: Probably should be its own ctor.  This is going to be a breaking change for anyone setting compliance for floats already.

You can have everything in one ctor and then have the other overloads call :this() with the defaults that aren't provided.

dlamkins: Ahh never mind.  I think this is mostly used internally and modules are calling the extension methods like SetRange, so this is _probably_ fine.

Blish HUD/GameServices/Settings/UI/Views/FloatSettingView.cs
dlamkins: ClipBounds lines can all be removed.

Blish HUD/Controls/NumberInputBase.cs
```diff
@@ -0,0 +1,501 @@
...
+                    UnsetFocus();
+                    break;
+                case NumberInputAction.None:
+                default:
+                    base.OnLeftMouseButtonReleased(e);
+                    break;
+            }
+
+            _action = NumberInputAction.None;
+        }
```
dlamkins, line 379: Does the control stop reacting to input if `.Enabled = false;`?  OnClick automatically checks before OnClick is called (and the event is fired), but the other click handlers don't, so you might need to check if the control is Enabled before performing certain actions.

Blish HUD/GameServices/Settings/_Compliance/FloatRangeRangeComplianceRequisite.cs
sliekens: I double checked to be sure, I think this change is safe, but now I am wondering why `FloatRangeRangeComplianceRequisite` is public rather than internal, and whether I should add `IncrementAmount` to the interface `INumericRangeComplianceRequisite<T>`

dlamkins: I *suspect* it was an oversight. 

I'll be honest, when I saw the compliance code, it made me want to do a partial rewrite of how that all works and what is available, so I wouldn't recommend getting too into the weeds on those details for this PR.

Blish HUD/Controls/NumberInputBase.cs
sliekens: Good point, I forgot to consider `Enabled`, updated to ignore mouse events and grey out the control when `Enabled = false`

Image

Blish HUD/GameServices/Settings/UI/Views/FloatSettingView.cs
sliekens: Removed

Blish HUD/GameServices/Settings/UI/Views/IntSettingView.cs
sliekens: Removed

Blish HUD/GameServices/Settings/UI/Views/FloatSettingView.cs
sliekens: Removed

Blish HUD/GameServices/Settings/_Compliance/FloatRangeRangeComplianceRequisite.cs
sliekens: I guess this can be marked resolved for now

entrhopi:
Hi!
This is looking good!
Not to be nitpicky or something, but maybe you wouldn't mind also implementing mouse up/down dragging the number while in the field? This would then align with all the number inputs in the game, which would be really nice!

sliekens:
@entrhopi I had mouse dragging in my original implementation, you can still find it in the Chat Links module, but it doesn't work great. In particular, there is no way to freeze and hide the cursor while dragging, so it bounces around the screen. I removed dragging specifically for this PR.

entrhopi:
> @entrhopi I had mouse dragging in my original implementation, you can still find it in the Chat Links module, but it doesn't work great. In particular, there is no way to freeze and hide the cursor while dragging, so it bounces around the screen. I removed dragging specifically for this PR.

I see. This makes sense and would indeed be more confusing than helpful.
Thanks for the explanation.

entrhopi:
Just a little late feedback on the matter:
I found that in the GW2 options menu there is a similar input/slider-combo for the fullscreen gamma. This does not support mouse dragging either. So all good here.
image

Thanks for making this!

sliekens:
Okay, so I'm just awaiting a response on how to proceed with large numbers. Maybe I could make the trackbar even smaller but I don't know what we want to do here. Alternatives exist like showing ellipsis with tooltips on hover but there is no precedent for something like that.

sliekens:
Okay, so I'm just awaiting a response on how to proceed with large numbers. Maybe I could make the trackbar even smaller but I don't know what we want to do here. Alternatives exist like showing ellipsis with tooltips on hover but there is no precedent for something like that.

dlamkins:
Could we see some screenshots of the trackbar squeezed a little smaller (maybe something conservative and then another squeezed how much it would have to be to see the full number in Pathing)?

I think for modules like Pathing I'll just want to adjust the range that users are working with, but maybe there is a small compromise that could help slightly larger numbers.

sliekens:
@dlamkins what do you think of this?

- Trackbar even smaller (sadge but still works)
- Textbox shrinks or grows to fit content (up to `Control.Width`)

https://github.com/user-attachments/assets/959cc5f7-34da-4cca-8452-050c085af73b

Same thing but 10px wider trackbar:

image

image

dlamkins:
Okay, I think the 10px wider one is almost certainly it.  If you can commit those changes, I'll give it another lap through some modules.  😄

sliekens:
@dlamkins  10px resize pushed

dlamkins (approved):

dlamkins:
Love it - thanks for putting this together!

greaka (approved):

agaertner:
I know this is already merged but what I hate about the in-game original one is that it does **not** wrap around to the maximum if you decrement below the minimum. It starts off at the minimum and forces you to go up all the way to the max or calc or type out what you can purchase. It's a hassle because there are a lot of vendors where you just want to purchase the maximum and be on your way.

Maybe add a public override (or settable Func<int>) that is invoked which calculates and sets a dynamic maximum when user would trigger a wrap-around?

```csharp
public override CalculateNextValue(int nextValue) {
    return nextValue > _max ? _min : (nextValue < _min ? _max : nextValue);
}
```

or as ``Func<int>`` member (if it's assigned wrap-around is enabled.)

```
Func<int> _nextValueFunc;

[...]

value = _nextValueFunc?.Invoke();
```

PR #1041, opened 2026-01-08
Wrap Debug logger in async target to avoid blocking calling thread

sliekens:
NLog's `MethodCallTarget` is pretty slow (4-5x slower than directly calling `Debugger.Log` on my machine) and ever so slightly slows down all modules while a debugger is attached, or when the user has enabled debug logging in the overlay settings.

Wrapping the debug target in an async target speeds up `Logger` methods dramatically and avoids blocking the calling thread, which is pretty important because calls to `Debugger.Log` are synchronized by a `lock` statement during debugging. (The lock only affects debugging sessions.)

This fix primarily helps speed up debugging code with verbose logging, but it also benefits users who play with debug logging enabled for whatever reason.

## Discussion Reference
https://discord.com/channels/531175899588984842/536970543736291346/1458955030781624447

## Is this a breaking change?
No

### DIFF
--- a/Blish HUD/GameServices/DebugService.cs
+++ b/Blish HUD/GameServices/DebugService.cs
@@ -134,8 +134,13 @@ namespace Blish_HUD {
                 }
             };
 
-            _logConfiguration.AddTarget(logDebug);
-            _logConfiguration.AddRule(LogLevel.Debug, LogLevel.Fatal, logDebug);
+            // MethodCallTarget is synchronous and also quite slow, wrap it in an async target to avoid blocking the main thread
+            var asyncDebug = new AsyncTargetWrapper("asyncdebug", logDebug) {
+                ForceLockingQueue = false
+            };
+
+            _logConfiguration.AddTarget(asyncDebug);
+            _logConfiguration.AddRule(LogLevel.Debug, LogLevel.Fatal, asyncDebug);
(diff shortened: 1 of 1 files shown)


dlamkins (approved):

greaka (approved):

snakefoot:
Notice the NLog [MethodCallTarget](https://github.com/NLog/NLog/wiki/MethodCall-target) also supports calling a method delegate, instead of resolving the callback-method through reflection.

The reflection path is much slower than providing a direct delegate and also gives extra allocations. Since you initialize the NLog MethodCallTarget from code, then it should be simple.

https://nlog-project.org/documentation/v4.6.0/html/M_NLog_Targets_MethodCallTarget__ctor_2.htm

Example code:
```c#
var target = new MethodCallTarget("Hello", (logEvent, parameters) => {
    string time = logEvent.TimeStamp.ToLocalTime().ToString("HH:mm:ss.ffff K");
    // Consider optimizing this with a switch-statement to skip string-allocations
    string level = logEvent.Level.ToString().ToUpper().PadLeft(5, ' ');
    string logger = logEvent.Logger;
    string message = logEvent.FormattedMessage;
    DebugService.TargetDebug(time, level, logger, message);
});
```

PR #1044, opened 2026-01-16
Use delegate-based MethodCallTarget to avoid reflection overhead

sliekens:
## Discussion Reference
This improvement was suggested by @snakefoot (largest NLog contributor by commits)  in #1041

## Is this a breaking change?

No

### DIFF
--- a/Blish HUD/GameServices/DebugService.cs
+++ b/Blish HUD/GameServices/DebugService.cs
@@ -100,7 +100,7 @@ namespace Blish_HUD {
         public static void TargetDebug(LogEventInfo logEvent, object[] parameters) {
             if (!Debugger.IsAttached) return;
 
-            string time = logEvent.TimeStamp.ToLocalTime().ToString("HH:mm:ss.ffff K");
+            DateTime localTime = logEvent.TimeStamp.ToLocalTime();
 
             // Use switch to return string constants for known log levels, avoiding string allocations
             string level = logEvent.Level.Ordinal switch {
@@ -118,7 +118,7 @@ namespace Blish_HUD {
             const int INTERNAL_DEBUG_WRITESIZE = 4091;
 
             lock (_debugLock) {
-                string outEntry = $"{time} | {level} | {logger} | {message}\r\n";
+                string outEntry = $"{localTime:HH:mm:ss.ffff K} | {level} | {logger} | {message}\r\n";
 
                 // Messages that are too large can cause issues for various debuggers
                 if (outEntry.Length >= INTERNAL_DEBUG_WRITESIZE) {
--- a/Blish HUD/GameServices/DebugService.cs
+++ b/Blish HUD/GameServices/DebugService.cs
@@ -95,8 +95,6 @@ namespace Blish_HUD {
             Logger = Logger.GetLogger<DebugService>();
         }
 
-        private static readonly object _debugLock = new object();
-
         public static void TargetDebug(LogEventInfo logEvent, object[] parameters) {
             if (!Debugger.IsAttached) return;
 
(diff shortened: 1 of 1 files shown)


Blish HUD/GameServices/DebugService.cs
```diff
@@ -123,18 +123,15 @@ public static void TargetDebug(string time, string level, string logger, string
...
-                    new MethodCallParameter(STRUCLOG_LEVEL),
-                    new MethodCallParameter(STRUCLOG_LOGGER),
-                    new MethodCallParameter(STRUCLOG_MESSAGE)
-                }
-            };
-
-            // MethodCallTarget is synchronous and also quite slow, wrap it in an async target to avoid blocking the main thread
+            var logDebug = new MethodCallTarget("logdebug", (logEvent, parameters) => {
+                string time = logEvent.TimeStamp.ToLocalTime().ToString("HH:mm:ss.ffff K");
+                string level = logEvent.Level.Name.ToUpperInvariant().PadRight(5);
```
snakefoot: Maybe introduce helper-logic for converting `logEvent.Level`:
```csharp
            string level = "FATAL";
            if (logEvent.Level == NLog.LogLevel.Trace)
                level = "TRACE";
            else if (logEvent.Level  == NLog.LogLevel.Debug)
                level = "DEBUG";
            else if (logEvent.Level  == NLog.LogLevel.Info)
                level = "INFO ";
            else if (logEvent.Level  == NLog.LogLevel.Warn)
                level = "WARN ";
            else if (logEvent.Level == NLog.LogLevel.Error)
                level = "ERROR";
```
Skipping string-allocations.

sliekens: Good idea, implemented as switch based on Level.Ordinal

Blish HUD/GameServices/DebugService.cs
```diff
@@ -97,9 +97,24 @@ internal static void InitDebug() {
...
+                4 => "ERROR",
+                5 => "FATAL",
+                _ => logEvent.Level.Name.ToUpperInvariant().PadRight(5)
+            };
+            string logger = logEvent.LoggerName;
+            string message = logEvent.FormattedMessage;
+
             const int INTERNAL_DEBUG_WRITESIZE = 4091;
 
             lock (_debugLock) {
```
sliekens: @snakefoot do you happen to know if NLog logger methods synchronize their calls to `MethodCallTarget`, or if multiple threads/loggers can enter the delegate at the same time? This `lock` was meant to solve an issue where logs bleed into each other, but I doubt that is still relevant.

greaka (approved):

Blish HUD/GameServices/DebugService.cs
snakefoot: `MethodCallTarget` has an implicit lock that prevents multiple threads to be inside the callback-delegate at the same time.

I now see that you are calling `Debugger.Log`. Notice NLog also has the [NLog Debugger target](https://github.com/NLog/NLog/wiki/Debugger-target) but doesn't have the chunking-up-very-large-messages logic.

But I guess the `lock` is there to ensure the chunking become a "single unit". But not necessary when called from NLog Target, but usually a `lock` without congestion is very cheap (interlocked-operation).

Blish HUD/GameServices/DebugService.cs
```diff
@@ -97,9 +97,24 @@ internal static void InitDebug() {
 
         private static readonly object _debugLock = new object();
 
-        public static void TargetDebug(string time, string level, string logger, string message) {
+        public static void TargetDebug(LogEventInfo logEvent, object[] parameters) {
             if (!Debugger.IsAttached) return;
 
+            string time = logEvent.TimeStamp.ToLocalTime().ToString("HH:mm:ss.ffff K");
```
snakefoot: Maybe instead of doing the time-format upfront, then maybe do it like this (skips string-allocation):
```csharp
DateTime localtime = logEvent.TimeStamp.ToLocalTime();
string outEntry = $"{localtime:HH:mm:ss.ffff K} | {level} | {logger} | {message}\r\n";
```

sliekens: Another good suggestion, thanks :)

Blish HUD/GameServices/DebugService.cs
sliekens: Okay, thanks for confirming my suspicion, I have removed the lock since NLog already has the implicit lock

snakefoot (approved):

dlamkins:
This isn't the first time you've helped us, @snakefoot!  We really appreciate it!

dlamkins (approved):
Thanks @sliekens for seeing the comment on the other PR and running with the improvements!

PR #1045, opened 2026-01-29
Load assets from disk first and then try background update

entrhopi:
## Discussion Reference
https://discord.com/channels/531175899588984842/536970543736291346/1463924298103263334

## Is this a breaking change?
No

Following a discussion about recent problems with Cloudflare and Deutsche Telekom, metadata will be loaded from disk first and updated in the background instead of waiting for the download to finish, every time Blish HUD loads.

### DIFF
--- a/Blish HUD/GameServices/Content/DatAssetCache.cs
+++ b/Blish HUD/GameServices/Content/DatAssetCache.cs
@@ -68,19 +68,18 @@ namespace Blish_HUD.Content {
         private async Task<Stream> DownloadMetadata() {
             string metadataCache = Path.Combine(_assetCachePath, METADATA_FILE);
 
-            var metadataStream = Stream.Null;
-
             try {
                 byte[] rawMetadata = await $"{ASSETSERV_HOST}/{METADATA_FILE}".GetBytesAsync();
 
                 File.WriteAllBytes(metadataCache, rawMetadata);
-                metadataStream = new MemoryStream(rawMetadata);
-                Logger.Info("Metadata update successful");
-            } catch (Exception ex) {
-                Logger.Warn(ex, "Failed to load asset metadata.");
-            }
+                Logger.Debug("Metadata update successful");
 
-            return metadataStream;
+                return new MemoryStream(rawMetadata);
+            } catch (Exception ex) {
+                Logger.Info(ex, "Failed to load asset metadata.");
+
+                return Stream.Null;
+            }
         }
 
         private void EarlyLoad() {
@@ -91,14 +90,14 @@ namespace Blish_HUD.Content {
                 // Open local file and run background update
                 using Stream localMetadataStream = File.Open(metadataCache, FileMode.Open, FileAccess.Read, FileShare.Read);
                 ProcessMetadataStream(localMetadataStream);
-                Logger.Info("Local metadata loaded, trying background update");
+                Logger.Debug("Local metadata loaded, trying background update");
 
                 _ = Task.Run(async () =>
                 {
                     try {
                         using var _ = await DownloadMetadata().ConfigureAwait(false);
                     } catch (Exception ex) {
-                        Logger.Warn(ex, "Background metadata refresh failed");
+                        Logger.Info(ex, "Background metadata refresh failed");
                     }
                 });
             } else {
@@ -111,7 +110,7 @@ namespace Blish_HUD.Content {
 
         private void ProcessMetadataStream(Stream metadataStream) {
             if (metadataStream.Length == 0) {
-                Logger.Warn("Failed to load asset metadata.  Textures won't be loaded.");
+                Logger.Info("Failed to load asset metadata. Textures won't be loaded.");
 
                 _textureReferences = new Dictionary<int, TextureReference>(0);
                 _textureSizes = Array.Empty<Point>();
--- a/Blish HUD/GameServices/Content/DatAssetCache.cs
+++ b/Blish HUD/GameServices/Content/DatAssetCache.cs
@@ -85,27 +85,25 @@ namespace Blish_HUD.Content {
         private void EarlyLoad() {
             string metadataCache = Path.Combine(_assetCachePath, METADATA_FILE);
 
-            // Check if metadata cache exists locally
+            // Use either existing metadata cache or fallback to ref.dat
             if (File.Exists(metadataCache)) {
-                // Open local file and run background update
                 using Stream localMetadataStream = File.Open(metadataCache, FileMode.Open, FileAccess.Read, FileShare.Read);
                 ProcessMetadataStream(localMetadataStream);
                 Logger.Debug("Local metadata loaded, trying background update");
-
-                _ = Task.Run(async () =>
-                {
-                    try {
-                        using var _ = await DownloadMetadata().ConfigureAwait(false);
-                    } catch (Exception ex) {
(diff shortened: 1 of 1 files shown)


Blish HUD/GameServices/Content/DatAssetCache.cs
```diff
@@ -65,59 +65,79 @@ private static Stream LoadFallbackMetadataStream() {
...
         }
 
-        private Stream LoadMetadataStream() {
+        private async Task<Stream> LoadMetadataStream() {
             string metadataCache = Path.Combine(_assetCachePath, METADATA_FILE);
 
             var metadataStream = Stream.Null;
 
             try {
                 // We block for this on purpose
```
sliekens: Comment no longer reflects behavior

Blish HUD/GameServices/Content/DatAssetCache.cs
```diff
@@ -65,59 +65,78 @@ private static Stream LoadFallbackMetadataStream() {
...
-                Logger.Warn(ex, "Failed to load asset metadata.  Will attempt to load from local cache instead");
-                metadataStream = File.Exists(metadataCache)
-                                     // Use the last one successfully downloaded
-                                     ? File.Open(metadataCache, FileMode.Open, FileAccess.Read, FileShare.Read)
-                                     // Yikes, use the one we ship with
-                                     : LoadFallbackMetadataStream(); 
+                Logger.Warn(ex, "Failed to load asset metadata.");
             }
 
             return metadataStream;
```
greaka: I think this function body could get cleaned up. The variable metadaStream could be elided and you could return the new stream directly to make the code more explicit

greaka (commented):
I think the logging is a bit too verbose. Imo, we don't need info logs for optional operations and also don't want warnings when caches are not warm.

Turning all info logs to debug and all warning logs to info would be appropriate I think.

dlamkins:
I agree with Greaka about the logging levels.  

1) Is it possible that the metadata.gz file on disk is still being read (and, as a result, is locked) when the downloaded version attempts to write over it?  Perhaps something can be introduced that ensures the writes don't start until the metadata file on disk is no longer being read.  This would never be an issue for a DTAG customer, though - Ha!

2) I think this loses the fallback (`LoadFallbackMetadataStream`) where we store a copy of the metadata file in the ref.  It's an old version of the metadata.gz (we should probably update it before the next release), but it's the true fallback if the user is never able to download a fresh copy from the web.

entrhopi:
>     1. Is it possible that the metadata.gz file on disk is still being read (and, as a result, is locked) when the downloaded version attempts to write over it?
Since the entire operation is async (and file size is also 'minimal') I don't think this could happen. It would wait for the read operation as well.

>     2. I think this loses the fallback (`LoadFallbackMetadataStream`) where we store a copy of the metadata file in the ref.  It's an old version of the metadata.gz (we should probably update it before the next release), but it's the true fallback if the user is never able to download a fresh copy from the web.
Good catch!

Blish HUD/GameServices/Content/DatAssetCache.cs
```diff
@@ -65,59 +65,75 @@ private static Stream LoadFallbackMetadataStream() {
...
             _transparentTextures = new Texture2D[sizeCount];
 
             for (int sizeIndex = 0; sizeIndex < sizeCount; sizeIndex++) {
-                int width  = parser.ReadInt32();
+                int width = parser.ReadInt32();
                 int height = parser.ReadInt32();
 
-                _textureSizes[sizeIndex]        = new Point(width, height);
+                _textureSizes[sizeIndex] = new Point(width, height);
                 _transparentTextures[sizeIndex] = new Texture2D(BlishHud.Instance.GraphicsDevice /* This is safe since we're loading early on the main thread */, width, height);
```
dlamkins, line 137: We might need to update this, actually.  The first read (which happens on main thread) would still be safe, but subsequent references wouldn't necessarily be safe.  You can probably get the graphics context outside of the for loop (so you're only requesting it once and then all of the new textures get created as part of one batch).

https://blishhud.com/docs/modules/guides/graphicsdevice#lendgraphicsdevicecontext

Blish HUD/GameServices/Content/DatAssetCache.cs
```diff
@@ -65,59 +65,75 @@ private static Stream LoadFallbackMetadataStream() {
...
+                } catch (Exception ex) {
+                    Logger.Info(ex, "Background metadata refresh failed");
+                }
+            });
+        }
 
+        private void ProcessMetadataStream(Stream metadataStream) {
             if (metadataStream.Length == 0) {
-                Logger.Warn("Failed to load asset metadata.  Textures won't be loaded.");
+                Logger.Info("Failed to load asset metadata. Textures won't be loaded.");
```
dlamkins: Okay, I disagree with Greaka on this, actually.  I think this should remain a warn.  There is likely to be user impact if this occurs, so it's above an info.

Blish HUD/GameServices/Content/DatAssetCache.cs
```diff
@@ -65,59 +65,75 @@ private static Stream LoadFallbackMetadataStream() {
...
+                ProcessMetadataStream(metadataStream);
+                Logger.Debug("Falling back to ref.dat, trying background update");
+            }
+
+            // Try background update of metadata cache
+            _ = Task.Run(async () => {
+                try {
+                    using var _ = await DownloadMetadata().ConfigureAwait(false);
+                } catch (Exception ex) {
+                    Logger.Info(ex, "Background metadata refresh failed");
```
dlamkins: I think this should be a warn as there is an exception involved (and there is the possibility of user impact if they needed that file to load new textures).

Blish HUD/GameServices/Content/DatAssetCache.cs
```diff
@@ -65,59 +65,75 @@ private static Stream LoadFallbackMetadataStream() {
...
+                return new MemoryStream(rawMetadata);
             } catch (Exception ex) {
-                Logger.Warn(ex, "Failed to load asset metadata.  Will attempt to load from local cache instead");
-                metadataStream = File.Exists(metadataCache)
-                                     // Use the last one successfully downloaded
-                                     ? File.Open(metadataCache, FileMode.Open, FileAccess.Read, FileShare.Read)
-                                     // Yikes, use the one we ship with
-                                     : LoadFallbackMetadataStream(); 
-            }
+                Logger.Info(ex, "Failed to load asset metadata.");
```
dlamkins: I think this should be a warn as there is an exception involved (and there is the possibility of user impact if they needed that file to load new textures).

dlamkins (changes requested):

Blish HUD/GameServices/Content/DatAssetCache.cs
dlamkins: Okay, now that I think about it, these textures likely don't need rebuilt entirely.  There is maybe a rare situation where a new size range needs to be added, but we can keep the existing range of them and just check if new ones need to be added.

entrhopi: @dlamkins and I discussed this, and we'll make a new PR for this feature, which will be merged before the next release.

dlamkins (approved):

greaka (approved):

PR #1053, opened 2026-02-10
feat: add basic StandardDialog control

agaertner:
A basic auto-sizing screen prompt similar to system prompts.

**A convinient way (and thus an incentive) for devs to alert the user when they are about todo a destructive or non-reversable action.**

It's basically what GW2 does when you disconnect or are about to throw away an item.

Obviously due to the auto-scaling the background stretches if the text is too long.


dlamkins:
Can you include some screenshots of examples from in-game for comparison?

agaertner:
image

image

Call examples
```csharp
ScreenPrompt.Show("Some text"); // No icon; default button: "OK"

ScreenPrompt.Show("Some text", DialogIcon.Exclamation);

ScreenPrompt.Show("Some text", DialogIcon.Exclamation, DialogButton.Close);

ScreenPrompt.Show("Some text", DialogIcon.Question, DialogButton.Yes | DialogButton.No, (bttn) => { 
   if (bttn = DialogButton.Yes) // yes was pressed to close the window.
});

AsyncTexture2D customIcon;
ScreenPrompt.Show("Some text", customIcon);

ScreenPrompt.Show("Some text", customIcon, DialogButton.Close);

ScreenPrompt.Show("Some text", customIcon, DialogButton.Confirm | DialogButton.Cancel, (bttn) => { 
   if (bttn = DialogButton.Confirm) // confirm was pressed to close the window.
});
```

greaka:
Here are a few ingame screenshots for reference.

image
image

agaertner:
oh I just wanted to redo the system one for a quick way to alert a user, much like ScreenNotification but with a callback.
You could be right, however, you could dive deeper into this and abstract it more.

dlamkins:
Can a see a screenshot of the in-game popup this implementation is replicating?  

I like the convenience of this implementation, but it immediately makes me want a way to provide custom button text or icons. 🤔 (but maybe that's not needed if the available button options cover most scenarios).

agaertner:
image (Original)

The disconnect one is colorless but the item destruction one is the same (except the displayed buttons of course). I thought of maybe adding an overload that takes LabelPart for colorization.

agaertner:
image
Here it is without icon. 

I refactored it to use FormattedLabelBuilder (yes, the builder. At first I wanted to use FormattedLabel but I want the prompt to enforce size and wrap. It's a win thought since the dev doesn't have to make those calls and Build it.)

Give me a second to clean it up.

Here's the result. 
image


Call example:
```csharp
var label = new FormattedLabelBuilder();
label.CreatePart(label.CreatePart("Are you sure you want to destroy "));
label.CreatePart("250 Refined Homestead Fibers", o => o.SetTextColor(new Color(255, 164, 5)));
label.CreatePart(label.CreatePart("?"));
ScreenPrompt.Show(label, ScreenPrompt.DialogIcon.Exclamation, ScreenPrompt.DialogButton.Yes | ScreenPrompt.DialogButton.No);
```

I noticed some strange overload of CreatePart in the formattedLabelBuilder. You can see that I am using CreatePart and then pass it to CreatePart again. It's because CreatePart(string) does not add it to the building process of FormattedBuilder.
I will weave in a fix for that. It doesn't break anything.

agaertner:
Alright. All done. :)

Call examples:

```csharp
ScreenPrompt.Show("Some text"); // No icon; default button: "OK"

ScreenPrompt.Show("Some text", DialogIcon.Exclamation);

ScreenPrompt.Show("Some text", DialogIcon.Exclamation, DialogButton.Close);

ScreenPrompt.Show("Some text", DialogIcon.Question, DialogButton.Yes | DialogButton.No, (bttn) => { 
   if (bttn = DialogButton.Yes) // yes was pressed to close the window.
});

AsyncTexture2D customIcon;
ScreenPrompt.Show("Some text", customIcon);

ScreenPrompt.Show("Some text", customIcon, DialogButton.Close);

ScreenPrompt.Show("Some text", customIcon, DialogButton.Confirm | DialogButton.Cancel, (bttn) => { 
   if (bttn = DialogButton.Confirm) // confirm was pressed to close the window.
});

// Instead of a plain string a FormattedLabelBuilder can be prepared and passed with the rest of the params of above.

var label = new FormattedLabelBuilder();
label.CreatePart("Are you sure you want to destroy ");
label.CreatePart("250 Refined Homestead Fibers", o => o.SetTextColor(new Color(255, 164, 5)));
label.CreatePart("?");

ScreenPrompt.Show(label, ScreenPrompt.DialogIcon.Exclamation, ScreenPrompt.DialogButton.Yes | ScreenPrompt.DialogButton.No);

ScreenPrompt.Show(label, ScreenPrompt.DialogIcon.Exclamation, ScreenPrompt.DialogButton.Yes | ScreenPrompt.DialogButton.No, (bttn) => { 
   if (bttn = DialogButton.Yes) // yes was pressed to close the window.
});
```
image

Stress test (yes, the button's width do auto adjust; the container's width does not, thereby keeping it "dialog-like"):

image

agaertner:
Overlapping the original:

image

Close enough. 😅

agaertner:
You may already see why I think this basic control may come in handy. It's just a convinient way (and thus an incentive) for devs to alert the user when they are about todo a destructive or non-reversable action.

Blish HUD/Controls/ScreenPrompt.cs
```diff
@@ -0,0 +1,336 @@
...
+            var noText = label == null;
+            if (noText) {
+                errorMessage += "Prompt dialog must have text content.";
+            }
+            return string.IsNullOrEmpty(errorMessage);
+        }
+
+        private void LoadTextures() {
+            _bgTexture ??= GameService.Content.GetTexture(@"controls/prompt/156003");
+            _icons ??= GameService.Content.GetTexture(@"controls/prompt/154985");
```
dlamkins: These should be pulled from the Asset CDN and not bundled in the ref.

Blish HUD/_Extensions/SpriteBatchExtensions.cs
```diff
@@ -237,5 +238,41 @@ public static void DrawStringOnCtrl(this SpriteBatch    spriteBatch,
...
+            DrawRectangleOnCtrl(spriteBatch, ctrl, bounds, lineWidth, Color.Black);
+        }
+
+        public static void DrawRectangleOnCtrl(this SpriteBatch spriteBatch, Control ctrl, Rectangle bounds, Color color) {
+            DrawRectangleOnCtrl(spriteBatch, ctrl, bounds, Math.Min(bounds.Width, bounds.Height) / 2, color);
+        }
+
+        public static void DrawRectangleOnCtrl(this SpriteBatch spriteBatch, Control ctrl, Rectangle bounds) {
+            DrawRectangleOnCtrl(spriteBatch, ctrl, bounds, Color.Black);
+        }
```
dlamkins: I'm not a _huge_ fan of adding 4 new extension functions for this one control to call.

Blish HUD/Controls/ScreenPrompt.cs
```diff
@@ -0,0 +1,336 @@
...
+            Yes = 1 << 3,
+            No = 1 << 4,
+            Ignore = 1 << 5,
+            Close = 1 << 6,
+            Apply = 1 << 7,
+            Decline = 1 << 8
+        }
+
+        public enum DialogIcon
+        {
```
dlamkins: Remove the newline

Blish HUD/Controls/ScreenPrompt.cs
```diff
@@ -0,0 +1,336 @@
...
+using System.Linq;
+using Color = Microsoft.Xna.Framework.Color;
+using Point = Microsoft.Xna.Framework.Point;
+using Rectangle = Microsoft.Xna.Framework.Rectangle;
+namespace Blish_HUD.Controls {
+    public sealed class ScreenPrompt : Container
+    {
+        [Flags]
+        public enum DialogButton : ushort
+        {
```
dlamkins: Remove the newline

Blish HUD/Controls/ScreenPrompt.cs
```diff
@@ -0,0 +1,336 @@
...
+using MonoGame.Extended.BitmapFonts;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Color = Microsoft.Xna.Framework.Color;
+using Point = Microsoft.Xna.Framework.Point;
+using Rectangle = Microsoft.Xna.Framework.Rectangle;
+namespace Blish_HUD.Controls {
+    public sealed class ScreenPrompt : Container
+    {
```
dlamkins: Remove the newline

Blish HUD/Controls/ScreenPrompt.cs
```diff
@@ -0,0 +1,336 @@
...
+            foreach (DialogButton button in Enum.GetValues(typeof(DialogButton))) {
+                if (button == DialogButton.None) continue;
+                if ((buttons & button) == button) _buttons.Add(button, null); // Each key will be assigned a button instance in CreateButtons().
+            }
+
+            _enterButton = enterButton;
+            _escapeButton = escapeButton;
+            _callback = callback;
+
+            this.ZIndex = 999;
```
dlamkins: I don't know the right answer to this, but things might get weird depending on how others have set the ZIndex on their controls.  999 isn't _that_ high (compared to the full range).

Blish HUD/Controls/ScreenPrompt.cs
```diff
@@ -0,0 +1,336 @@
...
+            None = 0,
+            OK = 1 << 0,
+            Confirm = 1 << 1,
+            Cancel = 1 << 2,
+            Yes = 1 << 3,
+            No = 1 << 4,
+            Ignore = 1 << 5,
+            Close = 1 << 6,
+            Apply = 1 << 7,
+            Decline = 1 << 8
```
dlamkins: I think this might enum-based implementation is too opinionated.  It would be great to be able to provide our own buttons.  For example, some that come to mind are:
- "Retry"
- "Continue"
- "I Accept"
- "Download"
- "Delete"
- "Don't Remind Me"

Instead of needing to list each, it'd be nice if there was a happy-medium where the user could specify their own (though, it is nice that you have translations for the enum options already).

dlamkins:
Your screenshots look pretty great!  Could you please send a screenshot of your implementation next to a matching in-game version?  I can see there are some differences, but it's hard to tell in the overlap.  😄

Blish HUD/Controls/ScreenPrompt.cs
```diff
@@ -0,0 +1,336 @@
...
+using Microsoft.Xna.Framework.Graphics;
+using Microsoft.Xna.Framework.Input;
+using MonoGame.Extended.BitmapFonts;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Color = Microsoft.Xna.Framework.Color;
+using Point = Microsoft.Xna.Framework.Point;
+using Rectangle = Microsoft.Xna.Framework.Rectangle;
+namespace Blish_HUD.Controls {
```
dlamkins: I maybe would have put this under `Blish_HUD.Common.Gw2.UI`.  I would have done the same with the `ScreenNotification` as well now with 20-20.  

It's not really used as a control in the same sense that the other types in this namespace are.

Blish HUD/_Extensions/SpriteBatchExtensions.cs
agaertner: Yeah I was thinking as these are in use quite often by me and others as well I might just add them as extensions. Drawing simple pixel borders is quite in demand. This draws a filled rect or border depending on if line width is passed.

Blish HUD/Controls/ScreenPrompt.cs
agaertner: Oops. Thought you wanted it this way in core. Did that before. Will revert it!

Blish HUD/Controls/ScreenPrompt.cs
agaertner: Me neither. I had it set way higher. I forgot why I set it lower. Will increase it.

Blish HUD/Controls/ScreenPrompt.cs
agaertner: The predefinition take inspiration from basic alert boxes of other frameworks. Most of the time you can work with these.
For a text like 'Don't remind me' it's more or less also a matter of what info was given before or in the prompt.
The action could state 'Do you want to be reminded?' then the prepared texts are sensible in meaning. 

I understand your suggestions thought.
They are more personal and use-case oriented and less techy in tone. The predifinitions are as generic as can be to be applicable anywhere. I like the idea of more personal buttons thought like in 'Don't Remind Me'. Makes you feel "Blishhud is my friend not a neutral robot." 😂

I'd have to change the callback to work with button index instead (😭). I'd keep the predefinitions and add the customButtonTexts.
~~I will remove the key shortcuts for the buttons and add customButtonTexts as param at the very end so it's easy to make one call without having to create a List.~~

Blish HUD/Controls/ScreenPrompt.cs
agaertner: Understood. I just tried to match ScreenNotification. You sure you don't want it match that?

agaertner: Changed.

Blish HUD/Controls/ScreenPrompt.cs
agaertner: I added an additional 9. You got a better idea? I think initially I looked at the default z indexes you defined and took something sensible. Not sure why I changed it.

Blish HUD/_Extensions/SpriteBatchExtensions.cs
agaertner: Resolved your reviews but can I leave those spriteBatch extensions please?

agaertner:
> Your screenshots look pretty great! Could you please send a screenshot of your implementation next to a matching in-game version? I can see there are some differences, but it's hard to tell in the overlap. 😄

image (Original)

image (Blish HUD)

The border appears 1px thicker when I have it on my greenscreen window. I previously had it 1px lower because it looked thinner in-game. Might be due to the fuzzyness of BlishHUD/Monogame. I would keep as is.
The button text is a bit bigger (I believe) in the original but the StandardButton has the scope of its font private.

 However, the remaining differences are so miniscule that it doesn't matter imo.

agaertner:
Custom button text:

image

Blish HUD/Controls/ScreenPrompt.cs
```diff
@@ -303,7 +306,7 @@ private static string GetDefaultButton(DialogButton button) {
             };
         }
 
-        private void CreateButtons() {
+        private void CalcButtonLayout() {
```
greaka: please use `Calculate` in full

agaertner:
Adjusted button size and reverted the extra border pixel.

image

Should be good now.

Blish HUD/Controls/ScreenPrompt.cs
agaertner: Zindex is now just below dropdown and tooltip.

Blish HUD/Controls/ScreenPrompt.cs
dlamkins: Thanks!

Blish HUD/Controls/ScreenPrompt.cs
dlamkins: Okay, that seems reasonable.  🙂

Blish HUD/Controls/ScreenPrompt.cs
dlamkins: Sorry to give you whiplash, but I thought about it more while laying in bed and I think the cat is out of the bag on this one and under Controls is probably the best spot.  😆  Since you've made the change, though, I'll leave it up to you which you'd prefer it end up as.

dlamkins:
Are these pictures of the popup being taken while UI size is set to large?  There shouldn't be any fuzziness while UI Size is set to large and that would help us determine if the lines are correct.  🤔  

I also see that the background texture is the same, but this implementation resizes it to fit whereas the real one does not squish it.  Can that be made to change?

Blish HUD/_Extensions/SpriteBatchExtensions.cs
dlamkins: I suppose it can stay.  I'm just not a fan of expanding the public API surface of Blish HUD that much in these particular areas.

I might at least ask that you do the "OnCtrl" math once and call spriteBatch.Draw directly in the call instead of calling DrawOnCtrl each time (which does the offset math 4 times).

agaertner:
> Are these pictures of the popup being taken while UI size is set to large? There shouldn't be any fuzziness while UI Size is set to large and that would help us determine if the lines are correct. 🤔
> 
> I also see that the background texture is the same, but this implementation resizes it to fit whereas the real one does not squish it. Can that be made to change?

UI is set to small actually. I was developing on small UI size.

I can try to change the bg texture

agaertner:
UI size set to Large now:
image (Original)

image
(BlishHUD)

Looks good

agaertner:
> Are these pictures of the popup being taken while UI size is set to large? There shouldn't be any fuzziness while UI Size is set to large and that would help us determine if the lines are correct. 🤔
> 
> I also see that the background texture is the same, but this implementation resizes it to fit whereas the real one does not squish it. Can that be made to change?

Done and done.
Example
image

Stress test
image

Text content area debug (still correct after refactor)
image

Blish HUD/Controls/_Types/FormattedLabelBuilder.cs
```diff
@@ -12,8 +12,11 @@ public class FormattedLabelBuilder {
         private HorizontalAlignment _horizontalAlignment = HorizontalAlignment.Left;
         private VerticalAlignment _verticalAlignment = VerticalAlignment.Middle;
 
-        public FormattedLabelPartBuilder CreatePart(string text)
-            => new FormattedLabelPartBuilder(text);
+        public FormattedLabelPartBuilder CreatePart(string text) {
+            var part = new FormattedLabelPartBuilder(text);
+            _parts.Add(part.Build()); // For cases where we want this shorthand but keep using this FormattedLabelBuilder.
+            return part;
+        }
```
Denrage: I really don't like this change in the API.
A builder should not be used anymore after the `Build`-Method was called.
I should have added a check since the beginning, if the `Build`-Method was already called and throw an exception on any subsequent actions on the builder.
The Builder creates an object that the builder has no reference anymore to (and shouldn't have).
Yes it seems weird to use (probably shouldn't the Method be called `CreatePart(FormattedLabelPartBuilder` but rather `AddPart(FormattedLabelPartBuilder)` but after the `Build` methdo on the PartBuilder is called, that part is done, and there shouldn't be a reason to use the builder afterwards.

The easy shorthand would be to just use `CreatePart("", null)`, that's what i used at the places where i just wanted to add a string

Blish HUD/Controls/ScreenPrompt.cs
```diff
@@ -0,0 +1,381 @@
...
+        /// <param name="text">Text inside the popup.</param>
+        /// <param name="buttons">Buttons that the prompt should have.</param>
+        /// <param name="callback">Function that is called when a button is pressed.</param>
+        /// <param name="enterButtonIndex">Index of a button that can be triggered via the Enter key on the keyboard.</param>
+        /// <param name="escapeButtonIndex">Index of a button that can be triggered via the Escape key on the keyboard.</param>
+        public static void Show(string text, DialogButton buttons = DialogButton.OK, Action<int> callback = null,
+                                      int enterButtonIndex  = 0,
+                                      int escapeButtonIndex = 1) {
+            Show(text, DialogIcon.None, buttons, callback, enterButtonIndex, escapeButtonIndex);
+        }
```
Denrage: Idk if i would use the button index to determine which is the `escapeButton` and which is the `enterButton`. I would rather put the button object that is in the Buttons-Array also in there or mark the button accordingly (with a bool or just an interface) and use the first one you find.
Indezes are so error prone. You cannot make sure at that point in the code that the order in the array is still correct. It feels like code smell and yearns for errors in the future.

Denrage (commented):
I don't know if I'm alone with these comments and i wasn't explicitly asked to look at this PR, so i won't Request Changes, but those are things that give me a bad feeling seeing in core. @greaka @dlamkins

Blish HUD/Controls/_Types/FormattedLabelBuilder.cs
Denrage: Actually after thinking about it, this change would make any code using the `CreatePart(string)` as it was supposed to, break.

Someone created a PartBuilder with this method, and added it later to the LabelBuilder, and now the text is duplicated there, but once without any changes he made, bc it was build once before.

agaertner: Then I guess I need a shorthand for that "", null call.

Man. I really should have taken a look at your PR of FormattefLabelBuilder. I feel like we missed so many things with it. I remember it even got merged with an AutoSize crash. Not to trash it or anything. The thing is awesome but the implementation seems to be less extensible and flexible than I'd like.

Blish HUD/Controls/ScreenPrompt.cs
agaertner: I know what you mean and I used DialogButton before but free wanted custom text so it had to be something.

The index is fine. I made sure to reflect it what you initialize it with. I worked with frameworks that do similarly. I hate it as well but it never breaked. No one is gonna do deep array work on it.

The index is always reflecting custom text array initialization / buttons left to right.

agaertner: I know what you mean and I used DialogButton before but free wanted custom text so it had to be something.

The index is fine. I made sure to reflect it what you initialize it with. I worked with frameworks that do similarly. I hate it as well but it never breaked. No one is gonna do deep array work on it.

The index is always reflecting custom text array initialization / buttons left to right.

Blish HUD/Controls/_Types/FormattedLabelBuilder.cs
Denrage: I would suggest adding the following method to the builder:
```
public FormattedLabelBuilder CreateAndAddPart(string text, Action<FormattedLabelPartBuilder> creationFunc = null)
```
and deprecate the method
```
[Obsolet("Use CreateAndAddPart", false)]
public FormattedLabelBuilder CreatePart(string text, Action<FormattedLabelPartBuilder> creationFunc = null)
```
Now you can just use `CreateAndAddPart("HelloWorld")` or `CreateAndAddPart("HelloWorld", o => o.MakeBold())`

agaertner: Fixed with #d09bf02

Blish HUD/_Extensions/SpriteBatchExtensions.cs
agaertner: Done.

I removed the other two overloads (filled rectangles are simple enough with DrawOnCtrl)
and renamed it to DrawBorderOnCtrl.

Blish HUD/Controls/ScreenPrompt.cs
agaertner: Resolved with new DialogButton approach.

agaertner:
- Fix some math for button fitting if there are over 5 buttons.
- Clean up margin consts and adjusted them more by comparison with original.
- Fix a race condition on GetDefaultIcon and TextureSwapped (icon could stay empty if spammed fast enough because the Texture may already be in the AsyncTexture2D from DatAssetCache before TextureSwapped is subscribed to. 
- Clean up & Testing

### Examples 

```csharp
var label = new FormattedLabelBuilder();
label.CreatePart("There's plenty of exotic", o => o.SetFontSize(FontSize.Size16));
label.CreatePart("fish", o => o.SetFontSize(FontSize.Size16).SetTextColor(new Color(255, 164, 5)));
label.CreatePart("in the sea.", o => o.SetFontSize(FontSize.Size16));
label.CreatePart("\n", o => o.SetFontSize(FontSize.Size16));
label.CreatePart("\nDo you agree?", o => o.SetFontSize(FontSize.Size16));
StandardDialog.Show(label,
    DialogIcon.Exclamation,
    DialogButton.Yes.Action(() => { }),
    DialogButton.No.Select());
```

image


```csharp
StandardDialog.Show("The quick brown fox jumps over the lazy dog.", DialogIcon.Question, 
    DialogButton.Create("Pet the Dog").Action(() => { }), 
    DialogButton.Ignore.Select());
```
image



image

-----------------------------------
In the following the icon margin have additionally been adjusted further to match the original.

```csharp
StandardDialog.Show("Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet. Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet.", DialogIcon.Question, 
                DialogButton.Create("Hello World").Action(() => { }), 
                DialogButton.Ignore.Select(),
                DialogButton.OK,
                DialogButton.Close,
                DialogButton.Cancel,
                DialogButton.Decline,
                DialogButton.Apply,
                DialogButton.Confirm,
                DialogButton.No);

```
image

```csharp
StandardDialog.Show("Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet. Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet.", 
                DialogIcon.Present,
                DialogButton.OK,
                DialogButton.Close);
 ```
image

agaertner:
image

PR #1054, opened 2026-02-11
Split module settings into their own files.

dlamkins:
## Problem

Module settings are stored along with all other settings within the `setting.json` file.  To avoid deserializing module-specific types when a module isn't loaded, the settings are kept as raw JSON strings (via `SettingCollection`'s lazy loading).  However, this means the full JSON blob for every module (including those that are disabled) is held in memory at all times.  For users with many modules, this results in a large `settings.json` and unnecessary memory usage.

## Solution

Module settings are migrated out of `settings.json` into individual files under a new `settings/` directory, named by module namespace (e.g. `settings/bh.community.pathing.json`).  Settings are only loaded into memory when a module is enabled, and flushed back to disk and freed when it's disabled.

## Settings migration

- A migration takes place if module settings are detected in the `settings.json` file.  When the migration occurs, a backup of the original settings file is made (if one doesn't already exist), just in case.

## Results

image

image

### DIFF
--- a/Blish HUD/GameServices/SettingsService.cs
+++ b/Blish HUD/GameServices/SettingsService.cs
@@ -1,11 +1,13 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using Blish_HUD.Content.Serialization;
 using Blish_HUD.Controls;
 using Blish_HUD.Settings;
 using Microsoft.Xna.Framework;
 using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
 
 namespace Blish_HUD {
 
@@ -16,7 +18,9 @@ namespace Blish_HUD {
 
         private const int SAVE_INTERVAL = 4;
 
-        private const string SETTINGS_FILENAME = "settings.json";
+        private const string SETTINGS_FILENAME  = "settings.json";
+        private const string SETTINGS_PMBACKUP  = "settings-premigration.json";
+        private const string SETTINGS_DIRECTORY = "settings";
 
         [Obsolete]
         public delegate void SettingTypeRendererDelegate(SettingEntry setting, Panel settingPanel);
@@ -27,10 +31,15 @@ namespace Blish_HUD {
         [JsonIgnore]
         private string _settingsPath;
 
+        [JsonIgnore]
+        private string _settingsDirectory;
+
         internal SettingCollection Settings { get; private set; }
 
         private bool   _dirtySave;
         private double _saveBuffer;
+
+        private readonly Dictionary<string, SettingCollection> _moduleSettings = new Dictionary<string, SettingCollection>();
 
         protected override void Initialize() {
             JsonReaderSettings = new JsonSerializerSettings() {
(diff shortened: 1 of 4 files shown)


Blish HUD/GameServices/SettingsService.cs
```diff
@@ -126,6 +222,64 @@ private void PerformSave() {
...
 
+        private static void WriteFileAtomic(string filePath, string content) {
+            string tempPath = $"{filePath}.new";
+
+            using (var writer = new StreamWriter(tempPath, false)) {
+                writer.Write(content);
+            }
+
+            if (File.Exists(filePath)) {
+                File.Replace(tempPath, filePath, null);
```
greaka, line 243: `File.Replace` can fail if the file is opened somewhere. A more conservative approach would be to move the existing file, if that fails choose another name and try again, move the new file to the intended location, then try to delete the existing file, ignore failures. Failed deletes can be retried on subsequent calls to this function.

If you choose to implement this then please consider writing a separate method for `MoveOrReplaceFile`.

Blish HUD/GameServices/SettingsService.cs
```diff
@@ -100,18 +199,15 @@ public void Save(bool forceSave = false) {
...
-                }
+                // Save settings.json
+                WriteFileAtomic(_settingsPath, rawSettings);
 
-                if (File.Exists(_settingsPath)) {
-                    File.Replace(tempSettingsPath, _settingsPath, null);
-                } else {
-                    File.Move(tempSettingsPath, _settingsPath);
+                // Save all registered module settings to their individual files
+                foreach (var mkp in _moduleSettings) {
```
entrhopi: Could this maybe cause a race condition if `_moduleSettings` is modified during write?

Blish HUD/GameServices/SettingsService.cs
dlamkins: Hmm this is the [same pattern we use currently](https://github.com/blish-hud/Blish-HUD/blob/266bc471c18abeaf8dd2e8dc2759fb30386f4e34/Blish%20HUD/GameServices/SettingsService.cs#L105-L121) and haven't had any reports of issues.

In the current implementation, failing to write leaves the last save there and the save services continues to consider the save as dirty (meaning it will try to save again in about 4 seconds.  With this implementation, the most you'll lose is the delta between the actually saved version and the .new version (which _can_ be manually restored).

Your approach has a few extra moving parts that I worry could result in losing a settings file because there isn't a settings file in the right spot at all times.

Blish HUD/GameServices/SettingsService.cs
dlamkins: Yeah, that could probably be a problem if a module is disabled off of the main thread.  I'll add something to ensure it always happens on the main thread.

dlamkins: I've added some locking.  👍

Blish HUD/GameServices/SettingsService.cs
entrhopi: Sounds reasonable to me, since it's a one time migration.

greaka: This code is used during normal operations, on every save of any settings file.
I have no strong opinion on this either way.

entrhopi (approved):
I've looked into the file saving mechanic again, and I think it is sufficiently safe. In case of it not working we still have the .new file for recovery.

greaka (approved):

PR #1055, opened 2026-02-12
Process asset metadata from server once downloaded

dlamkins:
Follow-up to #1045.  When the background asset metadata download completes, `ProcessMetadataStream` now merges new entries into the existing data.  This preserved already-loaded `TextureReference` objects and their `WeakReference<AsyncTexture2D>` targets, avoiding unnecessary texture re-fetching.

## Changes

- `DatAssetCache.ProcessMetadataStream` - On subsequent calls (after initial load), the method now:
  - Reuses existing texture sizes and their transparent placeholder textures, only creating new ones for previously unseen dimensions.
  - Copies existing `TextureReference` objects into a new dictionary (preserving live `WeakReference` targets), adds only new asset IDs, then swaps the references.
  - Skips resetting state if called with an empty stream when the data is already loaded.
- `GraphicsDeviceContext.Dispose` - I've added a null guard on `_service` so that calling `Dispose()` on a `default` instance is safe (`GraphicsDeviceContext is a readonly ref struct, so it can't be nullable or implement IDisposable, meaning I'd otherwise need to keep track if I leased it with a bool).

### DIFF
--- a/Blish HUD/GameServices/Content/DatAssetCache.cs
+++ b/Blish HUD/GameServices/Content/DatAssetCache.cs
@@ -3,6 +3,7 @@ using Blish_HUD.Graphics;
 using Flurl.Http;
 using Microsoft.Xna.Framework;
 using Microsoft.Xna.Framework.Graphics;
+using SharpDX.MediaFoundation;
 using System;
 using System.Collections.Generic;
 using System.IO;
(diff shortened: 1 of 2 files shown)


Blish HUD/GameServices/Content/DatAssetCache.cs
```diff
@@ -85,63 +86,156 @@ private async Task<Stream> DownloadMetadata() {
...
+                GraphicsDevice        graphicsDevice;
 
-            _textureSizes = new Point[sizeCount];
-            _transparentTextures = new Texture2D[sizeCount];
+                if (GameService.Graphics != null) {
+                    graphicsLease  = GameService.Graphics.LendGraphicsDeviceContext(true);
+                    graphicsDevice = graphicsLease.GraphicsDevice;
+                } else {
+                    graphicsDevice = BlishHud.Instance.GraphicsDevice;
+                }
```
greaka: What will happen if the graphics device is directly accessed and bhud then initializes the gameservice on another thread?

entrhopi:
I would suggest creating 3 helper methods for `ProcessMetadataStream` to make it more manageable and improve clarity, since it's getting pretty extensive:
`DecompressStream`, `InitializeMetadata` and `MergeMetadata`

Other than that everything else looks fine to me! 👍

Blish HUD/GameServices/Content/DatAssetCache.cs
dlamkins: That's a good point.  _Probably_ nothing, but there is once again that 1% of risk on the other side of it where something does try to access it and it causes a crash.

I suppose the only 100% safe options are either:
1) Skip loading the remote asset metadata entirely if `GameService.Graphics` is null (and we'll be sure to load it in the next session).
2) Have processing wait until `GameService.Graphics` is defined and then lend the graphics context.  I think this is the better option, but a bit unfortunate.  The longer we wait, the larger the chance that we miss early texture requests (but it should be a very short wait and it's unlikely that we have modules or core asking for anything that soon that isn't already likely to be cached).

I'll implement the second unless you have another other ideas/thoughts.  🙂

dlamkins:
Is this change what you're looking for?  I'm not really a fan of it - mostly because of how much the two new functions require from the processing done in the first (and splitting it out into a third for DecompressStream only to be returned and then offsets and the counts have to be done feels weird).

If we didn't have to do this in either case:

```csharp
int offset = 0;

int totalTextureCount = BitConverter.ToInt32(data, offset);
offset += 4;

int sizeCount = BitConverter.ToInt32(data, offset);
offset += 4;

bool isUpdate = _textureReferences != null && _textureReferences.Count > 0;
```

It would be much easier to split out, but this is shared and used by both paths.

Blish HUD/GameServices/Content/DatAssetCache.cs
greaka: That's fine, thanks! :)

dlamkins: We now wait to ensure GameService.Graphics is available :)

entrhopi (approved):

entrhopi:
Looking at this now, I agree with you about the third one. It was what came into mind at the time.
👍

greaka (approved):

Issue #392, opened 2021-07-20
Scrollbar speed is linear


dlamkins:
Currently the scrollbar speed is linear which means that how "correct" it feels completely changes depending on the height of the scrolling area.

The scroll speed should be proportional to the height of the window in some way so that smaller scrolling areas scroll in smaller amounts compared to larger panels which should scroll in larger amounts.  Testing in other applications seems to show that this is the typical behavior.

entrhopi:
Just noticed that mouse wheel input behaves strange too.

I can scroll one 'tik'. But as soon as I scroll multiple 'tiks' it doesn't seem to trigger every time, And switching the mouse to continuous scrolling will only scroll one 'tik'.

PR #1056, opened 2026-02-14
Fix: Mouse wheel scrolling

entrhopi:
## Description
When using the mouse wheel to scroll through a control, `GameService.Input.Mouse.State.ScrollWheelValue` seems to overflow the value from the up direction to -65536 when scrolling fast. This causes the direction to reverse.

Since the value is also proportional to the speed of the mouse wheel, this also seems to be a good candidate for #392 
It feels very natural scrolling through lists now.

## Is this a breaking change?
No

### DIFF
--- a/Blish HUD/Controls/Scrollbar.cs
+++ b/Blish HUD/Controls/Scrollbar.cs
@@ -16,7 +16,6 @@ namespace Blish_HUD.Controls {
         private const int SCROLL_ARROW      = 50;
         private const int SCROLL_CONT_ARROW = 10;
         private const int SCROLL_CONT_TRACK = 15;
-        private const int SCROLL_WHEEL      = 30;
 
         #region Load Static
 
@@ -162,13 +161,16 @@ namespace Blish_HUD.Controls {
 
             if (scrollValue == 0) return;
 
-            // Handle the scroll value overflow (happens when scrolling a lot in one direction)
+            // Handle the scroll value overflow (happens when scrolling fast in one direction)
             if (scrollValue < -32000) {
                 scrollValue += 65536;
             }
 
-            float scrollDirection = Math.Sign(scrollValue);
-            ScrollAnimated((int)scrollDirection * -SCROLL_WHEEL * System.Windows.Forms.SystemInformation.MouseWheelScrollLines);
+            // Propotional scroll amount based on the scroll value, with a minimum of 8
+            float scrollMultiplier = scrollValue / 8f;
+            scrollMultiplier = Math.Max(8f, Math.Abs(scrollMultiplier)) * Math.Sign(scrollValue);
+
+            ScrollAnimated(-(int)scrollMultiplier * System.Windows.Forms.SystemInformation.MouseWheelScrollLines);
         }
 
         private ClickFocus GetScrollFocus(Point mousePos) => mousePos switch {
--- a/Blish HUD/Controls/Scrollbar.cs
+++ b/Blish HUD/Controls/Scrollbar.cs
@@ -161,11 +161,6 @@ namespace Blish_HUD.Controls {
 
             if (scrollValue == 0) return;
 
-            // Handle the scroll value overflow (happens when scrolling fast in one direction)
-            if (scrollValue < -32000) {
-                scrollValue += 65536;
-            }
-
             // Propotional scroll amount based on the scroll value, with a minimum of 8
             float scrollMultiplier = scrollValue / 8f;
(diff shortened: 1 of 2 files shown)


entrhopi:
Quick explanation:
The original `MouseEventArgs` used `SystemInformation.MouseWheelScrollDelta` which is just the delta value of a single mouse wheel increment.
We're extracting the raw data and bit shifting the data. Scroll speed is either a positive value starting at 0 or a negative value going backwards from 65536.
The `WheelDelta` value has been adjusted accordingly.

Blish HUD/Controls/Scrollbar.cs
```diff
@@ -158,10 +157,15 @@ private void HandleWheelScroll(object sender, MouseEventArgs e) {
...
-            if (GameService.Input.Mouse.State.ScrollWheelValue == 0) return;
+            int scrollValue = GameService.Input.Mouse.State.ScrollWheelValue;
 
-            float normalScroll = Math.Sign(GameService.Input.Mouse.State.ScrollWheelValue);
-            ScrollAnimated((int)normalScroll * -SCROLL_WHEEL * System.Windows.Forms.SystemInformation.MouseWheelScrollLines);
+            if (scrollValue == 0) return;
+
+            // Propotional scroll amount based on the scroll value, with a minimum of 8
+            float scrollMultiplier = scrollValue / 8f;
+            scrollMultiplier = Math.Max(8f, Math.Abs(scrollMultiplier)) * Math.Sign(scrollValue);
```
dlamkins: > Propotional 

Small typo.

How was 8 chosen?  I think this is 120 / 8 = 15 (so halving the scroll speed).  We should probably get it back into a constant and explain what it means.

I also think that this might behave weird for high-res mouse scrolling mice where they'll report smaller increments more frequently.  It would still add up to the same amount as a normal mouse, but with the minimum being 8, that might end up adding up a bit funky.  I'm not sure the best way to test that, but we could normalize against the standard wheel delta of 120, yeah?  Something like:

float scrollAmount = (scrollValue / 120f) * SCROLL_WHEEL;
ScrollAnimated(-(int)scrollAmount * System.Windows.Forms.SystemInformation.MouseWheelScrollLines);

And SCROLL_WHEEL can be 30 or 15 (if 15 felt better) or whatever.

Blish HUD/GameServices/Input/Mouse/MouseEventArgs.cs
```diff
@@ -40,8 +40,8 @@ public class MouseEventArgs : EventArgs {
 
         internal int WheelDelta {
             get {
-                int v                                              = Convert.ToInt32((this.MouseData & 0xFFFF0000) >> 16);
-                if (v > SystemInformation.MouseWheelScrollDelta) v -= ushort.MaxValue + 1;
+                int v             = (int)((this.MouseData & 0xFFFF0000) >> 16);
+                if (v > 32768) v -= ushort.MaxValue + 1;
```
dlamkins: I think it'd be basically impossible for it to land exactly on 32768, but I think this should technically be >=.

Blish HUD/Controls/Scrollbar.cs
entrhopi: 8 was the number I got during testing for one scroll wheel increment. I will dig into this a bit more, also testing with different mice. But I think it's correlated to the raw data we get from `MouseEventArgs`.

Blish HUD/GameServices/Input/Mouse/MouseEventArgs.cs
entrhopi: I think the highest number I saw war around 2500, but you're right, this should be >= 👍

Blish HUD/Controls/Scrollbar.cs
entrhopi: Alright. 8 seems to be an arbitrary number specific to my test system. On another it is as low as 1 using a trackpad, or as high (an interestingly static) as the `SystemInformation.MouseWheelScrollDelta` of 120 with another mouse.

I'll read some more of the documentation and come up with a good solution.

Thanks for the input!

entrhopi:
I've tested this with different mice on different machines now and I'm quite happy atm.

Addressing some things to consider when reviewing this:

#392
Scrollbar speed is still linear for some mice, that don't have acceleration in their scroll data but a fixed value. There is nothing we can do about that except for maybe tracking how many events happen in a period of time and artificially recreating an acceleration effect.
Other software seem to not do this thou, so I would refrain from doing this in Blish.

If you agree and merge this, then #392 can be closed.

dlamkins (approved):

greaka (approved):

Issue #998, opened 2025-01-11
Bug: String Width Calculation Issues


Flyga-M:
The title is so vague, because there are multiple issues with the string width calculations. I wanted to document them in one place. Feel free to change it to a more appropriate name.

Prompted by the recent discussion on discord about the `Label` wrap not calculating the text width correctly, I delved a bit into the source to look for the issue.

Latest discussion: https://discord.com/channels/531175899588984842/599270434642460753/1327399531393978440
Previous discussion: https://discord.com/channels/531175899588984842/536970543736291346/1152261761467162848

# Problems

## `MeasureString(..)`

It turns out the `MonoGame.Extended.BitmapFonts.BitmapFont.MeasureString(..)` on it's own works mostly as intended.

It uses the `StringGlyphEnumerable` to calculate the glyph (characters) positions based on the provided `FontRegion` and the `LetterSpacing`. For the width it chooses the highest value of the glyphs width added to their respective positions. The height is calculated by multiplying the amount of lines by the `LineHeight`.

![size_calculation](https://github.com/user-attachments/assets/7c1feb75-d338-4488-b4b1-934d735ef454)

`SpriteBatch.DrawStringOnCtrl()` calls `MonoGame.Extended.BitmapFonts.BitmapFontExtensions.DrawString(..)`, which uses the same `StringGlyphEnumerable` to draw the glyphs onto the `SpriteBatch`. If you measure the whole string you want to draw, the `MeasureWidth(..)` gives the exact result.

The only thing that `MeasureString(..)` is missing: Most, if not all, first characters in a line start at a negative x position. Since `MeasureString(..)` only calculates `position + width`, the calculated width is missing the offset to the left.

This on it's own is not a cause for the wrap issue, because the calculation fits the visual width of the rendered text (relative to the `Label`s `(0, 0)`), because it's also rendered offset to the left.

## `DrawUtil.WrapText(..)`
The `WrapText(..)` method (more specifically the `WrapTextSegment(..)` method) that is used all over Blish HUD separates the string into words, calculates their respective width and then glues them back together.

It calculates the width of the space and the words using the `MeasureString(..)` method mentioned above.

Now the small issue with `MeasureString(..)` is amplified a lot. Every single word and every space is 'losing' the negative offset from the first character.

Adding the calculated width of the words and spaces together is in most cases not the same as calculating the width of the whole string.

> [!NOTE]
> ### Example DefaultFont14
[...]

PR #1057, opened 2026-02-16
Fix: Wrong size measurement for WrapTextSegment

Denrage:
In specific edge cases `DrawUtil.WrapText` produces a string that is greater than the `maxLineWidth` you give him.
This happens because the previous implementation of `WrapTextSegment` didn't properly measured the string after appending a word.
The size of the space changes depending on which letter the space precedes so we can't use a fixed SpaceWidth for the whole wrapping and we need to measure the actual size of the string to not exceeed the `maxLineWidth`.

I don't know all the implications this change could occure, but i can't think of any problem with it, but that is limited to the knowledge i have regarding this function and usage all around Blish.

Also i don't know if `MeasureString("some string") + MeasureString("some other string")` will be the same as `MeasureString("some string" + "some other string")`. This could create another even more edgy case where we need to measure the resulting string before we create it to not exceed the `maxLineWidth`

This is intended to not break any existing implementation and should only fix an edge case.

Also unrelated to the current issue: `DrawUtil.WrapText` adds an unnecessary space at every line at the end.

Minimal reproduction that produces a wrong LineWidth string:
```
var test = "invidunt ut labore et dolore magna aliquyam erat, sed";
var result = DrawUtil.WrapText(GameService.Content.GetFont(FontFace.Menomonia, FontSize.Size16, FontStyle.Regular), test, 360);
var lineWidth = GameService.Content.GetFont(FontFace.Menomonia, FontSize.Size16, FontStyle.Regular).MeasureString(result); // lineWidth = 363
```

### DIFF
--- a/Blish HUD/_Utils/DrawUtil.cs
+++ b/Blish HUD/_Utils/DrawUtil.cs
@@ -1,10 +1,10 @@
 ﻿using Microsoft.Xna.Framework;
 using Microsoft.Xna.Framework.Graphics;
 using MonoGame.Extended.BitmapFonts;
-using System;
 using System.Linq;
 using System.Text;
 using Blish_HUD.Controls;
+using System.Collections.Generic;
 
 namespace Blish_HUD {
     public static class DrawUtil {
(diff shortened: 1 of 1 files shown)


Denrage:
I talked with @greaka about the added MeasureString operations and it is sadly something we need to do.
The resulting string has a different MeasureString size than the sum of it parts.
image

dlamkins:
Worth bringing in some context from earlier looks into this.

https://github.com/blish-hud/Blish-HUD/issues/998
https://github.com/blish-hud/Blish-HUD/pull/997

